PR #9, opened 2020-11-18
Add fragment support

mrkoreye:
This PR adds support for using fragments, as well as adds another block to test (button). I was not sure how exactly we wanted to handle fragment generation for liquid, so I kept the fragment as a `div`. The other option is to not add any wrapper element since have a single node does not matter in liquid.


steve8708:
awesome work

PR #15, opened 2020-11-24
Relax version requirement for core; remove unused run scripts

mandx:


packages/core/package.json
```diff
@@ -33,12 +33,11 @@
   },
   "scripts": {
     "test": "jest",
-    "build": "tsc",
+    "build": "tsc --declaration",
```
steve8708, line 36: why not do `declaration: true` in tsconfig instead? iirc our build already does emit declarations

mandx: I'm fine with `declaration: true` in the tsconfig. However it doesn't emit `.d.ts` definitions without this flag - I got a compile error when running the fiddle after lerna-linking, which went away when I ran `tsc` with this flag.

The `build:declaration` run script was failing anyway, that's why I removed it - I assumed it was remnants of ejection/default boilerplate.

mandx: You were right, declarations where emitted just fine. Apparently Lerna linked the wrong thing causing my compilation errors.

steve8708 (approved):

PR #16, opened 2020-11-24
Add support for responsive styles

mrkoreye:
This adds support for responsive styling. I decided to only support jsx-lite -> builder media queries that are `max-width` and pixel based, since when writing back to jsx-lite from builder we convert to that format.

e.g.
```
    <div
      css={{
        width: "100px",
        color: "red",
        "@media (max-width: 640px)": {
          width: "50px",
          color: "rgba(54, 70, 175, 1)",
        },
      }}
    > hello </div>
```

Keeping as a draft since it needs more testing!

### DIFF
--- a/packages/core/src/parsers/builder.ts
+++ b/packages/core/src/parsers/builder.ts
@@ -4,6 +4,7 @@ import { last, omit, pickBy } from 'lodash';
 import { createJSXLiteComponent } from '../helpers/create-jsx-lite-component';
 import { createJSXLiteNode } from '../helpers/create-jsx-lite-node';
 import { JSXLiteNode } from '../types/jsx-lite-node';
+import { sizes, Size } from '../constants/media-sizes';
 
 // Omit some superflous styles that can come from Builder's web importer
 const styleOmitList: (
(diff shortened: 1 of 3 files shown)


steve8708 (approved):

packages/core/src/parsers/builder.ts
```diff
@@ -194,6 +195,37 @@ export const builderElementToJsxLiteNode = (
...
+  const sizeKeys = Object.keys(block.responsiveStyles || {});
+  const hasCss = Boolean(sizeKeys.length);
+  const css: { [key: string]: Partial<CSSStyleDeclaration> } = {};
+  if (hasCss) {
+    for (const size of sizeKeys) {
+      if (size === 'large') {
+        css.large = omit(
+          {
+            ...css.large,
+            ...block.responsiveStyles!.large,
```
mandx, line 207:
```suggestion
            ...block.responsiveStyles?.large,
```

packages/core/src/parsers/builder.ts
```diff
@@ -194,6 +195,37 @@ export const builderElementToJsxLiteNode = (
     properties.builderTag = block.tagName;
   }
 
+  const sizeKeys = Object.keys(block.responsiveStyles || {});
```
mandx, line 198:
```suggestion
  const blockSizes: Size[] = Object.keys(block.responsiveStyles || {}).filter(k => sizeNames.includes(k));
```
The idea would be do all the validation (excluding what's not a `size`) and the type casting (`string as Size`) in one line/step, then the rest of the code should be type-safe-_ish_. The thing is that later in the code we do `sizes[size as Size].max` but if `size` is not really a `Size`, this will crash.

Also renaming the `const` reads better (the sizes associated to this block).

packages/core/src/parsers/builder.ts
```diff
@@ -194,6 +195,37 @@ export const builderElementToJsxLiteNode = (
...
+  const hasCss = Boolean(sizeKeys.length);
+  const css: { [key: string]: Partial<CSSStyleDeclaration> } = {};
+  if (hasCss) {
+    for (const size of sizeKeys) {
```
mandx, line 202: It's not necessary to check for length to start looping, just loop directly, and you save one level of nesting.

steve8708 (approved):

PR #33, opened 2020-12-02
Relax version requirements for `core`, so `lerna link` works

mandx:


packages/core/src/generators/react.ts
```diff
@@ -57,19 +57,21 @@ const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
   const children = json.children.filter(filterEmptyTextNodes);
 
   if (json.name === 'For') {
-    str += `{${processBinding(json.bindings.each as string, options)}.map(${
+    str += `<>{${processBinding(json.bindings.each as string, options)}.map(${
       json.bindings._forName
     } => (
-      ${children.length === 1 ? '' : '<>'}
+      ${children.length === 1 ? '' : ''}
```
steve8708, line 63: why this change? the goal is to do proper idiomatic code

{foo.map(item => <OneChild>...</OneChild />)}

{foo.map(item => <><FirstChild /><SecondChild></>)}

did the old code, as is, have problems?

packages/core/src/__tests__/__snapshots__/react.test.ts.snap
```diff
@@ -475,9 +467,11 @@ export default function MyComponent(props) {
...
       name={props.name}
     >
-      {props.options.map((option) => (
-        <option value={option.value}>{option.name || option.value}</option>
-      ))}
+      <>
+        {props.options.map((option) => (
+          <option value={option.value}>{option.name || option.value}</option>
+        ))}
+      </>
```
steve8708, line 474: per my other comment, the prior code looks like right here and this adds an unnecessary fragment

packages/core/src/generators/react.ts
mandx: My bad, this is a mistake, should be something like
```suggestion
      ${children.length === 1 ? '' : '<>'}
```

packages/core/src/generators/react.ts
```diff
@@ -57,19 +57,21 @@ const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
   const children = json.children.filter(filterEmptyTextNodes);
 
   if (json.name === 'For') {
-    str += `{${processBinding(json.bindings.each as string, options)}.map(${
+    str += `<>{${processBinding(json.bindings.each as string, options)}.map(${
```
steve8708, line 60: this is not needed either

mandx, line 60: @steve8708 Without this fragment, the `forms.raw` test generates this invalid code:
```typescript
{Boolean(props.builderBlock && props.builderBlock.children) && (
    // This leading `{` is a syntax error in this JSX context,
    // so wrapping it with a fragment fixes it
    {props.builderBlock.children.map(block => (          
        <BuilderBlockComponent  key={block.id}  block={block} ></BuilderBlockComponent>
    ))}
)}
```

steve8708: oh I see hmm that would affect everything and not just loops - e.g. will affect text, nested conditions, etc

we may want to, for now, remove all children.length === 1 logic and always wrap contents in fragments. I wish prettier could remove them when not needed. maybe we can do some other post processing with babel to remove them when not needed

but anyway, we won't want this guard here, we want it in all the places we wrap in a fragment only if children.length != 1

mandx: The problem is the braces that follow the `<>`. Instead of forcing the fragment, the better solution IMO would be to conditionally add them, depending of the context, like they would be illegal in a boolean context (like when using `&&`) but they should be added when the expression is used as JSX content/children.

steve8708: yeah that could get complex but is also an option. we just don't know how children will render until they do, necessarily. I think the only resilient option is what I mention below is to postprocess the final code with babel and remove unnecessary fragments

steve8708, line 60:
```suggestion
    str += `{${processBinding(json.bindings.each as string, options)}.map(${
```

packages/core/src/generators/react.ts
```diff
@@ -57,19 +57,21 @@ const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
...
       ${children.length === 1 ? '' : '</>'}
-    ))}`;
+    ))}</>`;
   } else if (json.name === 'Show') {
     str += `{Boolean(${processBinding(
       json.bindings.when as string,
       options,
     )}) && (
-      <>${children.map((item) => blockToReact(item, options)).join('\n')}</>
+      ${children.length === 1 ? '' : '<>'}
```
steve8708, line 72:
```suggestion
     <>
```

packages/core/src/generators/react.ts
```diff
@@ -57,19 +57,21 @@ const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
...
+        ${children.map((item) => blockToReact(item, options)).join('\n')}
+      ${children.length === 1 ? '' : '</>'}
```
steve8708, line 74:
```suggestion
    </>
```

packages/core/src/generators/react.ts
```diff
@@ -57,19 +57,21 @@ const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
...
   if (json.name === 'For') {
-    str += `{${processBinding(json.bindings.each as string, options)}.map(${
+    str += `<>{${processBinding(json.bindings.each as string, options)}.map(${
       json.bindings._forName
     } => (
       ${children.length === 1 ? '' : '<>'}
         ${children.map((item) => blockToReact(item, options)).join('\n')}
       ${children.length === 1 ? '' : '</>'}
-    ))}`;
+    ))}</>`;
```
steve8708, line 66:
```suggestion
    ))}`;
```

steve8708 (approved):

PR #35, opened 2020-12-10
First pass at the video block

mandx:
Currently WIP until I figure out a way to use effects (some props are supposed to be used as "boolean" attributes; as in, the attribute must be added or removed, not just set to `true` or `false`.

Unsure if this is even possible, since this quirk is highly specific to HTML; I guess it breaks the abstraction a bit...

### DIFF
--- a/packages/core/src/__tests__/liquid.test.ts
+++ b/packages/core/src/__tests__/liquid.test.ts
@@ -8,6 +8,7 @@ const submitButtonBlock = require('./data/blocks/submit-button.raw');
 const button = require('./data/blocks/button.raw');
 const textarea = require('./data/blocks/textarea.raw');
 const img = require('./data/blocks/img.raw');
+const video = require('./data/blocks/video.raw');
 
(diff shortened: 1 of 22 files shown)


steve8708 (approved):
this works 👍  fancy lazy stuff isn't critical for first pass

packages/core/src/__tests__/liquid.test.ts
```diff
@@ -57,4 +58,10 @@ describe('Liquid', () => {
     const output = componentToLiquid(json);
     expect(output).toMatchSnapshot();
   });
+
+  test('Video', () => {
```
steve8708, line 62: it may be nice to ultimately just update these tests to read the `data/blocks` dir and run tests against each automatically so it's easier to manage these tests

mandx: See #36 :D

steve8708: 👍 👍

PR #37, opened 2020-12-10
Add primitive support for `{props.children}` & Section block

mandx:


packages/core/src/__tests__/__snapshots__/react.test.ts.snap
```diff
@@ -464,6 +464,47 @@ export default function MyComponent(props) {
...
+            {builderBlock?.children.map((block) => (
+              <>
+                <BuilderBlockComponent
+                  key={block.id}
+                  block={block}
+                ></BuilderBlockComponent>
+              </>
+            ))}
+          </>
+        )}
```
steve8708, line 500:
```suggestion
      {props.children}
```

packages/core/src/__tests__/data/blocks/section.raw.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {...props.attributes}
+      style={
+        maxWidth && typeof maxWidth === 'number' ? { maxWidth } : undefined
+      }
+    >
+      <Show when={builderBlock && builderBlock.children}>
+        <For each={builderBlock?.children}>
+          {(block) => <BuilderBlockComponent key={block.id} block={block} />}
+        </For>
+      </Show>
```
steve8708, line 27:
```suggestion
     {props.children}
```

so this will work in any react-ish frameworks out of the box (react, react native, solid)

then we can handle this for other frameworks. e.g. for vue, svelt you can do 

```ts
if (block.bindings._text?.replace(/\s+/g, '') === 'props.children') {
  return `<slot></slot>`
}
```

e.g. here https://github.com/BuilderIO/jsx-lite/blob/main/packages/core/src/generators/svelte.ts#L44:L44

for angular, do the same but `return '<ng-content></ng-content'`

I would also add one more `return '<slot></slot>'` implementation for custom elements too (in html.ts)

the other frameworks (e.g. liquid, swift) don't worry about supporting this feature for now

mandx: @steve8708 there's now some WIP code in this PR that focus on `{children}` codegen

packages/core/src/generators/svelte.ts
```diff
@@ -36,11 +36,18 @@ const mappers: {
...
-export const blockToSvelte = (json: JSXLiteNode, options: ToSvelteOptions) => {
+export const blockToSvelte = (
+  json: JSXLiteNode,
+  options: ToSvelteOptions,
+): string => {
   if (mappers[json.name]) {
     return mappers[json.name](json, options);
   }
 
+  if (`${json.bindings._text || ''}`.replace(/\s+/g, '') === 'props.children') {
```
steve8708, line 47: since we do this in a few places now would be good to make like a helpers/is-children

so you can do `if (isChildren(json))` 

```
const isChildren = (json: JSXLiteNode) => ...
```

steve8708 (approved):

PR #38, opened 2020-12-15
CLI proposal

andreruffert:
Hi lovely fellows 👋

Have you ever thought about a `jsx-lite` CLI (command line interface)?
I think this would be super useful!

The purpose of this PR is to get some feedback and start a discussion how the API could look like.
Here is a simple proposal:

```
$ jsx-lite --help

  Usage
    $ jsx-lite <files> <command> [options]

  Commands
    vue, react, angular, svelte, reactNative, solid, webcomponents, html, liquid, json, builder

  Options
    --output-dir, -d    Output directory (default: ".")         [string]

    <svelte>
    --state-handling    variables, proxies                      [string]

    <react>
    --style-library     emotion, styledComponents, styledJsx    [string]
    --state-library     useState, mobx, solid                   [string]

    <reactNative>
    --state-library     useState, mobx, solid                   [string]

  Examples
    $ jsx-lite example.jsx vue
    $ jsx-lite example.jsx svelte --state-handling=proxies
    $ jsx-lite source/**/*.jsx react --output-dir=packages/react

```

Wdyt?

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,42 @@
+# @jsx-lite/cli
+
+> JSX lite command line interface.
+
+
+## Install
+
+```
+$ npm i @jsx-lite/cli -D
+```
+
+
+## Usage
+
+```
+$ jsx-lite --help
+
+  Usage
+    $ jsx-lite <files> <command> [options]
+
+  Commands
+    vue, react, angular, svelte, reactNative, solid, webcomponents, html, liquid, json, builder
+
+  Options
+    --output-dir, -d    Output directory (default: ".")         [string]
+
+    <svelte>
+    --state-handling    variables, proxies                      [string]
+
+    <react>
+    --style-library     emotion, styledComponents, styledJsx    [string]
+    --state-library     useState, mobx, solid                   [string]
+
+    <reactNative>
+    --state-library     useState, mobx, solid                   [string]
+
+  Examples
+    $ jsx-lite example.jsx vue
+    $ jsx-lite example.jsx svelte --state-handling=proxies
+    $ jsx-lite source/**/*.jsx react --output-dir=packages/react
+
+```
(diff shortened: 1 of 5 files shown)


packages/cli/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Usage
+
+```
+$ jsx-lite --help
+
+  Usage
+    $ jsx-lite <files> <command> [options]
+
+  Commands
+    vue, react, angular, svelte, reactNative, solid, webcomponents, html, liquid, json, builder
```
mandx, line 22: How about making the command strings be actual verbs? Then the target library framework would be another (required) argument? This would also allow for adding future commands that don't necessarily relate to a framework.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      <svelte>
+      --state-handling    variables, proxies                      [string]
+
+      <react>
+      --style-library     emotion, styledComponents, styledJsx    [string]
+      --state-library     useState, mobx, solid                   [string]
+
+      <reactNative>
+      --state-library     useState, mobx, solid                   [string]
```
mandx, line 40: Is there a way to have the `meow` library generate usage text based on the flags? I just worry about the actual flags/arguments getting out of sync with the help text (a JSX-Lite CLI has the potential to grow a lot!)

packages/cli/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+  Options
+    --output-dir, -d    Output directory (default: ".")         [string]
```
mandx, line 25: This might be a personal thing, but sometimes I like to just dump stuff to my terminal for quick inspection, instead of writing to files that I would have to open with some viewer/editor... How do you think we could accommodate that workflow?

IMO that would be specially helpful during this development stage when a lot of prototyping and testing is going on...

steve8708:
thanks for the ideas @andreruffert - we've got an official CLI over here #60 that will close this in favor of

PR #40, opened 2021-01-08
Vue Gen fixes + `innerHTML` support

mandx:
* Vue: Fixing value interpolation and prop enumeration
    * This (for now at least) requires that the `props` param must not be destructured, otherwise members plucked from the props won't be recognized.
* Allow using `innerHTML` JSX-Lite prop (maps to React's `dangerouslySetInnerHTML` and Vue's `v-html`)
* Added the `Text` Builder block


steve8708 (approved):

steve8708:
great stuff 👍

PR #45, opened 2021-01-18
rename afterRender hook -> onMount

steve8708:
agree Vue's hooks names are the best so far so standardizing on that feels right @mandx


steve8708:
going to do some work that will depend on this, so eager merging and happy to followup with additional PRs to address feedbaack

PR #48, opened 2021-01-19
Transform `refs` in hooks as well + misc cleanups

mandx:


packages/core/src/types/jsx-lite-component.ts
```diff
@@ -36,8 +36,7 @@ export type JSXLiteComponent = {
   meta: { [key: string]: JSON | undefined };
   state: { [key: string]: JSON | undefined };
   hooks: {
-    [key: string]: string | undefined;
```
steve8708, line 39: would still like to keep this here, so people know this is user extensible (e.g. by plugins), and we can hardcode the things we know we want to support here below too

also fyi I have another PR changing this name to "onmount" could use a review on

steve8708 (approved):

packages/core/src/types/jsx-lite-component.ts
steve8708: we can add a comment here to make this more clear if you like - what this is here for so it stays

packages/core/src/generators/vue.ts
```diff
@@ -174,6 +186,8 @@ export const componentToVue = (
 
   const elementProps = getProps(component);
 
+  const hasSetup = !!component.hooks.afterRender;
```
steve8708, line 189: just realized I already merged in the `afterRender` -> `onMount` PR, so we should use that name now here https://github.com/BuilderIO/jsx-lite/pull/45

packages/core/src/generators/vue.ts
```diff
@@ -196,6 +210,20 @@ export const componentToVue = (
...
         }
+        ${
+          hasSetup
+            ? `setup() {
+              ${
+                component.hooks.afterRender
+                  ? `onMounted(${processBinding(
+                      component.hooks.afterRender,
+                      options,
+                    )})`
```
steve8708, line 221: also fyi I've changed onMount hook to only take the function body, so this will need to be like

`onMount(() => { ${...} })`

packages/core/src/types/jsx-lite-component.ts
mandx: My argument is to leave this out for now, so we have stricter typings while we do all of this crazy, high-paced refactorings. For example, after `afterRender` -> `onMount` PR got merged, my code kept typechecking, even after `afterRender` being renamed...

Then we things slow down and "solidify" can add it back for all the plugins' needs.

steve8708: 👍

PR #49, opened 2021-01-21
Add custom code block

mandx:


steve8708 (approved):
great 👍  

it's real useful seeing these ported over so we can also see that we will ultimately probably want a shared library (maybe @builder.io/utils) that we can import some shared logic like findAndRunScripts

this is really hepful for making progress and having a further understnading of some of the parts we'll want to ultimately add

PR #51, opened 2021-01-21
A probably buggy `onUnMounted` hook implementation

mandx:
Maybe has some problems, but I think it's a good starting point to revisit/improve later.

This is mostly needed to cleanup DOM event listeners.


steve8708 (approved):
perfect 👍  we'll find and squash the bugs later

PR #61, opened 2021-02-25
Fix `functions/satisfies` error when parsing JSLite code

mandx:


packages/core/src/parsers/jsx.ts
```diff
@@ -12,6 +12,7 @@ import { stripNewlinesInStrings } from '../helpers/replace-new-lines-in-strings'
 
 const jsxPlugin = require('@babel/plugin-syntax-jsx');
 const tsPreset = require('@babel/preset-typescript');
+const _semver = require('semver');
```
steve8708, line 15: can you explain more here? not sure I understand the problem or the solution

steve8708: if you just need to pull it in maybe just do `require('semver')` ?

steve8708: seems the fiddle is unaffected by this, maybe this should be in the builder webapp instead? maybe just the require and a comment. I'll take a look at

steve8708: I'm unable to reproduce locally so seems this is a production only issue, how confident are you that this is the fix? and can it live in the Builder.io webapp instead if that's the only place this issue exists?

mandx: I was able to repro locally... I'll try again, and also try adding a `require('semver')` in the webapp instead and check if the error goes away.

steve8708 (approved):
perfect, thanks 🙏

PR #74, opened 2021-03-09
Cli build command

steve8708:


steve8708:
will continue to flesh out then document, but want to first get eyes on this starting point to make sure I didn't bust anything for the current CLI setup

Dkendal (approved):
Looks good! 

To me, the name `generate` seems to suggest that it's generating new files, or otherwise scaffolding files based on some flags, but not based on source input. Given the prevalence of tools referring to stuff like this as "generators" there might be a fair bit of confusion. I'm curious why you chose this name over the `compile` option you mentioned earlier?

steve8708:
ah yes, was torn on compile v generate, will flip back to compile 👍

Issue #36, opened 2020-12-10
Iterate over all submodules in `core/src/__tests__/data/blocks/*` and run snapshot tests on each


mandx:
> It may be nice to ultimately just update these tests to read the data/blocks dir and run tests against each automatically so it's easier to manage these tests

luislpv22:
Hello,

I've made some progress on this, but right now you are running different test cases for each suite, so could you tell me how you'd like to manage that?

Example:
**parse-jsx.test.ts** is running a single test case on this block: **button-with-metadata.raw.tsx**
If we loop through /data/blocks, this suite (and all of them) would run a test case for every block.

Thanks.

mandx:
Hi @luislpv22! First, apologies for the sparse text here, I guess it looks like more like a paper note than a proper ticket.

The idea is to enumerate all test blocks (all the modules/components in `packages/core/src/__tests__/data`, which should contain one component/block each), then enumerate all available targets (React, Vue, etc...) and for each of these combinations of (block, target) simply do what most of the tests are doing already:

- Parse the current block as JSXLite code (for example, the `Image` block)
- Use the parsed block to output a component in the current target (ex Vue)
- Assert the generated component code matches the previous snapshot.

In `packages/core/src/__tests__/` we have:

```
data/ 👈 This is where all the blocks live

👇 These are the tests modules that would be replaced by a single test module that enumerates blocks + targets:
builder.test.ts
html.test.ts
liquid.test.ts
react.test.ts
solid.test.ts
vue.test.ts

👇 This is a different type of test, this can be left as is.
parse-jsx.test.ts
```
I guess we would end up with something like:
```
data/ 👈 This is where all the blocks live
block-snapshots-tests.ts 👈 this would be the new module that iterates and generates the snapshot tests
parse-jsx.test.ts 👈 same file as before, unchanged.
```
(`block-snapshots-tests.ts` is a placeholder name btw)

I think for now just having a `const` array/set at the top of the `block-snapshots-tests.ts`, since the target list doesn't change that often.

samijaber:
Closing this because of the progress done on #484. It's not iterating over everything, but allowing each generator to opt-in to the blocks it needs to test, which helps account for discrepancies between them.

PR #75, opened 2021-03-09
Dynamically loading test blocks

luislpv22:
Some progress made on issue #36 

A bunch of tests are failing now because we are testing all blocks under /data/blocks, and some of them are giving errors

We might check them.


packages/core/src/__tests__/snapshots.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      const jsxLite = componentToJsxLite(backToJsxLite);
+      expect(jsxLite).toMatchSnapshot();
+    });
+  });
+});
+
+describe('Html', () => {
+  testCases.forEach((data: TestData) => {
+    test(data.name, () => {
+      const jsx: string = require(data.file);
```
mandx, line 35: I don't think we need to use `require` and invoke Node's module resolution algorithm, using `fs.readFileSync` should be enough, something like:
```suggestion
      const jsx = fs.readFileSync(data.file, 'utf8');
```

mandx (changes requested):
I added some minor suggestions, in general this is a great starting point. I'll look into why the tests are failing; looks like for some reason no code is being transpiled, so code that expect non-empty strings is crashing.

luislpv22:
I've added your suggestions. Those are the blocks failing in tests:

 - All parsers › builder-render-block

```js
SyntaxError: JSON5: invalid end of input at 1:1

      490 |   );
      491 |   try {
    > 492 |     return JSON5.parse(toParse);
```
 - All parsers › builder-render-content

```js
SyntaxError: JSON5: invalid end of input at 1:1

      490 |   );
      491 |   try {
    > 492 |     return JSON5.parse(toParse);
```
- Html parser › form

```js
SyntaxError: unknown: Unexpected token, expected ";" (1:12)

    > 1 | let _ = get submissionState() {
        |             ^
      2 |   return Builder.isEditing && props.previewState || state.state;
      3 | }
```
- Html parser › columns

```js
SyntaxError: unknown: Unexpected token, expected ";" (3:13)

      1 | let _ = // TODO: These methods are not used right now, but they will be when
      2 | // support for dynamic CSS lands
    > 3 | getColumns() {
        |              ^
```

mandx:
Thanks @luislpv22 I'll take another look later today :)

steve8708:
hey @mandx - any update here?

mandx:
@luislpv22 Looks like this PR has diverged a lot from the main branch; I'll suggest two things to get this restarted:

- Start over with a new/fresh branch, copy code as needed from this branch.
- Try make _additive_ changes, that is, let's introduce new unit-tests modules/files/snapshots instead of overwriting existing ones; worst case, we would be testing the same thing twice, but this will allow avoiding rebase/merge hell since the new PR will be mostly adding code instead of modifying existing code. We can always remove duplicated tests in a secondary PR.

steve8708:
closing out due to inactivity to keepour PR list clean, please feel free to reopen @luislpv22 if you'd like to continue on this

PR #82, opened 2021-03-22
Progress on TodoMVC examples w/ several outputs

steve8708:


steve8708:
merging to keep this from divering, as at this point I am going to be implementing features like services/context

PR #83, opened 2021-03-24
eslint jsx callback name rule

Dkendal:
Adds an eslint rule to check if that a jsx-attribute's callback name is valid (given the current limitations).

Adds the pkg `ts-pattern` for writing rules with pattern matching. IMO pattern matching makes it much easier to write these rules and makes them easier to reason about. I chose ts-pattern over a few other options because it has 0 dependencies, stable development typescript integration, and no additional tooling requirements.


steve8708 (approved):
sweet

steve8708 (approved):
awesome

PR #93, opened 2021-06-01
Fiddle UI improvements

steve8708:
Screen Shot 2021-06-01 at 7 07 25 AM


steve8708:
@Dkendal for some reason I can't get tests to pass - unable to require.resolve the json modules, have you seen this before or have any ideas?

steve8708:
CLI tests don't seen to work on `main` branch either so maybe we can ignore in regards to this PR as it's unrelated and worth a look separately

Dkendal:
@steve8708 the require.resolve issues are resolved, the other errors look (more) legit, did something change with the jsx-lite/core build process?

steve8708:
Shouldn't have no. I'm seeing different errors locally (trying to run .d.ts files as tests and complaining you cannot have files with no tests)

Dkendal:
I ran into that as well it was just a left over from running TSC and then getting compiled for some reason. You can delete any .d.ts and js files there. If there are any js files they'll break the tests as they ll get sourced instead of the ts files.

Dkendal:
Alternatively run`git clean packages/cli/src -i`

steve8708:
Still am having issues getting it to work - can you take over on getting cli tests passing when time permits?

Dkendal:
Sure thing

Dkendal:
@steve8708 fixed

packages/fiddle/src/index.tsx
```diff
@@ -2,22 +2,26 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
 import App from './components/App';
-import reportWebVitals from './reportWebVitals';
 import { configure } from 'mobx';
 
 configure({
   enforceActions: 'never',
+  // useProxies: 'ifavailable',
```
Dkendal, line 9: :scissors:

Dkendal (approved):

packages/fiddle/src/index.tsx
Dkendal, line 9:
```suggestion
```

Dkendal (approved):

packages/fiddle/src/index.tsx
steve8708: oops! nice catch :)

steve8708:
@Dkendal 🙏 🙏🙏

PR #95, opened 2021-06-03
Handle HTML void elements when reading/parsing Builder content

mandx:
Since some Builder blocks can contain arbitrary HTML code, valid HTML like:

```html
<p><br></p>
```

Can crash the parsing stage, so this is simple fix to convert that to:

```html
<p><br /></p>
```

So it's safer to parse.

### DIFF
--- a/packages/core/src/parsers/builder.ts
+++ b/packages/core/src/parsers/builder.ts
@@ -781,19 +781,23 @@ export const builderContentToJsxLiteComponent = (
       const voidElemRegex = /(<area|base|br|col|embed|hr|img|input|keygen|link|meta|param|source|track|wbr[^>]+)>/gm;
 
       try {
-        elem.component.options.text = elem.component.options.text.replace(
-          voidElemRegex,
-          '$1 />',
-        );
+        if (elem.component?.name === 'Text') {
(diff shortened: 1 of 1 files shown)


packages/core/src/parsers/builder.ts
```diff
@@ -769,6 +771,35 @@ export const builderContentToJsxLiteComponent = (
...
+  builderContent = fastClone(builderContent);
+  traverse(builderContent).forEach(function(elem) {
+    if (elem && elem['@type'] === '@builder.io/sdk:Element') {
+      // Try adding self-closing tags to void elements, since Builder Text
+      // blocks can contain arbitrary HTML
+      // List taken from https://developer.mozilla.org/en-US/docs/Glossary/Empty_element
+      // TODO: Maybe this should be using something more robust than a regular expression
+      const voidElemRegex = /(<area|base|br|col|embed|hr|img|input|keygen|link|meta|param|source|track|wbr[^>]+)>/gm;
+
+      try {
```
steve8708, line 783: would be great to wrap this in an 

```js
if (component.options?.name === 'Text') { 
```

steve8708 (approved):

PR #97, opened 2021-06-11
chore: Add `bazel` build system for `packages/core` so that this repo…

mhevery:
… can be a sub-repo of `builder-internal`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,6 +49,7 @@
   "dependencies": {
     "@angular/compiler": "^11.2.11",
     "@babel/core": "^7.12.3",
+    "@babel/generator": "^7.14.3",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@babel/plugin-syntax-decorators": "^7.12.1",
     "@babel/plugin-syntax-jsx": "^7.12.1",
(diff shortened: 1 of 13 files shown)


packages/core/package.json
```diff
@@ -81,17 +82,22 @@
...
     "@types/prettier": "^2.1.5",
     "@types/rollup__plugin-virtual": "^2.0.1",
     "@types/traverse": "^0.6.32",
     "jest": "^26.6.3",
     "jest-raw-loader": "^1.0.1",
-    "ts-jest": "^26.4.4"
+    "lru-cache": "^6.0.0",
+    "prettier": "^2.3.0",
+    "ts-jest": "^26.4.4",
+    "universalify": "^2.0.0"
```
Dkendal, line 101: I'm curious why these new dependencies are required? Some of them are missing from the bazel BUILD file as well.

Dkendal (approved):

mrkoreye (approved):

packages/core/package.json
mhevery: Good question. If I remove it I get errors like these:
```
could not find dependency 'universalify' of '@jsx-lite/eslint-plugin/node_modules/ts-jest'
```
I think it may have something to do with having to explicitly declare peer dependencies. Not sure.

PR #102, opened 2021-06-22
migrate to subset of React API 

steve8708:
also adds support for inline symbols

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
@@ -478,7 +478,7 @@ export default function FormComponent(props) {
       action={!props.sendWithJs && props.action}
       method={props.method}
       name={props.name}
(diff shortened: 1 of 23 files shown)


steve8708:
still have more testing and one more thing to complete before this is ready

steve8708:
I have one last thing to fix but shouldn't block review so openning it back up

packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
```diff
@@ -478,7 +478,7 @@ export default function FormComponent(props) {
       action={!props.sendWithJs && props.action}
       method={props.method}
       name={props.name}
-      onSubmit={(event) => state.onSubmit(event)}
+      onSubmit={(event) => state.state.onSubmit(event)}
```
steve8708, line 481: ignore this `state.state` stuff, I am looking into :)

PR #104, opened 2021-06-22
tabs component setup example

steve8708:


packages/core/src/__tests__/data/blocks/tabs.raw.tsx
```diff
@@ -0,0 +1,41 @@
...
+        flexDirection: 'row',
+        overflow: 'auto',
+        overflowScrolling: 'touch',
+      }}
+      style={{
+        justifyContent: props.tabHeaderLayout!,
+      }}
+      class="builder-tabs-wrap"
+    >
+      <For each={props.tabs}>{(child) => <div>{child.content}</div>}</For>
```
mandx, line 38:
```suggestion
      <For each={props.tabs}>{(child) => <div class="builder-tab-content>{child.content}</div>}</For>
```

mandx (approved):

packages/core/src/__tests__/data/blocks/tabs.raw.tsx
steve8708: oh sorry @mandx this is just an example for Adam, no review needed (accidentally triggered the auto assign)

PR #105, opened 2021-06-28
fix(qwik): Updated file format to match latest Qwik best practice

mhevery:


steve8708 (approved):

packages/core/src/generators/qwik.ts
```diff
@@ -586,3 +586,7 @@ export const componentToQwik = async (
 
   return output;
 };
+
+function htmlAttributeEscape(value: string): string {
+  return value.replace(/"/g, '&quot;').replace(/\n/g, '\\n');
+}
```
steve8708, line 592: not critical - but we have `packages/cores/src/helpers` this could go in, like 

packages/core/src/helpers/html-attribute-escape
```tsx
export function htmlAttributeEscape(value: string): string {
  return value.replace(/"/g, '&quot;').replace(/\n/g, '\\n');
}
```

mhevery: ok, will move.

PR #106, opened 2021-06-30
fix: Revert to `legacy` format; fix code-generation to be more resiliant

mhevery:


steve8708 (approved):

packages/core/src/__tests__/__snapshots__/react.test.ts.snap
```diff
@@ -8,7 +8,7 @@ export default function MyBasicComponent(props) {
 
   return (
     <div>
-      <input value={name} onChange={(event) => setName(event.target.value)} />
+      <input onChange={(event) => setName(event.target.value)} />
```
steve8708, line 11: something looks wrong here

packages/core/src/generators/react.ts
```diff
@@ -160,7 +163,9 @@ export const blockToReact = (json: JSXLiteNode, options: ToReactOptions) => {
         str += ` ${BINDING_MAPPERS[key]}={${useBindingValue}} `;
       }
     } else {
-      str += ` ${key}={${useBindingValue}} `;
+      if (!isValidAttributeName(key)) {
```
steve8708, line 166:
```suggestion
      if (isValidAttributeName(key)) {
```

this should fix the snapshots

mhevery: fixed

packages/core/src/__tests__/__snapshots__/react.test.ts.snap
mhevery: Nice catch! fixed.

PR #122, opened 2021-08-10
fix: explicitly pass `configFile: false` to prevent mitosis from pull…

mhevery:
…ing configuration from environment.


steve8708 (approved):
🙏 🙏

PR #128, opened 2021-08-20
fix example codes

tomByrer:
I think these are typos?


steve8708:
Great finds - thanks @tomByrer !

Issue #134, opened 2021-10-22
Migrate Mitosis to latest Qwik format


mhevery:
Currently, Mitosis generates the original prototype version of Qwik. Update the mitosis to generate the latest `qHook` syntax.

- Idea is that this will be compatible with the optimizer. (https://github.com/BuilderIO/qwik/issues/60)
- But should run without optimizer as well.

## Mitosis
- [X] Incorrectly treats `css` binding as `on:s`
- [x] Incorrectly binds `css` instead of `style`

## Builder
- [ ] Loading `@builder.io/qwik` for each request. Should be moved to shared for performance.
- [ ] Unable to load relative requests.
- [ ] Fix `underine` => `underline`

samijaber:
@mhevery can this be closed?

PR #147, opened 2021-12-07
fix(preview): underine -> underline

Haroenv:
could be related to https://github.com/BuilderIO/mitosis/issues/134


steve8708:
thanks @Haroenv !

PR #150, opened 2021-12-21
Feature: Configurable Target Options

samijaber:
Related PR: https://github.com/BuilderIO/builder/pull/674
### Main changes:
- added an `options` object in `MitosisConfig` containing options for each framework (starting with Vue).
- remove our harcoded builder config relating to injecting a `registerComponent` import for our Vue SDK.
- added `path` as an (optional) argument for transpilers.
- all framework transpilers are now curried with signature `(options) => ({ component, path }) => string`. This makes it easier to add arguments to transpilers in the future, and allows us to initialize the transpilers at a higher level, before components are provided for transpilation.


packages/core/src/generators/vue.ts
```diff
@@ -501,33 +504,37 @@ export const componentToVue = (
...
-  return str;
-};
+    for (const pattern of removePatterns) {
+      str = str.replace(pattern, '');
+    }
+
+    // Transform <FooBar> to <foo-bar> as Vue2 needs
+    return str.replace(/<\/?\w+/g, (match) =>
+      match.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase(),
+    );
```
samijaber, line 536: This component name transform was being done outside of `componentToVue` which seems highly odd. Moved it in here for consistency.

packages/core/src/generators/vue.ts
```diff
@@ -387,31 +388,33 @@ export const componentToVue = (
...
+    let str = dedent`
     <template>
       ${template}
     </template>
     <script>
       ${renderPreComponent(component)}
       ${
-        !builderRegister
-          ? ''
-          : `import { registerComponent } from '@builder.io/sdk-vue'`
```
samijaber, line 406: This Builder-specific line (which is causing bugs for us) is now replaced with a more generic `registerComponentPrepend` option that can be provided by users at run-time.

packages/core/src/generators/vue.ts
```diff
@@ -387,31 +388,33 @@ export const componentToVue = (
...
 
       export default ${!builderRegister ? '' : 'registerComponent('}{
         ${
           !component.name
             ? ''
             : `name: '${
-                options.namePrefix ? options.namePrefix + '-' : ''
+                options.namePrefix?.(path)
+                  ? options.namePrefix?.(path) + '-'
+                  : ''
```
samijaber, line 417: This option is now a function that receives the path, because the prefix is generated based off of the path.

packages/core/src/types/config.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export type MitosisConfig = {
+  type?: 'library'; // Only one type right now
+  targets: Target[];
+  dest?: string;
+  files?: string | string[];
+  overridesDir?: string;
+  mapFile?: (info: FileInfo) => FileInfo | Promise<FileInfo>;
+  options: Partial<GeneratorOptions>;
```
samijaber, line 35: `MitosisConfig` now includes an options object with a key per each target/framework

packages/cli/src/build/build.ts
```diff
@@ -129,31 +173,12 @@ async function outputTsxLiteFiles(
...
-      (target === 'reactNative'
-        ? componentToReactNative(mitosisJson, {
-            stateType: 'useState'
-          })
-        : target === 'vue'
-        ? componentToVue(mitosisJson, {
-            cssNamespace: id,
-            // TODO: config for this
-            namePrefix: path.includes('/blocks/') ? 'builder' : undefined,
-            builderRegister: true
```
samijaber, line 144: these are the Vue options that are now provided by our mitosis config higher up

packages/core/src/__tests__/__snapshots__/vue.test.ts.snap
```diff
@@ -272,23 +272,23 @@ exports[`Vue Form block 1`] = `
     :name=\\"name\\"
     @submit=\\"onSubmit(event)\\"
   >
-    <BuilderBlockComponent
+    <builder-block-component
```
samijaber, line 275: These tests were updated because of xxx

steve8708 (approved):
LGTM 👍 

I'm not a huge fan of currying myself, an object with named arguments I generally find preferable over currying (aka `foo({ a, b, c })` vs `foo(a)(b)(c)`) but I don't feel super strongly on it

PR #155, opened 2022-01-05
fix(qwik): handle statements by wrapping them in IIF.

mhevery:
- Binding expressions are sometimes statements and need to be converted into IIF.
- `OnMount` code can declare `state` which fails execution. Wrapping it in IIF solves it.

https://builder-io.atlassian.net/browse/ENG-2020

### DIFF
--- a/packages/core/src/generators/qwik/src-generator.ts
+++ b/packages/core/src/generators/qwik/src-generator.ts
@@ -279,12 +279,11 @@ export class SrcBuilder {
         }
         this.isJSX ? this.emit(key) : this.emit(possiblyQuotePropertyName(key));
         this.isJSX ? this.emit('={') : this.emit(':', WS);
-        const bindingExpressionOrStatement = bindings[key];
-        this.emit(
-          isExpression(bindingExpressionOrStatement)
-            ? bindings[key]
-            : iif(bindingExpressionOrStatement),
-        );
+        let binding = bindings[key];
+        if (typeof binding == 'string' && isStatement(binding)) {
+          binding = iif(binding);
+        }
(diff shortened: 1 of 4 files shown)


steve8708 (approved):

packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -414,10 +422,30 @@ function literalTagName(symbol: string | Symbol): string | Symbol {
...
+function isExpression(code: string) {
+  const identifiers = code.split(EXPRESSION_SEPARATORS);
+  return identifiers.filter((ident) => ident && !ident.match(EXPRESSION_IDENTIFIER)).length == 0;
+}
+
+// https://regexr.com/6cpk4
+const EXPRESSION_SEPARATORS = /[()\[\]{}.\-+/*,]+/;
+
+// https://regexr.com/6cpka
+const EXPRESSION_IDENTIFIER = /^\s*[a-zA-Z_$][a-zA-Z0-9_$]*\s*$/;
```
steve8708, line 451: if I read this right it may not handle the case of ternaries (but maybe easy to add)

it can be common for a binding value to be like `state.foo ? 'bar' : 'baz'`

also is common to have strings in there too which can have any character, which is why we took the prior approach in the past

```
state.foo ? 'I have lots of special characters !@#$*&")O*(' : 'bar'
```

mhevery: OK, let me fix that.

mhevery: fixed

PR #158, opened 2022-01-19
Fix: strip `this.` in react-native

samijaber:
Changes:

- Strip `this.` for state objects in react-native
- Add symlinking
- move to node 16
- some tidying up


packages/core/src/generators/react.ts
```diff
@@ -212,6 +212,17 @@ const getRefsString = (json: MitosisComponent, refs = getRefs(json)) => {
...
+/**
+ * Removes all `this.` references.
+ */
+const stripThisRefs = (str: string, options: ToReactOptions) => {
+  if (options.stateType !== 'useState') {
+    return str;
+  }
+
+  return str.replaceAll('this.', '');
+};
```
samijaber, line 224: These were causing pages to crash, because we had dead `this.` references still floating around after we flattened the state object into the render body. See https://github.com/BuilderIO/builder/pull/717/files#diff-69d8b0d6908ff898e0ea908897a6ead6c77d96e3473d2e9dc75aa6d10c36b56aL15-L23

mhevery (approved):

PR #159, opened 2022-01-19
fix(qwik): improve `isStatement` heuristics

mhevery:


packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -453,9 +453,11 @@ export function isStatement(code: string) {
   }
   code = code.replace(STRING_LITERAL, 'STRING_LITERAL');
   const identifiers = code.split(EXPRESSION_SEPARATORS);
-  return (
-    identifiers.filter((i) => i && !i.match(EXPRESSION_IDENTIFIER)).length !== 0
-  );
+  const filteredIdentifiers = identifiers.filter((i) => {
+    i = i.trim();
+    return i && !i.match(EXPRESSION_IDENTIFIER);
```
samijaber, line 458: Small detail: I try to stray away from mutation, specially mutating function arguments. It's fine in this simple case, but we do a lot of mutations to append & build complex values, which I find hard to track.

What do you think?

```suggestion
    const trimmedI = i.trim();
    return trimmedI && !trimmedI.match(EXPRESSION_IDENTIFIER);
```

I might enable https://eslint.org/docs/rules/no-param-reassign on this repo 😁

samijaber (approved):

packages/core/src/generators/qwik/src-generator.ts
mhevery: yes, agree. Let me change that.

mhevery: fixed

PR #163, opened 2022-01-22
Add on update lifecycle method

sahilmob:
This PR brings the `onUpdate` lifecycle hook to Mitosis components; for example, it will be replaced with a `useEffect` hook with its dependencies array if the compilation target is Reactjs, it will also be replaced with the `updated` hook for vue and so on.

### How does it work?
It works by simply adding a `useEffect` hook with its dependencies in the dependencies array **(must)** _or_ by omitting the dependencies array.

```jsx
import {  useEffect } from "@builder.io/mitosis";

  // with deps. array
  useEffect(() => {
    console.log("on mount");
  }, [dep1, dep2,...etc]);
 
  // without deps. array
  useEffect(() => {
    console.log("on update");
  });

  return (
    <div>
     ...
    </div>
  );
}

```

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
@@ -989,7 +989,6 @@ Object {
   "hooks": Object {
     "onMount": Object {
       "code": "",
-      "deps": "[]",
     },
   },
   "imports": Array [],
--- a/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
@@ -241,8 +241,6 @@ Object.assign(state, {
     }
   },
 });
-
-state.findAndRunScripts();
 ",
     "tsCode": "var props = state;
 
@@ -292,7 +290,7 @@ useState({
   },
 });
 
-onMount(() => {
+onUpdate(() => {
   state.findAndRunScripts();
 });
 ",
@@ -353,7 +351,10 @@ export default function MyComponent(props) {
   const elem = useRef();
 
   onMount(() => {
-    state.findAndRunScripts();
+    var props = state;
+    onUpdate(() => {
+      state.findAndRunScripts();
+    });
   });
 
   return (
@@ -438,8 +439,6 @@ Object.assign(state, {
     }
   },
 });
-
-state.findAndRunScripts();
 ",
     "tsCode": "var props = state;
 
@@ -489,7 +488,7 @@ useState({
   },
 });
 
-onMount(() => {
+onUpdate(() => {
   state.findAndRunScripts();
 });
 ",
@@ -550,7 +549,10 @@ export default function MyComponent(props) {
   const elem = useRef();
 
   onMount(() => {
-    state.findAndRunScripts();
+    var props = state;
+    onUpdate(() => {
+      state.findAndRunScripts();
+    });
   });
 
   return <div class=\\"builder-embed\\" ref={elem} innerHTML={props.content} />;
@@ -679,29 +681,6 @@ Object.assign(state, {
     return !!props.lazy && state.isBrowser();
   },
 });
-
-if (state.useLazyLoading()) {
-  // throttled scroll capture listener
-  const listener = () => {
-    if (pictureRef) {
-      const rect = pictureRef.getBoundingClientRect();
-      const buffer = window.innerHeight / 2;
-
-      if (rect.top < window.innerHeight + buffer) {
-        state.load = true;
-        state.scrollListener = null;
-        window.removeEventListener(\\"scroll\\", listener);
-      }
-    }
-  };
-
-  state.scrollListener = listener;
-  window.addEventListener(\\"scroll\\", listener, {
-    capture: true,
-    passive: true,
-  });
-  listener();
-}
 ",
     "tsCode": "var props = state;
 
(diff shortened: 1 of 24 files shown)


steve8708:
this is awesome - thanks for the PR @sahilmob ! @samijaber will take a look shortly. also feel free to join us in discord to talk about what you are up to with Mitosis and more we can collaborate on! we have a #mitosis room over here https://discord.com/channels/842438759945601056/935218469799071835/935569066666651722

steve8708:
@samijaber just want to bump on your radar!

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -298,7 +296,7 @@ useState({
   },
 });
 
-onMount(() => {
+onUpdate(() => {
```
steve8708, line 299: I think in most of these examples, onMount is actually what we want. e.g. in this specific case we definitely want this to run only once on first mount and never again

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -359,7 +357,10 @@ export default function MyComponent(props) {
   const elem = useRef();
 
   onMount(() => {
-    state.findAndRunScripts();
+    var props = state;
+    onUpdate(() => {
+      state.findAndRunScripts();
+    });
```
steve8708, line 363: looks like this is double nested but right now hooks need to be all top level only (no hooks within hooks)

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -556,7 +555,10 @@ export default function MyComponent(props) {
...
```
steve8708, line 561: similar issue of hooks within hooks

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -785,28 +764,31 @@ export default function MyComponent(props) {
...
+        const listener = () => {
+          if (pictureRef.current) {
+            const rect = pictureRef.current.getBoundingClientRect();
+            const buffer = window.innerHeight / 2;
+
+            if (rect.top < window.innerHeight + buffer) {
+              state.load = true;
+              state.scrollListener = null;
+              window.removeEventListener(\\"scroll\\", listener);
+            }
```
steve8708, line 780: another hook within hook we don't want

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -1787,22 +1775,12 @@ Object {
...
-  .then((data) => {
-    state.reviews = data.data;
-  });
 ",
     "tsCode": "var props = state;
 
 useState({ reviews: [], showReviewPrompt: false });
 
-onMount(() => {
+onUpdate(() => {
```
steve8708, line 1783: this is another one we definitely want to be onMount and not unUpdate

packages/core/src/__tests__/__snapshots__/react.test.ts.snap
```diff
@@ -155,7 +159,7 @@ export default function CustomCode(props) {
 
   useEffect(() => {
     findAndRunScripts();
-  }, []);
+  }, undefined);
```
samijaber, line 162: This change indicates that something is broken. I think that instead of printing `${json.hooks.onUpdate.deps}` in the React generator, we want to make sure to handle the empty case:

```
${json.hooks.onUpdate.deps || '[]'}
```

packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
```diff
@@ -579,31 +571,6 @@ export default function Image(props) {
...
-      };
-
-      state.scrollListener = listener;
-      window.addEventListener(\\"scroll\\", listener, {
-        capture: true,
-        passive: true,
-      });
-      listener();
-    }
-  });
```
samijaber, line 605: Kind of odd that this change caused the `onMount` parts of the snapshot test to completely disappear. That doesn't look right to me. cc @steve8708  not sure if you've got any ideas on this 🤔

samijaber (changes requested):
@sahilmob Looking good! I found some issues worth looking into.

Additionally, could you add usage of the `onUpdate` lifecycle method to the snapshot tests? That way we'll have snapshot instances where it is working as intended, and can confirm it's being generated properly for all frameworks.

steve8708:
This is really great stuff @sahilmob 

I added some comments, and looking deeper I think the confusion is from all of our code examples using `onMount` what we want is for that to run only once, so the equivalent of `mounted() {}` in vue or `useEffect(() => , [])` in React

I think we may want to avoid using the name useEffect for a Mitosis hook as it may confuse people, as we don't support things like returning a function to run on unmount. I think for now sticking to conventions that vuejs and solidjs use may be best, where we can have explicit `onMount` and `onUpdate`

maybe, for this purposes of this PR
1. keep the onMount behavior the same
2. rename `useEffect` to `onUpdate` which in the output frameworks runs after every render, or optionally can take a dependencies array (basically the same thing you have here, just with a new name consistent with our other hooks)

what are your thoughts @sahilmob ?

sahilmob:
@steve8708 I agree 100%. The reason why I went with the `useEffect` is that originally, it compiles to the `mounted` hook if the target is vue for example, so I thought that we could depend on the existence of the dependencies inside the dependencies array of the `useEffect` to figure out whether its an `onMount` or an `onUpdate` hook.

steve8708:
This is awesome @sahilmob! Thanks for the amazing work

PR #164, opened 2022-01-26
Fix: react-native MQs

samijaber:
MQs are not supported in react-native. Long-term solution is to add support (see https://github.com/kasinskas/react-native-media-query). For now, we add a console warning and strip them out when generating the `StyleSheet`.

### DIFF
--- a/packages/core/src/__tests__/data/blocks/columns.raw.tsx
+++ b/packages/core/src/__tests__/data/blocks/columns.raw.tsx
@@ -46,6 +46,12 @@ export default function Column(props: ColumnProps) {
         flexDirection: 'column',
         alignItems: 'stretch',
         lineHeight: 'normal',
+        '@media (max-width: 999px)': {
+          flexDirection: 'row',
+        },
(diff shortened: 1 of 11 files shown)


packages/core/src/__tests__/data/blocks/columns.raw.tsx
```diff
@@ -46,6 +46,12 @@ export default function Column(props: ColumnProps) {
         flexDirection: 'column',
         alignItems: 'stretch',
         lineHeight: 'normal',
+        '@media (max-width: 999px)': {
+          flexDirection: 'row',
+        },
+        '@media (max-width: 639px)': {
+          flexDirection: 'row-reverse',
+        },
```
samijaber, line 54: I added these MQs to one snapshot to make sure that they do _not_ appear in the react-native jest snapshots. They indeed don't 👍🏽

steve8708 (approved):

adamdbradley (approved):

PR #165, opened 2022-01-27
fix(qwik): Fix Symbol bindings

mhevery:
- Symbols did not support bindings from parent
- It was not possible to have client based Components
- Created `Image` as a client side component so that it can render
  correctly under different bindings.

### DIFF
--- a/packages/core/src/generators/qwik/component.ts
+++ b/packages/core/src/generators/qwik/component.ts
@@ -1,4 +1,3 @@
-import { omit } from 'lodash';
 import {
   compileAwayBuilderComponentsFromTree,
   components as compileAwayComponents,
@@ -84,10 +83,10 @@ export function addComponent(
   opts: { isRoot?: boolean; shareStyles?: boolean } = {},
 ) {
   const _opts = { isRoot: false, shareStyles: false, ...opts };
-  compileAwayBuilderComponentsFromTree(
-    component,
-    omit(compileAwayComponents, ['Image']),
-  );
+  compileAwayBuilderComponentsFromTree(component, {
+    ...compileAwayComponents,
+    Image: undefined!,
+  });
   const componentName = component.name;
   const handlers = renderHandlers(
     fileSet.high,
--- a/packages/core/src/generators/qwik/component.ts
+++ b/packages/core/src/generators/qwik/component.ts
@@ -1,3 +1,4 @@
+import { omit } from 'lodash';
 import {
   compileAwayBuilderComponentsFromTree,
   components as compileAwayComponents,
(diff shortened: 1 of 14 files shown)


packages/core/src/generators/qwik/component.ts
```diff
@@ -82,7 +84,10 @@ export function addComponent(
   opts: { isRoot?: boolean; shareStyles?: boolean } = {},
 ) {
   const _opts = { isRoot: false, shareStyles: false, ...opts };
-  compileAwayBuilderComponentsFromTree(component, compileAwayComponents);
+  compileAwayBuilderComponentsFromTree(
+    component,
+    omit(compileAwayComponents, ['Image']),
+  );
```
samijaber, line 90: I am slightly weary of pulling in `lodash` dependencies given how notorious it is for increasing bundle size.

For a relatively simple use-case like this one, how do you feel about omitting using the spread operator?

```suggestion
  const { Image, ...compileAwayComponentsWithoutImage } = compileAwayComponents
  compileAwayBuilderComponentsFromTree(
    component,
    compileAwayComponentsWithoutImage,
  );
```

packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -573,6 +563,29 @@ export const compileAwayBuilderComponents = (
...
 
+function updateQueryParam(uri = '', key: string, value: string) {
+  const re = new RegExp('([?&])' + key + '=.*?(&|$)', 'i');
+  const separator = uri.indexOf('?') !== -1 ? '&' : '?';
+  if (uri.match(re)) {
+    return uri.replace(re, '$1' + key + '=' + encodeURIComponent(value) + '$2');
+  }
+
+  return uri + separator + key + '=' + encodeURIComponent(value);
+}
```
samijaber, line 574: Curious why `updateQueryParam` is duplicated here and in `qwik/directives.ts`. They look identical to me. 

What do you think of sharing the helper between both places to reduce duplication?

samijaber (approved):

packages/core/src/symbols/symbol-processor.ts
```diff
@@ -0,0 +1,176 @@
...
+
+const createBuilderElement = (
+  options?: Partial<BuilderElement>,
+): BuilderElement => ({
+  '@type': '@builder.io/sdk:Element',
+  id:
+    'builder-' +
+    Math.random()
+      .toString(36)
+      .split('.')[1],
```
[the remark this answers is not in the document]
mhevery: fixed

packages/core/src/plugins/compile-away-builder-components.ts
samijaber: I see, makes sense. Thanks for explaining 👍🏽

PR #166, opened 2022-01-31
fix(qwik): When props and bindings are same assume string literal

mhevery:


samijaber (approved):

packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -285,7 +285,11 @@ export class SrcBuilder {
         }
         this.isJSX ? this.emit(key) : this.emit(possiblyQuotePropertyName(key));
         this.isJSX ? this.emit('={') : this.emit(':', WS);
-        if (typeof binding == 'string' && isStatement(binding)) {
+        if (binding === props[key]) {
+          // HACK: workaround for the fact that sometimes the `bindings` have string literals
+          // We assume that when the binding content equals prop content.
+          binding = JSON.stringify(binding);
```
steve8708, line 291: oh this just means invalid code, we don't want to assume it's a string. it's always possible someone has written invalid code in a binding, how we check for this in other places is essentially like

```tsx
try { 
  babel.parse(expression)
  return true;
} catch (err) {
  // This is invalid code, do not forward it to the generated code
  return false;
}
```

so in short, if we detect invalid code, we want this to be ignored entirely as this is what would happen at runtime too using our other SDKs as well as the visual editor

PR #178, opened 2022-02-15
Eng 2253

mhevery:


samijaber (approved):

packages/core/src/symbols/symbol-processor.ts
```diff
@@ -99,7 +99,7 @@ export function convertBuilderContentToSymbolHierarchy(
         if (collectComponentState) {
           const symbol: BuilderSymbol = el.component.options.symbol;
```
samijaber, line 100: Out of curiosity, which one of these 2 statements is true, causing the `symbol.content.data.state` to not exist?

- `symbol` is not always a `BuilderSymbol`
- `symbol` is always a `BuilderSymbol`, but the latter is not guaranteed to have `content.data.state` every time.

If it's the former, we could add a user-made type-guard. If it's the latter, we can update the type to be nullable. Trying to think of solutions to prevent this problem from happening again in the future!

steve8708: both can be missing. symbols don't always have content provided, and not all content always has data (the latter is more rare, but not impossible)

right now our current typings I believe are valid. BuilderSymbol type says content can be null. and BuilderContent type says that data can be null. in this case though, we aren't using the typings, but that helps us not have to import those types from the react SDKs specifically. we could shorthand them here tho if we really want, right now they are all `any` type in the code

steve8708: ah, I just realized BuilderSymbol is already shorthand here. yeah we should make clear in that type that content and content.data are nullable

Issue #161, opened 2022-01-21
webcomponents: how to use [shadow] in mitosis?


jackieyoho:

steve8708:
ah yes, not supported currently but that wouldn't be too hard to add - e.g. as an option for `componentToCustomElement`, pull requests welcome!

https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/html.ts#L573

PatrickJS:
maybe something like this?
https://github.com/BuilderIO/mitosis/pull/179/files

PR #179, opened 2022-02-16
feat(componentToCustomElement): shadow option

PatrickJS:
initial code for https://github.com/BuilderIO/mitosis/issues/161


steve8708:
love it!

packages/core/src/generators/html.ts
```diff
@@ -643,6 +649,12 @@ export const componentToCustomElement = (
           `
           }
 
+          ${
+            !hasShadowConfig
+              ? ''
+              : `this._shadowRoot = this.attachShadow({ mode: '${useOptions.shadow}' });`
```
PatrickJS, line 655: if mode is `"closed"` the shadowRoot is returned and we can maintain the reference. If `"open"` is selected `this.shadowRoot` will be available.

Issue #181, opened 2022-02-22
Missing "index" in React output when using "For"


edolix:
Hey there! This project looks _amazing_ ❤️ !

Trying it out i see that the React output with `For` is missing the `index` argument. See [this test Fiddle](https://mitosis.builder.io/?outputTab=react&code=import+%7B+useState%2C+For+%7D+from+%27%40builder.io%2Fmitosis%27%0A%0Aexport+default+function+List%28%29+%7B%0A++const+state+%3D+useState%28%7B%0A++++myArray%3A+%5B1%2C+2%2C+3%5D%2C%0A++++list%3A+%5B%27hello%27%2C+%27world%27%5D%2C%0A++%7D%29%0A%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%3CFor+each%3D%7Bstate.myArray%7D%3E%0A++++++++%7B%28theArrayItem%2C+index%29+%3D%3E+%3Cdiv+key%3D%7Bindex%7D%3E%7BtheArrayItem%7D%3C%2Fdiv%3E%7D%0A++++++%3C%2FFor%3E%0A++++%3C%2Fdiv%3E%0A++%29%0A%7D%0A).

Note: this is for testing purposes only, i know that using `key={index}` is a bad idea but it's enough to see the issue.

> Input

```ts
import { useState, For } from '@builder.io/mitosis'

export default function List() {
  const state = useState({
    myArray: [1, 2, 3],
    list: ['hello', 'world'],
  })

  return (
    <div>
      <For each={state.myArray}>
        {(theArrayItem, index) => <div key={index}>{theArrayItem}</div>}
      </For>
    </div>
  )
}
```

> React Output

```ts
import { useState } from "react";

export default function List(props) {
  const [myArray, setMyArray] = useState(() => [1, 2, 3]);

  const [list, setList] = useState(() => ["hello", "world"]);

  return (
    <div>
      {myArray?.map((theArrayItem) => (
        <div key={index}>{theArrayItem}</div>
      ))}
    </div>
  );
}

```

samijaber:
@edolix Thanks for raising this issue! We have noticed this bug internally as well. I appreciate the simple test-case, that makes it easier to reproduce. 👍🏽 

A PR contribution to fix this issue would be welcome if you're feeling up for it!

PR #185, opened 2022-02-26
Add missing "index" in React generator when using For

edolix:
Hey! I tried to follow other patterns for this fix, just _always_ add the `index` argument. 

Closes https://github.com/BuilderIO/mitosis/issues/181

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/react-native.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/react-native.test.ts.snap
@@ -80,6 +80,8 @@ export default function Column(props) {
       {props.columns?.map((column, index) => (
         <View style={styles.view2}>
           <Text>{column.content}</Text>
+
+          <Text>{index}</Text>
         </View>
       ))}
     </View>
@@ -438,7 +440,7 @@ export default function FormComponent(props) {
       {props.builderBlock && props.builderBlock.children ? (
         <>
           {props.builderBlock?.children?.map((block, index) => (
-            <BuilderBlockComponent key={block.id} block={block} />
+            <BuilderBlockComponent key={block.id} block={block} index={index} />
           ))}
         </>
       ) : null}{\\" \\"}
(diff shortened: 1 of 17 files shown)


steve8708 (approved):
Excellent, thanks @edolix !

packages/core/src/__tests__/__snapshots__/react-native.test.ts.snap
```diff
@@ -77,7 +77,7 @@ export default function Column(props) {
 
   return (
     <View style={styles.view1}>
-      {props.columns?.map((column) => (
+      {props.columns?.map((column, index) => (
```
samijaber, line 80: Could you edit this snapshot test's input component so that it relies on `index`? That way our tests will guarantee that:
- the `index` value is exposed as the second argument
- the `index` value is not being stripped out wherever we try to consume it

samijaber (commented):
Amazing, thanks @edolix! Just wanted to add some testing coverage for usage of `index` and we're all set

packages/core/src/__tests__/__snapshots__/react-native.test.ts.snap
edolix: Sure, i'll make this change for each test 👍

edolix: Sorry @samijaber the delay 🙏  - I went a bit _aggressive_ on https://github.com/BuilderIO/mitosis/pull/185/commits/f626f03990ea6651aabeeaa4c3515d19f0be40eb but let me know if works for you!
Thanks!

samijaber (approved):
Great work. Thanks for contributing this fix 🏆

packages/core/src/__tests__/__snapshots__/react-native.test.ts.snap
samijaber: Thanks @edolix, looks good to me!

samijaber:
@edolix can you run `prettier`? It's not yet automated on this repo, and is complaining. You can do so with `npm run fmt` from the root of the mono-repo.

edolix:
@samijaber yup - apologies! Should be GTG now 🤞

PR #219, opened 2022-03-17
Pr hotfix

mhevery:
This is a hotfix for a broken homepage. 

The issue is that because code that is lazy loaded does not have its state correctly restored fetching state results in an error. This fix does not generate a state restoration code if no one is referring to the state.

### DIFF
--- a/packages/core/src/generators/qwik/handlers.ts
+++ b/packages/core/src/generators/qwik/handlers.ts
@@ -62,10 +62,13 @@ export function renderHandlers(
 }
 
 function renderHandler(file: File, symbol: string, code: string) {
+  const body = [wrapWithUse(file, code)];
+  const shouldRenderStateRestore = code.indexOf('state') !== -1;
+  if (shouldRenderStateRestore) {
+    body.unshift(renderUseLexicalScope(file));
+  }
   file.exportConst(symbol, function (this: SrcBuilder) {
-    this.emit([
-      arrowFnBlock([], [renderUseLexicalScope(file), wrapWithUse(file, code)]),
-    ]);
+    this.emit([arrowFnBlock([], body)]);
   });
   file.src.emit(NL);
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "prettier": "^1.19.1",
     "typescript": "^4.2.3",
     "@babel/preset-env": "^7.6.0",
+    "esbuild": "0.14.25",
```
samijaber, line 39: I might have been the one to remove this dependency! Can you point me to what it was missing from? 👀

packages/core/src/generators/qwik/handlers.ts
```diff
@@ -62,10 +62,13 @@ export function renderHandlers(
 }
 
 function renderHandler(file: File, symbol: string, code: string) {
+  const body = [wrapWithUse(file, code)];
+  const shouldRenderStateRestore = code.indexOf('state') !== -1;
```
samijaber, line 66: **❔ question:** ‏What can `code` be in this scenario? Is it possible for it to be something like `someVarContainingTheWordstate`? Would that cause a false positive here?

samijaber (approved):

package.json
mhevery: I forget to add it. It was needed for the https://github.com/BuilderIO/mitosis-build but I did not see it in PR because the code only gets excersized on merge.  It is. needed here: https://github.com/BuilderIO/mitosis/blob/main/packages/core/package.json#L44

packages/core/src/generators/qwik/handlers.ts
mhevery: Yes, it would, but it is fine as generating the above code would just create a variable that would not be used. This is a bit of a hack to fix our homepage.

package.json
samijaber: I see! Thanks for explaining.

Issue #216, opened 2022-03-15
CHORE: inconsistent Prettier config


samijaber:
Our prettier config is inconsistent across the monorepo. We have different versions installed in core, cli and the root. This is causing contributors' code editors to `formatOnSave` in an incorrect way, which breaks CI and forces them to run the command again before being able to merge.

Steps to reproduce:

1. edit and save a file in `packages/core` or `packages/cli` (using VSCode or some other editor that has formatOnSave functionality)
2. run `npm run fmt` from the root

Expected output: both steps format files using the same version of prettier

A working solution will likely entail removing all `prettier` versions from the `packages/*`, and relying only on the `prettier` version in the root `package.json`

PR #220, opened 2022-03-17
chore(build): Developer Improvements (Fix Root Dependencies + Update Prettier Config)

Harkunwar:
## Description
At the root of our project, we're making use of several dependencies in our scripts however we're not including them in `devDependencies`. This PR aims to fix that, as well as unify the prettier config across the mono repo (#216). Also updated the `developer.md` to use `npm run setup` to bootstrap all the `lerna` dependencies.

### Next issues to cover post merge
 - Fix eslint config
 - Fix stylelint


Dkendal (approved):
Thanks for the contribution @Harkunwar!

Issue #215, opened 2022-03-15
Custom Elements: Updating element class


ja-rivas:
**Describe the bug**
The class of an element isn't updated correctly when generating a customElement. Inside the `update` method, the class is updated using the `el.class` property instead of `el.className`. For example:
```js
this.querySelectorAll("[data-name='div-1']").forEach((el) => {
  el.class = this.state.name === "Steve" ? "bg-success" : "bg-primary";
});
```

**To Reproduce**
1. Create a mitosis file in which an element class's value depends on a variable (e.g. `<div class={backgroundClass}>`).
2. Compile the file to a customElement.
3. Note that in the `update` method the class value is being set by `el.class = value`.

Fiddle link: https://mitosis.builder.io/?outputTab=webcomponents&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+%5Bname%2C+setName%5D+%3D+useState%28%22Steve%22%29%3B%0A%0A++return+%28%0A++++%3Cdiv+class%3D%7Bname+%3D%3D%3D+%27Steve%27+%3F+%27bg-success%27+%3A+%27bg-primary%27%7D%3E%0A++++++Hello%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D

**Expected behavior**
The element class should be updated using: `el.className = newValue;`.

**Screenshots**
Generated file with the issue:
![image](https://user-images.githubusercontent.com/8509401/158478900-74c05545-8e39-4328-b6c3-f3fff2afd307.png)

steve8708:
great catch, simplest fix is likely to update [this](https://github.com/BuilderIO/jsx-lite/blob/fix-unmount-hook/packages/core/src/generators/html.ts#L304:L304)

to be 

```ts
const useKey = key === 'class' ? 'className' : key;
...
: `;el.${key} = ${useValue}`,
```

PatrickJS:
oh this is fixed for webcomponents

Issue #208, opened 2022-03-15
Custom Elements: Properties with dashes in their name


ja-rivas:
**Describe the bug**
When compiling to custom elements, if an element property is bind to a variable and that property name has dashes, the binding fails because the property name is incorrectly used in the `el.setAttribute` method.

**To Reproduce**
1. Create a jsx mitosis file that has a binding to a property with dashes. For example:
```jsx
<div data-user={name}></div>
```
2. Compile the mitosis file to customElements
3. Note that inside the `update` method of the customElement, the name of the property is incorrectly used in the `el.setAttribute` method. It's missing the quotes. For example:
```js
this.querySelectorAll("[data-name='div-1']").forEach((el) => {
  el.setAttribute(data - user, this.state.name);
});
```

Fiddle with bug:
https://mitosis.builder.io/?outputTab=webcomponents&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+%5Bname%2C+setName%5D+%3D+useState%28%22Steve%22%29%3B%0A%0A++return+%28%0A++++%3Cdiv+data-user%3D%7Bname%7D%3E%0A++++++%3Cinput%0A++++++++css%3D%7B%7B%0A++++++++++color%3A+%22red%22%2C%0A++++++++%7D%7D%0A++++++++value%3D%7Bname%7D%0A++++++++onChange%3D%7B%28event%29+%3D%3E+setName%28event.target.value%29%7D%0A++++++%2F%3E%0A++++++Hello%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D

**Expected behavior**
The name of the property in the setAttribute call should be inside quotes and without spaces:
```js
this.querySelectorAll("[data-name='div-1']").forEach((el) => {
  el.setAttribute("data-user", this.state.name);
});
```

steve8708:
In case helpful: [here](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/html.ts#L303:L303) is probably where to look for the fix

PR #224, opened 2022-03-17
Fix: Addresses attributes with - in the string and creates a map for special cases like class vs className

alujs:
#215 
#208


packages/core/src/generators/html.ts
```diff
@@ -40,6 +40,30 @@ type InternalToHtmlOptions = ToHtmlOptions & {
...
 
+const attributeKeyExceptionsMap: { [key: string]: string } = {
+  class: 'className',
+};
+
+const updateKeyIfException = (key: string): string => {
+  return attributeKeyExceptionsMap[key] ?? key;
+};
+const normalizeAttributeKey = (key: string): string => {
+  return `"${key.replace(/\s/g, '')}"`;
```
samijaber, line 51: I'd drop the string interpolation since it's not needed:

```suggestion
  return key.replace(/\s/g, '');
```

packages/core/src/generators/html.ts
```diff
@@ -40,6 +40,30 @@ type InternalToHtmlOptions = ToHtmlOptions & {
   namesMap: NumberRecord;
 };
 
+const attributeKeyExceptionsMap: { [key: string]: string } = {
```
samijaber, line 43: **🎯 suggestion:** ‏I'd suggest we fully capitalize constants like this map:
```suggestion
const ATTRIBUTE_KEY_EXCEPTIONS_MAP: { [key: string]: string } = {
```

samijaber (approved):
Nice @alujs 🎉

packages/core/src/generators/html.ts
alujs: Just going over what we talked about. It turns out that the prettier plugin is the culprit so we don't need the replace anymore, however we do need the ""  somewhere so we removed the helper function and put it directly on the string itself.

packages/core/src/generators/html.ts
alujs: Updated based on feedback.

samijaber (approved):

PR #226, opened 2022-03-18
stencil generator v0

steve8708:
Adds initial support for Stencil as an output. This should likely be preferred for non-trival webcomponents

Loom walking through the entire process: https://www.loom.com/share/3e8bf7e667314a829a0b5d4b61c58cc0

~Most things are implemented, the things to add when needed will be~
- ~`onUpdate` hook support (should be easy)~ 
- ~`onUnMount` hook support (very easy)~ 
- ~`ref` support (should be easy)~

**Update**: added support for onUpdate, onUnMount, and refs: https://www.loom.com/share/b86313913d094fcca1e160162d6b982b

Screen Shot 2022-03-18 at 12 23 30 PM

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -108,7 +108,7 @@ _Mitosis input_
 export default function MyComponent() {
   const state = useState({
     doSomethingAsync(event) {
-      void (async function() {
+      void (async function () {
         const response = await fetch(); /* ... */
       })();
     },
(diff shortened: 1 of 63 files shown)


samijaber (approved):
🏆 

For the CLI, we'll need a follow-up PR after publishing `core` to add stencil there too

Issue #210, opened 2022-03-15
Custom Elements: Missing imports from the mitosis file


ja-rivas:
**Describe the bug**
The imports that are used in the original mitosis file aren't moved to the generated file when compiling to `customElement`.

**To Reproduce**
1. Create a mitosis file that has imports at the top of the file (e.g. importing constants, functions, etc. from another file).
2. Compile it to a customElement.
3. Note that the imports are missing in the generated file

Fiddle link: https://mitosis.builder.io/?outputTab=webcomponents&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0Aimport+%7B+myConstants+%7D+from+%22.%2Fpage.constants.js%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+%5Bname%2C+setName%5D+%3D+useState%28%22Steve%22%29%3B%0A++%0A++onMount%28%28%29+%3D%3E+%7B%0A++++name+%3D+myConstants%3B%0A++%7D%29%3B%0A%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%3Cinput%0A++++++++css%3D%7B%7B%0A++++++++++color%3A+%22red%22%2C%0A++++++++%7D%7D%0A++++++++value%3D%7Bname%7D%0A++++++++onChange%3D%7B%28event%29+%3D%3E+setName%28event.target.value%29%7D%0A++++++%2F%3E%0A++++++Hello%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D

**Expected behavior**
The imports of the mitosis file should be put in the generated file.

**Screenshots**
![image](https://user-images.githubusercontent.com/8509401/158475008-0b6097c3-c020-4743-bbd2-4cc888ad8703.png)

steve8708:
Ah great catch - to solve thise you will want to add `renderPreComponent(json)` like [this](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/react.ts#L543) over [here](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/html.ts#L615)

PR #227, opened 2022-03-18
Fix: Imports for webcomponents is now fixed.

alujs:
#210

### DIFF
--- a/packages/core/src/__tests__/data/blocks/stamped-io.raw.tsx
+++ b/packages/core/src/__tests__/data/blocks/stamped-io.raw.tsx
@@ -1,6 +1,7 @@
 import { useState, onMount, For, Show } from '@builder.io/mitosis';
 import { kebabCase } from 'lodash';
 import { snakeCase } from 'lodash';
+import { reduce } from 'lodash';
 
 type SmileReviewsProps = {
   productId: string;
@@ -8,13 +9,16 @@ type SmileReviewsProps = {
 };
 
 export default function SmileReviews(props: SmileReviewsProps) {
-  kebabCase('testThat');
-  snakeCase('testThis');
-
   const state = useState({
     reviews: [] as any[],
     name: 'test',
     showReviewPrompt: false,
+    kebabCaseValue() {
+      return kebabCase('testThat');
+    },
+    snakeCaseValue() {
+      return snakeCase('testThis');
+    },
   });
 
   // TODO: allow async function here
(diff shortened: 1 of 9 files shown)


packages/core/src/__tests__/data/blocks/stamped-io.raw.tsx
```diff
@@ -1,11 +1,16 @@
...
   apiKey: string;
 };
 
 export default function SmileReviews(props: SmileReviewsProps) {
+  kebabCase('testThat');
+  snakeCase('testThis');
+
   const state = useState({
     reviews: [] as any[],
     name: 'test',
```
samijaber, line 16: **🔨 issue:** ‏You'll notice that in the vue, react & react-native tests, these lines of code don't appear, only the imports. This is due to https://github.com/BuilderIO/mitosis/blob/main/packages/eslint-plugin/docs/rules/no-var-declaration-or-assignment-in-component.md

Can you put these somewhere in the `state` instead so it's valid Mitosis?

```suggestion
  const state = useState({
    reviews: [] as any[],
    name: 'test',
    a: kebabCase('testThat'),
    b: snakeCase('testThis'),
```

packages/core/src/__tests__/data/blocks/stamped-io.raw.tsx
```diff
@@ -1,11 +1,16 @@
...
+import { snakeCase } from 'lodash';
 
 type SmileReviewsProps = {
   productId: string;
   apiKey: string;
 };
 
 export default function SmileReviews(props: SmileReviewsProps) {
+  kebabCase('testThat');
+  snakeCase('testThis');
```
samijaber, line 12: Can you also add a 3rd lodash import that is _not_ used anywhere? Some of our generators are smart enough to strip that out: would be good to have that seen in the snapshots!

samijaber (changes requested):

packages/core/src/generators/html.ts
```diff
@@ -797,5 +800,6 @@ export const componentToCustomElement =
     if (options.plugins) {
       str = runPostCodePlugins(str, options.plugins);
     }
-    return str;
+
+    return dedent(renderPreComponent(json)) + '\n' + str;
```
samijaber, line 804: (had written a comment about, this but somehow it didn't make it in the first review!)

One tiny issue: ordering matters in our generators. We want `renderPreComponent(json)` to run within the first instance of `str` and to `dedent` at that point too. The reason for that is to stay consistent with how our other generators work. See:

https://github.com/BuilderIO/mitosis/blob/fdebe412793fd9f1438282b9cc72e5782504fd66/packages/core/src/generators/angular.ts#L175-L180

https://github.com/BuilderIO/mitosis/blob/fdebe412793fd9f1438282b9cc72e5782504fd66/packages/core/src/generators/mitosis.ts#L172-L186

samijaber (approved):

PR #234, opened 2022-03-22
remove unneeded dependencies and files

sachinraja:
## Description
I noticed mitosis [had quite a big install size](https://packagephobia.com/result?p=%40builder.io%2Fmitosis) (153 MB). This PR removes unneeded library dependencies. Most of them were just moved to `devDependencies` instead. I don't think this will be that big of a decrease but it's a start. 

Note: I think the biggest issue with the package size is that a user has to install the dependencies for every framework even if they're only compiling to one, would be nice if the generators were separated into different packages.


samijaber (approved):
👍🏽

Issue #228, opened 2022-03-18
Feature matrix


samijaber:
We currently do not have feature parity across all frameworks. We should put together a simple feature matrix, where each column is a `target` (Angular, React, React-Native, Vue, etc.), and each row is a feature.

For now, a simple labelling system would work:
 🏗 in-progress/partially implemented 
❌ not implemented
✅ implemented 

Here is a non-comprehensive example list of rows:
- lifecycle methods (each would be a separate row): onUnMount, onMount, onUpdate
- dynamic styles
- state
- refs
- Show/For/etc. (Mitosis-specific components)
- children

An easy way to identify support for a feature is to use the [fiddle](https://mitosis.builder.io/) and compare the outputs.

We can start with a partial list, and then start adding more as we notice discrepancies

samijaber:
Closed by #247

kylecordes:
E2E testing should validate the features that work across N targets. The initial fully-supported set (features that work "everywhere") might be pretty small, but should grow nicely over time.

PR #247, opened 2022-03-26
docs: the bone of feature matrix draft

initdc:
Commitor: initdc

## Description
 
Add a short description of:
ref:  (#228)

**add feature matrix draft, I provide two types.
Type B will be better when edit, 
type A really need a huge-wide screen to edit, but it provide better rendering
I prefer type B**


samijaber:
@initdc Thanks for getting this draft started! I also prefer type B, so let's go with that for now :)

PatrickJS:
can we get this in so others can also update it

PatrickJS:
can we get this in so others can also update it

samijaber (approved):

PR #252, opened 2022-03-29
fix(qwik): Support events on components; Support `hide`/`show` bindings

mhevery:


packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -254,26 +251,42 @@ export class SrcBuilder {
...
           binding = JSON.stringify(binding);
         } else if (typeof binding == 'string' && isStatement(binding)) {
           binding = iif(binding);
         }
-        this.emit(binding);
-        this.isJSX ? this.emit('}') : this.emit(',');
+        if (key === 'hide' || key === 'show') {
+          let [truthy, falsy] =
+            key == 'hide' ? ['"none"', '"inherit"'] : ['"inherit"', '"none"'];
+          emitJsxProp('style', function (this: SrcBuilder) {
```
Dkendal, line 264: Why pass `this`, rather than a variable here?

samijaber (approved):

packages/core/src/generators/qwik/src-generator.ts
samijaber: Curious about this as well!

mhevery: Because the `emit` function is declared as:
```
export type EmitFn = (this: SrcBuilder) => void;
```

Perhaps not the best choice, but keep it consistent.

PR #254, opened 2022-03-30
Fix(webcomponents): issue of bindings/state being missed during update()

alujs:
Refactor update function to run user updated code first, then run internal querySelector updates next.

Refactor connectedCallback to run onMounted before update.

### DIFF
--- a/packages/core/src/generators/html.ts
+++ b/packages/core/src/generators/html.ts
@@ -743,7 +743,7 @@ export const componentToCustomElement =
           this.updateBindings();
         }
 
-        updateBindings {
+        updateBindings() {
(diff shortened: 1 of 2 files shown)


packages/core/src/generators/html.ts
```diff
@@ -723,30 +726,37 @@ export const componentToCustomElement =
...
-              const value = useOptions.onChangeJsById[key];
-              if (!value) {
-                return '';
-              }
-              return `
+          this.onUpdate();
+          this.updateBindings();
+        }
+
+        updateBindings {
```
samijaber, line 746: I think you want this to be a function:

```suggestion
        updateBindings() {
```

samijaber (approved):

PR #256, opened 2022-03-30
docs(components): remove this and prefer state

PatrickJS:
we shouldn't show any code using `this` since it can be confusing since the compile code is correct even though the dev code is not correct

onMount(() => {
  this.something
})

^ this is a problem

onMount(() => {  <- because of this arrow function `this` should be the global context which is a JavaScript gotcha
  this.something
})

the code is compiled correctly but can be confusing and I don't want devs to refactor their code thinking they are fixing a js gotcha 

onMount(function () {  <- because of this arrow function `this` should be the global context which is a js gotcha
  this.something
})

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -23,7 +23,10 @@ export default function MyComponent(props) {
     newItemName: 'New item',
     list: ['hello', 'world'],
     addItem() {
-      this.list = [...this.list, this.newItemName];
+      state.list = [
+        ...state.list,
+        state.newItemName
+      ];
     },
   });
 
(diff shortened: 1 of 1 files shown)


samijaber (approved):
Fully agree. I think the `this` references are simply outdated docs. Thanks for catching this!

PatrickJS:
Do you nave auto merge for docs?

samijaber:
@PatrickJS we do not have that setup at the moment unfortunately. 

BTW, this PR is failing because of a `prettier` check on the markdown files. You can fix that by running `npm run fmt` from the root

PR #262, opened 2022-04-04
read generator config from mitosis.config.js

### LINKED ISSUE #199: Support reading generator options from `mitosis.config.js` in `mitosis compile`
Similar to how `mitosis build` works: it imports `getMitosisConfig()` and then override defaults:

https://github.com/BuilderIO/mitosis/blob/76bcc9732e213b7e736ee05117e66b5c40c8e787/packages/cli/src/commands/build.ts#L9
https://github.com/BuilderIO/mitosis/blob/76bcc9732e213b7e736ee05117e66b5c40c8e787/packages/cli/src/build/build.ts#L28-L55

Place to apply changes for `mitosis compile`: 

https://github.com/BuilderIO/mitosis/blob/76bcc9732e213b7e736ee05117e66b5c40c8e787/packages/cli/src/commands/compile.ts#L24-L27

sahilmob:
fixes #199

### DIFF
--- a/packages/cli/mitosis.config.js
+++ b/packages/cli/mitosis.config.js
@@ -1,0 +1,1 @@
+module.exports = {};
--- a/packages/cli/mitosis.config.js
+++ b/packages/cli/mitosis.config.js
@@ -1,1 +1,0 @@
-module.exports = {};
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -52,7 +52,10 @@ const command: GluegunCommand = {
     const generatorOpts: Partial<{ [K in AllGeneratorOptionKeys]: any }> = {
       ...generatorOptions,
       prettier: opts.prettier ?? true,
-      plugins: plugins,
+      plugins: [
+        ...plugins,
+        ...generatorOptions.plugins
+      ],
       format: opts.format,
       prefix: opts.prefix,
       includeIds: opts.includeIds,
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -47,7 +47,7 @@ const command: GluegunCommand = {
     }
 
     const mitosisConfig = getMitosisConfig();
-    const generatorOptions = mitosisConfig.options?.[to];
+    const generatorOptions = mitosisConfig?.options?.[to];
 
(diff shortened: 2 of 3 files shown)


packages/cli/mitosis.config.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
samijaber, line 1: Why is this file needed? 🤔

sahilmob: Without this file tests won't pass because `require(process.cwd() + '/mitosis.config')` throws if `mitosis.config` doesn't exist.

The other solution is to wrap all the logic inside `getMitosisConfig` with a try catch: 

```js
import { MitosisConfig } from '@builder.io/mitosis';

export function getMitosisConfig(): MitosisConfig | null {
  try {
    const module = require(process.cwd() + '/mitosis.config');
    return module?.default || module || null;
  } catch (_) {
    return null
  }
}
```
What do you think?

samijaber: Yeah, I think we should change the logic in `getMitosisConfig()` to check for the existence of the file before importing it. You can use `fs.existsSync(fileName)` for that

packages/cli/src/commands/compile.ts
```diff
@@ -45,17 +46,19 @@ const command: GluegunCommand = {
       plugins.push(compileAwayBuilderComponents());
     }
 
+    const mitosisConfig = getMitosisConfig();
+    const generatorOptions = mitosisConfig?.options?.[to];
+
     const generatorOpts: Partial<{ [K in AllGeneratorOptionKeys]: any }> = {
+      ...generatorOptions,
       prettier: opts.prettier ?? true,
       plugins: plugins,
```
samijaber, line 55: One small detail I noticed here: we have a `plugins` array that is declared here in the function. We want to merge it with the contents of `generatorOptions.plugins` if there are any. I think that'll look like this:

```suggestion
    const mitosisConfig = getMitosisConfig();
    const generatorOptions = mitosisConfig?.options?.[to];

    const generatorOpts: Partial<{ [K in AllGeneratorOptionKeys]: any }> = {
      ...generatorOptions,
      prettier: opts.prettier ?? true,
      plugins: [
        ...plugins,
        ...generatorOptions.plugins
      ],
```

samijaber (approved):

PR #270, opened 2022-04-06
refactor: remove unused dependency(JSON)

hseoy:
## Description

I found an imported but unused dependency(`JSON`) in `mitosis/packages/core/src/types/mitosis-component.ts` and deleted it.


Dkendal:
Thanks!

PR #273, opened 2022-04-08
Fix ng-slot to ng-content.

alujs:
Changed ng slot to ng content. 

https://angular.io/guide/content-projection - I don't think ng-slot exists. I could be wrong though.


samijaber (commented):
Nice, good catch @alujs. 

Would you be able to add an Angular test that includes children, so that we don't get regressions in the future?

samijaber (approved):

PR #287, opened 2022-04-16
fix(examples): replace npm with yarn

akrabdev:
## Description

Replace `npm` with `yarn` in the readme  setup instructions, because yarn.lock is tracked. This will avoid having yarn.lock and package-lock.json simultaneously which may cause inconsistencies in dependency tree.


samijaber (approved):
Thanks for catching this 👍🏽

Issue #291, opened 2022-04-20
Svelte: support self-referencing component


samijaber:
For a component to recursively reference itself, it needs to call `<svelte:self>` instead of `<component-name>`: https://svelte.dev/tutorial/svelte-self

We need to bake this logic into the Svelte generator, specifically the `blockToSvelte` function: 

https://github.com/BuilderIO/mitosis/blob/fbedeeedd016df678bbc92be3143c272a3d5127c/packages/core/src/generators/svelte.ts#L54

We should be able to compare `json.name` of the _block_ against the `name` of the original root Mitosis JSON, and if they're equal, replace the value with `'svelte:self'`

samijaber:
Example: https://mitosis.builder.io/?outputTab=svelte&code=import+%7B+useState%2C+Show+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%7Bprops.name%7D%0A++++++%3CShow+when%3D%7Bprops.name+%3D%3D%3D+%27Sami%27%7D%3E%0A++++++++%3CMyComponent+name%3D%7B%27Jaber%27%7D+%2F%3E%0A++++++%3C%2FShow%3E%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D

Current output:

```svelte
<script>
  export let name;

</script>

<div>
  {name}

  {#if name === 'Sami'}
    <MyComponent name={'Jaber'} />
  {/if}
</div>
```

expected output:

```svelte
<script>
  export let name;

</script>

<div>
  {name}

  {#if name === 'Sami'}
    <svelte:self name={'Jaber'} />
  {/if}
</div>

```

sanyamkamat:
Fixed by #293

PR #293, opened 2022-04-21
fix(svelte): support self-referencing component (#291)

sanyamkamat:
## Description

Svelte uses `svelte:self` to reference itself recursively. More about it [here](https://svelte.dev/tutorial/svelte-self).

This PR passes the parent component information to all the child components so that if any of children name match the parents, we introduce Svelte's self tag in place of the component name.

### DIFF
--- a/packages/core/src/__tests__/data/blocks/self-referencing-component.raw.tsx
+++ b/packages/core/src/__tests__/data/blocks/self-referencing-component.raw.tsx
@@ -1,0 +1,12 @@
+import { Show } from '@builder.io/mitosis';
+
+export default function MyComponent(props: any) {
+  return (
+    <div>
+      {props.name}
+      <Show when={props.name === 'Batman'}>
+        <MyComponent name={'Bruce Wayne'} />
+      </Show>
+    </div>
+  );
+}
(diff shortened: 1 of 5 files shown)


packages/core/src/__tests__/data/blocks/self-referencing-component.raw.tsx
```diff
@@ -0,0 +1,12 @@
+import { Show } from '@builder.io/mitosis';
+
+export default function MyComponent(props: any) {
+  return (
+    <div>
+      {props.name}
+      <Show when={props.name === 'Batman'}>
+        <MyComponent name={'Bruce Wayne'} />
```
samijaber, line 8: 😄

packages/core/src/__tests__/svelte.test.ts
```diff
@@ -16,4 +17,10 @@ describe('Svelte', () => {
     const output = componentToSvelte()({ component });
     expect(output).toMatchSnapshot();
   });
+
+  test('selfReferencingComponent', () => {
+    const component = parseJsx(selfReferencingComponent);
+    const output = componentToSvelte()({ component });
+    expect(output).toMatchSnapshot();
+  });
```
samijaber, line 25: Can we also add a test case where the self-referencing component receives children? Something like:

```ts
import { Show } from '@builder.io/mitosis';

export default function MyComponent(props: any) {
  return (
    <div>
      {props.name}
      {props.children}
      <Show when={props.name === 'Batman'}>
        <MyComponent name={"Bruce"}>
          <div>Wayne</div>
        </MyComponent>
      </Show>
    </div>
  );
}
```

samijaber (approved):
Welcome to Mitosis! @sanyamkamat 🥇

samijaber (approved):
Excellent @sanyamkamat . Thank you!!! 🎉

PR #301, opened 2022-04-22
Svelte: fixes

### LINKED ISSUE #290: implement Svelte Context API
Similar to how we do it in React: https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/context/react.ts
See https://svelte.dev/tutorial/context-api

we need a `contextToSvelte` that will import the proper helpers. First part is to duplicate the React context generator and swap the import with the `svelte` one. The React and Svelte Context APIs are fairly similar.

samijaber:
## Description

- All: fix bug in context import paths (contexts were not getting imported at all)

Svelte:
- implement Context API
- fix hooks code (it was not being transpiled down from TS to JS)
- strip `this.` prefixes
- support `<svelte:component>` for dynamic component setting

closes #290


packages/core/src/helpers/parsers.ts
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    const body = parseCode(code);
+    return isCodeBodyExpression(body);
+  } catch (e) {
+    return false;
+  }
+}
+
+export const isCodeBodyIdentifier = (body: babel.types.Statement[]) =>
+  body.length == 1 && babel.types.isIdentifier(body[0]);
```
samijaber, line 38: Was going to use these in the Svelte generator, ended up not. Decided to keep them in an extracted helper as they might be helpful later.

PR #303, opened 2022-04-28
Various fixes

samijaber:
## Description

- [refactor(cli): extract buildContextFile](https://github.com/BuilderIO/mitosis/pull/303/commits/b58f66f933c5d7a43254035b9c219d873f64ce6e)
- [fix(cli): preserve file extensions for targets in .ts files](https://github.com/BuilderIO/mitosis/pull/303/commits/8eef4355b551d795ba8f35d4a13e70aaeb2a3704)
- [docs: improve types and document](https://github.com/BuilderIO/mitosis/pull/303/commits/dcb74672b7b5e10668329a9654771125f5d9aacf)
- [fix(react): add fragment for Show root nodes](https://github.com/BuilderIO/mitosis/pull/303/commits/b3a65c2ec6df0a9f3c9c52788bcb4544d93bde03)

### DIFF
--- a/packages/core/src/types/mitosis-node.ts
+++ b/packages/core/src/types/mitosis-node.ts
@@ -4,8 +4,29 @@ export type MitosisNode = {
   '@type': '@builder.io/mitosis/node';
   name: string;
   meta: JSONObject;
+  /**
+   * Key-value store of static values for DOM attributes.
+   * ```js
+   * {
+   *   disabled: false,
+   *   defaultValue: 'initial text`,
+   *   width: 100,
+   * }
+   * ```
+   */
   properties: { [key: string]: string | undefined };
   // TODO: I think in our usage these values are always strings. Make a decision and change to reflect
+  /**
+   * Key-value store of _dynamic_ values for DOM attributes.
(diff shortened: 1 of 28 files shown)


adamdbradley (approved):

packages/core/src/types/mitosis-node.ts
```diff
@@ -4,8 +4,29 @@ export type MitosisNode = {
...
+  /**
+   * Key-value store of static values for DOM attributes.
+   * ```js
+   * {
+   *   disabled: false,
+   *   defaultValue: 'initial text`,
+   *   width: 100,
+   * }
+   * ```
+   */
```
steve8708, line 16: @samijaber this isn't quite right, properties is always string values, aka it's always
```
<... foo="bar"
```

and so the object here would be 

```js
{ defaultValue: "foobar', width: "100" }
```

in the above example that would be if the tag looked like

```
<... defaultValue="foo" width="100"
```

if width is a number, it goes into bindings, because it's value is an expression container

```
<... defaultValue="foo" width={100} />
```

PR #305, opened 2022-04-29
bug fixes for solidjs

steve8708:
- convert style object keys to kebabcase
- HACK: force remove `state.state.`. this needs more looking into but was becoming a rabbit hole and we need to get this delivered
- fix missing `useContext` reference


samijaber (approved):
Neat! Thanks for leaving a trail of comments for future us the next time we iterate on the Solid generator 👍🏽

PR #308, opened 2022-04-30
don't transpile solidjs jsx

steve8708:
similar to vue - we don't want to precompile these so they can compile differently per target (client vs server, etc)

also fixes solidjs index

### DIFF
--- a/packages/core/src/generators/solid.ts
+++ b/packages/core/src/generators/solid.ts
@@ -133,11 +133,11 @@ const blockToSolid = (
   if (json.name === 'For') {
     const needsWrapper = json.children.length !== 1;
     return `<For each={${json.bindings.each}}>
-    {(${json.properties._forName}, index) =>
-      ${needsWrapper ? '<>' : ''}
-        ${json.children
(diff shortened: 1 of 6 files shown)


packages/core/src/generators/solid.ts
```diff
@@ -133,11 +133,11 @@ const blockToSolid = (
   if (json.name === 'For') {
     const needsWrapper = json.children.length !== 1;
     return `<For each={${json.bindings.each}}>
-    {(${json.properties._forName}, index) =>
-      ${needsWrapper ? '<>' : ''}
-        ${json.children
-          .filter(filterEmptyTextNodes)
-          .map((child) => blockToSolid(child, options))}}
+    {(${json.properties._forName}, _index) => {
+      const index = _index();
```
samijaber, line 137:
```suggestion
      // The SolidJS `<For>` component has a special index() signal function.
      // https://www.solidjs.com/docs/latest#%3Cfor%3E
      const index = _index();
```

samijaber (approved):

Issue #306, opened 2022-04-29
Remove Span Tags for <For> and {} : Accessibility issues when using span tags everywhere, also CSS issues.


alujs:
[...]

PR #318, opened 2022-05-03
fix: #306 Remove Span Tags

PatrickJS:
## Description
this is the first step before removing span tags for #306


packages/core/src/__tests__/__snapshots__/html.test.ts.snap
```diff
@@ -351,14 +362,17 @@ exports[`Html Stamped 1`] = `
       });
 
       document.querySelectorAll(\\"[data-name='div-3']\\").forEach((el, index) => {
+        const index = this.getContext(el, \\"index\\");
```
PatrickJS, line 365: we can get even more clever with this to remove walking up the tree. in the generated code we can reference a DOM node that has the values. This is good for now

packages/core/src/generators/html.ts
```diff
@@ -199,17 +204,39 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
   if (json.bindings._text) {
     // TO-DO: textContent might be better performance-wise
-    addOnChangeJs(elId, options, `el.innerText = ${json.bindings._text};`);
+    addOnChangeJs(
+      elId,
+      options,
+      `
+      ${parentScopeVars
+        .filter((scopeVar) => {
+          return new RegExp(scopeVar).test(json.bindings._text as string);
```
PatrickJS, line 213: we can get more clever here but this should be good enough

packages/core/src/generators/html.ts
```diff
@@ -199,17 +204,39 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
+        })
+        .map((scopeVars) => {
+          return `const ${scopeVars} = this.getContext(el, "${scopeVars}")`;
+        })
+        .join('\n')}
+      el.innerText = ${json.bindings._text};`,
```
PatrickJS, line 219: the goal is not to change anything the dev wrote

packages/core/src/generators/html.ts
```diff
@@ -199,17 +204,39 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
+    } --></span>`;
   }
 
   let str = '';
 
   if (json.name === 'For') {
     const itemName = json.properties._forName;
+    const indexName = json.properties._indexName;
+    const collectionName = json.properties._collectionName;
+    const scopedVars: ScopeVars = [
```
PatrickJS, line 233: we can get more clever and create a DOM id to the DOM node with the state then walk up each ref in the DOM

packages/core/src/generators/html.ts
```diff
@@ -222,15 +249,21 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
         }.querySelector('[data-template-for="${elId}"]');
         ${
           options.format === 'class' ? 'this.' : ''
-        }renderLoop(el, array, template, "${itemName}");
+        }renderLoop(el, array, template, ${
+        itemName ? `"${itemName}"` : 'undefined'
+      }, ${indexName ? `"${indexName}"` : 'undefined'}, ${
+        collectionName ? `"${collectionName}"` : 'undefined'
```
PatrickJS, line 255: undefined is fine because any optimizer would fix this

packages/core/src/generators/html.ts
```diff
@@ -607,7 +675,6 @@ export const componentToCustomElement =
       json = runPreJsonPlugins(json, options.plugins);
     }
     const componentHasProps = hasProps(json);
-    replaceForNameIdentifiers(json, useOptions);
```
PatrickJS, line 610: we don't need this anymore for web components

packages/core/src/generators/html.ts
```diff
@@ -309,19 +344,31 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
               ? `this.${fnName} = (event) => {`
               : `function ${fnName} (event) {`
           }
-            ${removeSurroundingBlock(updateReferencesInCode(useValue, options))}
+            ${parentScopeVars
+              .filter((scopeVar) => {
+                return new RegExp(scopeVar).test(content as string);
+              })
+              .map((scopeVars) => {
+                return `const ${scopeVars} = this.getContext(event.currentTarget, "${scopeVars}")`;
```
PatrickJS, line 362: we need the `el` here

packages/core/src/generators/react.ts
```diff
@@ -75,7 +75,11 @@ const NODE_MAPPERS: {
     const wrap = wrapInFragment(json);
     return `{${processBinding(json.bindings.each as string, options)}?.map((${
       json.properties._forName
-    }, index) => (
+    }${json.properties._indexName ? ',' + json.properties._indexName : ''}${
```
PatrickJS, line 78: react hardcoded index for some reason

packages/core/src/__tests__/__snapshots__/html.test.ts.snap
```diff
@@ -335,14 +348,16 @@ exports[`Html Stamped 1`] = `
       document.querySelectorAll(\\"[data-name='for']\\").forEach((el, index) => {
         let array = state.reviews;
         let template = document.querySelector('[data-template-for=\\"for\\"]');
-        renderLoop(el, array, template, \\"review\\");
+        renderLoop(el, array, template, \\"review\\", \\"index\\", undefined);
       });
 
       document.querySelectorAll(\\"[data-name='review']\\").forEach((el, index) => {
+        const review = this.getContext(el, \\"review\\");
         el.key = getContext(el, \\"review\\").id;
```
PatrickJS, line 356: slight problem. I need to remove getContext from wrapping this

packages/core/src/generators/html.ts
PatrickJS: if we wanted to fix this more later we can pass the vars into the functions or keep it inline

packages/core/src/generators/html.ts
```diff
@@ -309,19 +324,31 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
           }
-            ${removeSurroundingBlock(updateReferencesInCode(useValue, options))}
+              ${addScopeVars(
+                parentScopeVars,
+                codeContent,
+                (scopeVar: string) =>
+                  `const ${scopeVar} = ${
+                    options.format === 'class' ? 'this.' : ''
+                  }getContext(event.currentTarget, "${scopeVar}")`,
+              )}
```
samijaber, line 344: Can we add some tests whose snapshots will highlight what this will generate? Our current snapshot tests do not.

packages/core/src/generators/html.ts
```diff
@@ -335,13 +362,32 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
+            `
+            ${addScopeVars(
+              parentScopeVars,
+              useValue as string,
+              (scopeVar: string) =>
+                `const ${scopeVar} = ${
+                  options.format === 'class' ? 'this.' : ''
+                }getContext(el, "${scopeVar}")`,
+            )}
+            ;Object.assign(el.style, ${useValue});`,
```
samijaber, line 374: Same here, re: snapshots that include this generated code

packages/core/src/__tests__/__snapshots__/webcomponent.test.ts.snap
```diff
@@ -167,91 +278,2138 @@ class SmileReviews extends HTMLElement {
...
+      this.renderLoop(el, array, template, \\"item\\", \\"itemIndex\\", undefined);
+    });
+
+    this._root.querySelectorAll(\\"[data-name='img-1']\\").forEach((el, index) => {
+      el.className = \\"custom-class\\";
+
+      var item = this.getContext(el, \\"item\\");
+      el.src = item;
+
+      var item = this.getContext(el, \\"item\\");
```
PatrickJS, line 719: we have to use var because of how the code generation works. I would have to refactor it or do a hacky way to hoist the const

PatrickJS:
can you publish after merge

packages/core/src/generators/html.ts
```diff
@@ -293,19 +308,31 @@ const blockToHtml = (json: MitosisNode, options: InternalToHtmlOptions) => {
...
               : `function ${fnName} (event) {`
           }
-            ${removeSurroundingBlock(updateReferencesInCode(useValue, options))}
+              ${addScopeVars(
+                parentScopeVars,
+                codeContent,
+                (scopeVar: string) =>
+                  `const ${scopeVar} = ${
+                    options.format === 'class' ? 'this.' : ''
+                  }getContext(event.currentTarget, "${scopeVar}");`,
```
samijaber, line 327: @PatrickJS as far as I can tell there is still no test for this change in the snapshots. Mind adding this one as well?

PatrickJS: I updated one of the tests to include this
https://github.com/BuilderIO/mitosis/pull/318/commits/b12d4757539fcc86ee1a87c9ece9fc6b88003e61

samijaber (approved):
Looks good. Thanks for adding all of these snapshot tests 👌🏽

PR #319, opened 2022-05-04
Fix for show/hide span tag issues.

alujs:
Discovered issue where nested Show tags would not behave properly, relying on re-rendering in order to construct the corect view.


packages/core/src/__tests__/__snapshots__/html.test.ts.snap
```diff
@@ -84,28 +84,9 @@ exports[`Html Shadow DOM 1`] = `
         });
 
       document.querySelectorAll(\\"[data-name='show']\\").forEach((el, index) => {
```
PatrickJS, line 86: the problem was how the update function works. if we don't remove previous template elements then this will grab them too and duplicate code

packages/core/src/generators/html.ts
```diff
@@ -732,6 +724,16 @@ export const componentToCustomElement =
...
 
+
+  showContent(el) {
+    const elementFragment = el.content.cloneNode(true);
+    const children = Array.from(elementFragment.childNodes)
+    children.forEach(child => {
+      this.destroyNodes.push(child);
+    });
+    el.after(elementFragment);
+  }
```
samijaber, line 735: As someone not super familiar with cloning elements in this way, I find this logic a little bit convoluted. Could we add some comments explaining the general idea behind this logic, for future reference?

From what I can tell, we are:
- cloning the `content` of content we want to show. Can we link to these docs, so it's clear this is specific to HTML templates? https://developer.mozilla.org/en-US/docs/Web/API/HTMLTemplateElement/content
- adding all its children to an array of nodes that we eventually want to destroy (when we un-mount)
- add this content to the element using `el.after`

PatrickJS: yeah I'll add comments

samijaber (approved):

PR #323, opened 2022-05-08
Fix parser jsx custom event argument name

PatrickJS:
## Description
fix a gotcha

Add a short description of:
- what changes you made, 
  removed the gotcha where `event` is hardcoded
- why you made them, and 
  for custom event names
N/A

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -6,10 +6,10 @@
 - [Gotchas and limitations](#gotchas-and-limitations)
   - [Defining variables with the same name as a state property will shadow it](#defining-variables-with-the-same-name-as-a-state-property-will-shadow-it)
   - [Async methods can't be defined on "state"](#async-methods-cant-be-defined-on-state)
-  - [Callback implicitly have an "event" arg](#callback-implicitly-have-an-event-arg)
+  - [Callback implicitly have an "$event" arg in Angular](#callback-implicitly-have-an-event-arg-in-angular)
   - [Functions can't be passed by reference to JSX callbacks](#functions-cant-be-passed-by-reference-to-jsx-callbacks)
-  - [Can't assign "params" to "state"](#cant-assign-"params"-to-"state")
-  - [Can't assign function output to "state"](#cant-assign-function-output-to-"state")
+  - [Can't assign "params" to "state"](#cant-assign-params-to-state)
+  - [Can't assign function output to "state"](#cant-assign-function-output-to-state)
   - [Can't destructure assignment from state](#cant-destructure-assignment-from-state)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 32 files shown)


packages/core/src/parsers/jsx.ts
```diff
@@ -527,6 +527,18 @@ const jsxElementToJson = (
...
+          if (types.isArrowFunctionExpression(expression)) {
+            if (key.startsWith('on')) {
+              // TODO: we need a better place to store these arguments
+              const args = expression.params.map(
+                (node) => (node as babel.types.Identifier)?.name,
+              );
+              memo['_arguments_' + key] = args;
+            }
+          }
+        }
```
samijaber, line 541: This isn't quite the right place for this logic. I recently documented `properties` vs `bindings`: https://github.com/BuilderIO/mitosis/blob/ccb51aed001181d55030011379175f58c478fea7/packages/core/src/types/mitosis-node.ts#L7-L30

A better approach is to modify the `bindings` signature to look something like:

```ts
bindings: {
  code: string;
  arguments?: string[];
}
```

And move this logic to the `reduce` right below in `bindings`. 

This will also greatly clean up how we consume these arguments in the generators.

You alluded to something similar in https://github.com/BuilderIO/mitosis/issues/326

docs/gotchas.md
```diff
@@ -135,32 +135,44 @@ export default function MyComponent(props) {
 }
 ```
 
-### Callback implicitly have an "event" arg
+### Callback implicitly have an "$event" arg in Angular
```
samijaber, line 138: This caveat should remain general to all generators, since this PR only fixes it for a couple of them. Most of them still rename the argument to `event`

samijaber (changes requested):

packages/core/src/parsers/jsx.ts
PatrickJS: bindings: {
  onClick: string;
  arguments?: string[];
}

my issue is if we have more than one event.

bindings: {
  onClick: string;
  onInput: string
  _arguments_onClick?: string[];
  _arguments_onInput?: string[];
}

After looking into the code I'm thinking I can also store this in meta.

Ideally we want 
bindings: {
  onClick: {
    arguments: string[],
    code: string
  };
}

PatrickJS: if everyone is okay with adding scope to the nodes it would clean up a lot of code and I can just refactor everything to use it

samijaber: Sorry, my types example was missing the index signature! I meant to say the same thing you did:

```ts
bindings: {
  [index: string]: {
    arguments: string[],
    code: string
  };
}
```

> if everyone is okay with adding scope to the nodes 

I personally prefer we keep the change small and located to just the `bindings` properties: that's a similar API/interface to what we've done for `onUpdate` hooks and their `dependencies` array: https://github.com/BuilderIO/mitosis/blob/ccb51aed001181d55030011379175f58c478fea7/packages/core/src/types/mitosis-component.ts#L34

steve8708: I think changing bindings to code & arguments is a good idea. and agreed in theory properties shouldn't need this as it only takes simple string values

PatrickJS:
I'l refactor it to 
```
bindings: {
  [index: string]: {
    arguments: string[],
    code: string
  };
}
```
and remove the docs edit

PatrickJS:
I'm closing this to create an issue https://github.com/BuilderIO/mitosis/issues/351. I'll end up doing this if no one else picks it up.

PR #331, opened 2022-05-10
feat(angular/react/html): onInit

PatrickJS:
## Description
add onInit to react/angular


packages/core/src/generators/angular.ts
```diff
@@ -199,10 +199,10 @@ export const componentToAngular =
         .join('\n')}
 
       ${
-        !component.hooks.onMount
+        !component.hooks.onInit
           ? ''
           : `ngOnInit() {
-              ${stripStateAndPropsRefs(component.hooks.onMount.code, {
+              ${stripStateAndPropsRefs(component.hooks.onInit.code, {
```
samijaber, line 205: We still want to handle `onMount` hooks though, otherwise `onMount` code will compile away for Angular.

Can probably duplicate this code block for both `onInit` and `onMount`.

PatrickJS: I added both

samijaber (approved):

PR #333, opened 2022-05-10
fix(html): show getContext and batch scope vars

PatrickJS:
## Description
fix multiple `getContext` where we have to use `var` keyword.
fix Show not including `getContext`


packages/core/src/generators/react.ts
```diff
@@ -216,10 +237,23 @@ export const blockToReact = (json: MitosisNode, options: ToReactOptions) => {
...
+  if (json.children) {
+    childrenNodes = json.children
+      .map((item) => blockToReact(item, options, needsToRenderSlots))
+      .join('\n');
+  }
+  if (needsToRenderSlots.length) {
+    needsToRenderSlots.forEach(({ key, value }) => {
+      str += ` ${key}={${value}} `;
+    });
+  }
```
samijaber, line 252: Isn't this a no-op until another piece gets implemented? In this PR it'll never run since `needsToRenderSlots` is always empty

samijaber (approved):

PR #336, opened 2022-05-11
Fix: SolidJS

samijaber:
## Description

- Consolidate file extension logic into a few helpers
- Add missing `.jsx` extension for SolidJS
- fix(solid): add missing Show `fallback` prop


packages/core/src/helpers/replace-idenifiers.ts
```diff
@@ -9,9 +9,10 @@ export const replaceIdentifiers = (
   return babelTransformExpression(code, {
     Identifier(path: babel.NodePath<babel.types.Identifier>) {
       if (
-        // This is not a property access - like `foo` in `this.foo`
+        // This is not an (optional) property access - like `foo` in `this.foo` or `this?.foo`
         !(
-          types.isMemberExpression(path.parent) &&
+          (types.isMemberExpression(path.parent) ||
+            types.isOptionalMemberExpression(path.parent)) &&
```
samijaber, line 15: We to `isMemberExpression` checks in many other parts of the codebase, without also checking `isOptionalMemberExpression`. Too reluctant to make a sweeping change as those parsers do a lot of critical work, although it should theoretically be a safe thing to do.

Something to keep in mind down the line...

PR #337, opened 2022-05-11
docs(components): Slot/Children

PatrickJS:
## Description
explaining slot and children

preview link
https://github.com/BuilderIO/mitosis/blob/70f7270c3ea75c47dc3e5e2c616ef3777d9043c5/docs/components.md#children

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -236,7 +236,7 @@ function Layout(props) {
 }
 ```
 
-Mitosis compiles one component at a time and is only concern of outputting the correct method for each framework. In the two example above this is the angular output.
+Mitosis compiles one component at a time and is only concerned with outputting the correct method for each framework. For the two examples above, here are the angular and html outputs.
 
 ```html
 <div>
--- a/docs/components.md
+++ b/docs/components.md
@@ -194,19 +194,18 @@ export default function MyComponent(props) {
 ```
 
 ### Children
+
 We use the standard method for passing children with `props.children`
+
 ```jsx
 function MyComponent(props) {
-  return (
-    <div>
-      {props.children}
-    </div>
-  );
+  return <div>{props.children}</div>;
 }
 ```
 
 ### Slot
-When you want to register a named slot you do so using the `slot` prop similar to `on` convention. 
+
+When you want to register a named slot you do so using the `slot` prop similar to `on` convention.
 
 ```jsx
 <div>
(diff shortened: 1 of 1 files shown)


docs/components.md
```diff
@@ -190,3 +192,81 @@ export default function MyComponent(props) {
...
+    slotCenter={<Content/>}
+  />
+    anything else
+  </Layout>
+</div>
+```
+
+In this example we are registering `top`, `left`, and `center`. for the `Layout` component to project.
+
+If we the `Layout` component was also a Mitosis component then we simply use the reference in the props
```
samijaber, line 224:
```suggestion
If the `Layout` component was also a Mitosis component then we simply use the reference in the props
```

docs/components.md
```diff
@@ -190,3 +192,81 @@ export default function MyComponent(props) {
...
+    slotTop={<NavBar/>}
+    slotLeft={<Sidebar/>}
+    slotCenter={<Content/>}
+  />
+    anything else
+  </Layout>
+</div>
+```
+
+In this example we are registering `top`, `left`, and `center`. for the `Layout` component to project.
```
samijaber, line 222:
```suggestion
In this example we are registering `top`, `left`, and `center` for the `Layout` component to project.
```

docs/components.md
```diff
@@ -190,3 +192,81 @@ export default function MyComponent(props) {
...
+      <div className="top">{props.slotTop}</div>
+      <div className="left">{props.slotLeft}</div>
+      <div className="center">{props.slotCenter}</div>
+      {props.children}
+    </div>
+  );
+}
+```
+
+Mitosis compiles one component at a time and is only concern of outputting the correct method for each framework. In the two example above this is the angular output.
```
samijaber, line 239:
```suggestion
Mitosis compiles one component at a time and is only concerned with outputting the correct method for each framework. For the two examples above, here are the angular and html outputs.
```

docs/components.md
```diff
@@ -190,3 +192,81 @@ export default function MyComponent(props) {
...
+
+```jsx
+function MyComponent(props) {
+  return <div>{props.children}</div>;
+}
+```
+
+### Slot
+
+When you want to register a named slot you do so using the `slot` prop similar to `on` convention.
```
samijaber, line 208: what do you mean by the "`on` convention"? I don't quite get that.

samijaber (approved):

PR #345, opened 2022-05-13
feat(angular): outputs via metadata

PatrickJS:
## Description

outputs via metadata for now and I'll update the detector later so we won't need useMetadata

```
import { useState, useMetadata, onMount } from '@builder.io/mitosis';

useMetadata({
  outputs: ['onMessage', 'onEvent'],
});
export default function MyBasicOutputsComponent(props: any) {
  const state = useState({
    name: 'PatrickJS',
  });

  onMount(() => {
    props.onMessage(state.name);
    props.onEvent(props.message);
  });

  return <div></div>;
}
```

### DIFF
--- a/packages/core/src/generators/angular.ts
+++ b/packages/core/src/generators/angular.ts
@@ -27,17 +27,22 @@ export interface ToAngularOptions extends BaseTranspilerOptions {}
 
 interface AngularBlockOptions {
   contextVars?: string[];
+  outputVars?: string[];
 }
 
 const mappers: {
-  [key: string]: (json: MitosisNode, options: ToAngularOptions) => string;
+  [key: string]: (
+    json: MitosisNode,
+    options: ToAngularOptions,
+    blockOptions?: AngularBlockOptions,
+  ) => string;
 } = {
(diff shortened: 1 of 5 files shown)


packages/core/src/generators/angular.ts
```diff
@@ -200,6 +219,10 @@ export const componentToAngular =
     });
 
     const props = getProps(component);
+    // remove props for outputs
+    outputVars.forEach((variableName) => {
+      props.delete(variableName);
```
PatrickJS, line 224: removing function props

packages/core/src/generators/angular.ts
```diff
@@ -261,6 +290,7 @@ export const componentToAngular =
         .map((refName) => `@ViewChild('${refName}') ${refName}: ElementRef`)
         .join('\n')}
 
+      ${dataString}
```
PatrickJS, line 293: it's better to have these defined before the lifecycle methods

samijaber (approved):

steve8708:
neat idea

PR #348, opened 2022-05-15
Update README to rely less on LLVM context

kylecordes:
This revision still uses the LLVM analogy extensively, but with more hints
to clarify for developers not familiar with LLVM.

Tried some other tweaks also.

Hopefully useful, feel free to reject without offense if you prefer the original!


steve8708 (approved):
This is great, thanks for sending @kylecordes!

PatrickJS:
Nice

PR #364, opened 2022-05-19
fix(qwik): properly handle undefined values

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/symbols/symbol-processor.ts
+++ b/packages/core/src/symbols/symbol-processor.ts
@@ -65,7 +65,7 @@ interface BuilderSymbol {
   data: Record<string, any>;
   content: {
     data: {
-      state: Record<string, any>;
+      state?: Record<string, any>;
     };
   };
 }


PatrickJS (approved):

packages/core/src/symbols/symbol-processor.ts
```diff
@@ -100,9 +100,11 @@ export function convertBuilderContentToSymbolHierarchy(
           const symbol: BuilderSymbol = el.component.options.symbol;
           const props = symbol.data;
           const state = symbol.content.data.state;
-          const id = toHash(state);
-          props['serverStateId'] = id;
-          collectComponentState[id] = state;
+          if (state) {
```
samijaber, line 103: Should we improve the type of `state` above? It's currently not nullable. Might be better to change it to:


```ts
interface BuilderSymbol {
  data: Record<string, any>;
  content: {
    data: {
+++     state?: Record<string, any>;
    };
  };
}
```

This will codify this knowledge into the types and avoid similar errors in other places

mhevery: Good idea! Fixed!

Issue #351, opened 2022-05-17
core: update MitosisComponent.bindings to include arguments 


PatrickJS:
```
bindings: {
  [index: string]: {
    arguments: string[],
    code: string
  };
}
```
in order to allow devs to change the event name we need to get the arguments from jsx and rewrite the code in the generator.

This PR is an example of that without arguments 
https://github.com/BuilderIO/mitosis/pull/323

decadef20:
I want to have a try

PatrickJS:
@decadef20 if you look at my PR half the code is not needed because it's guarding against overloading props.   https://github.com/BuilderIO/mitosis/pull/323#discussion_r868124062

in the code here https://github.com/BuilderIO/mitosis/blob/ccb51aed001181d55030011379175f58c478fea7/packages/core/src/types/mitosis-component.ts#L34
add `arguments` which is a type `string[]`


once we get that PR merged we can do the rest of the PR which is storing the arguments then using it in the generated code

alujs:
@decadef20 

他希望你先做 PR，然后他会 merge。

decadef20:
@PatrickJS @alujs 
抱歉 我误解了@PatrickJS
Sorry I misunderstood @PatrickJS 
When I finish, I will make a PR

PatrickJS:
done via https://github.com/BuilderIO/mitosis/pull/369

PR #369, opened 2022-05-19
refactor(MitosisComponent): refactor binding type

decadef20:
## Description

Add a short description of:
- what changes you made, 
 refactor binding 
```js
bindings: {
    [key: string]: string | undefined;
  }

```
```ts
bindings: {
    [key: string]:
      | { code: string | undefined; arguments?: string[] }
      | undefined;
  }
```

- why you made them, and  
   [core: update MitosisComponent.bindings to include arguments ](https://github.com/BuilderIO/mitosis/issues/351)

### DIFF
--- a/packages/core/src/types/mitosis-node.ts
+++ b/packages/core/src/types/mitosis-node.ts
@@ -27,6 +27,10 @@ export type MitosisNode = {
    * }
    * ```
    */
-  bindings: { [key: string]: string | undefined };
+  bindings: {
+    [key: string]:
+      | { code: string | undefined; arguments?: string[] }
+      | undefined;
+  };
   children: MitosisNode[];
 };
--- a/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
@@ -199,9 +199,15 @@ Object {
         "@type": "@builder.io/sdk:Element",
         "actions": Object {},
         "bindings": Object {
-          "class": "'builder-custom-code' + (props.replaceNodes ? ' replace-nodes' : '')",
-          "innerHTML": "props.code",
-          "ref": "elem",
+          "class": Object {
+            "code": "'builder-custom-code' + (props.replaceNodes ? ' replace-nodes' : '')",
+          },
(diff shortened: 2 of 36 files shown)


PatrickJS:
@samijaber we probably do need to extend MitosisNode bindings. For binding like `_spread`, `$`, or `on` which is make it easier to add a type prop here too. Maybe just end up making a MitosisBinding

samijaber:
@decadef20 The build is failing:

https://github.com/BuilderIO/mitosis/runs/6508592405?check_suite_focus=true

It looks like you still need to migrate a handful of cases where `MitosisComponent.bindings` is being used

decadef20:
> 

Thanks. I am trying to get familiar with the codebase as soon as possible.

Errors are fixed.

PatrickJS:
LGTM

PatrickJS (approved):

packages/core/src/types/mitosis-node.ts
```diff
@@ -27,6 +27,10 @@ export type MitosisNode = {
    * }
    * ```
    */
-  bindings: { [key: string]: string | undefined };
+  bindings: {
+    [key: string]:
+      | { code: string | undefined; arguments?: string[] }
+      | undefined;
+  };
```
samijaber, line 34: This type information isn't quite accurate. the `code` is always going to be defined. the `undefined` possibility bubbles up to the overall object now:

```suggestion
  bindings: {
    [key: string]:
      | { code: string; arguments?: string[] }
      | undefined;
  };
```

If we make this change, then a lot of type casts that were added in this PR become unnecessary.

@decadef20 If possible, it would be great to make a follow-up PR that makes this improvement!

decadef20: Thanks for your CR. I'll fix it later

decadef20: image

I change the `binding` type, and delete type cast,  but compile fails. 

`binding[key]` maybe `undefined`,  It seems like we have to use type cast here.

PatrickJS: If it's only thats the only place where there's a problem then you can change it

Issue #376, opened 2022-05-20
webcomponent: useRef


PatrickJS:
useRef in html which is just a querySelector

Issue #374, opened 2022-05-20
webcomponent: useContext


PatrickJS:
useContext for Web Component

allow devs to specify what they want to useContext otherwise default to an html version

PR #377, opened 2022-05-20
fix(webcomponent): useRef

PatrickJS:
## Description
webcomponent 

- useRef
- useContext for webcomponent
- setContext for webcomponent
- hydrate stateful DOM nodes
- rename data-name to data-el
- setContext pass code to object
- useContext allow for strings
- onUpdate with deps 

angular

- fix angular event name replacement

fixes https://github.com/BuilderIO/mitosis/issues/376
fixes https://github.com/BuilderIO/mitosis/issues/378
fixes https://github.com/BuilderIO/mitosis/issues/374
fixes https://github.com/BuilderIO/mitosis/issues/395
fixes https://github.com/BuilderIO/mitosis/issues/380
fixes https://github.com/BuilderIO/mitosis/issues/416

I tested that useContext/setContext for webcomponent works in the internal app


packages/core/src/__tests__/__snapshots__/webcomponent.test.ts.snap
```diff
@@ -224,6 +224,181 @@ customElements.define(\\"my-basic-child-component\\", MyBasicChildComponent);
...
+"/**
+ * Usage:
+ *
+ *  <my-masic-ref-component></my-masic-ref-component>
+ *
+ */
+class MyMasicRefComponent extends HTMLElement {
+  get inputRef() {
+    return this._root.querySelector(
+      \\"[data-ref='MyMasicRefComponent-inputRef']\\"
```
PatrickJS, line 237: might be losing context here

PatrickJS: Actually if ref is used in a function it's lost since this is state object

PatrickJS:
I need to rebase

packages/core/src/parsers/jsx.ts
```diff
@@ -185,6 +185,8 @@ const componentFunctionToJson = (
               const valueNode = expression.arguments[1];
               setContext[keyPath] = {
                 name: keyNode.name,
+                // allow for ref injector
+                ref: generate(valueNode).code,
```
PatrickJS, line 189: I need a way to get the code for when the user is referencing a factory function to create the context

packages/core/src/types/mitosis-component.ts
```diff
@@ -49,7 +49,13 @@ export type MitosisComponent = {
...
   context: {
     get: { [key: string]: ContextInfo };
-    set: { [key: string]: { name: string; value?: JSONObject } }; // TODO: support non object values
+    set: {
+      [key: string]: {
+        name: string;
+        value?: JSONObject;
+        ref?: any;
+      };
+    }; // TODO: support non object values
```
PatrickJS, line 58: ref is a workaround to support non-objects

PatrickJS:
@decadef20 when you are available can you review this PR?

packages/core/src/__tests__/__snapshots__/webcomponent.test.ts.snap
```diff
@@ -224,6 +288,433 @@ customElements.define(\\"my-basic-child-component\\", MyBasicChildComponent);
...
+    this.myService = this.getContext(this._root, MyService);
+
+    // used to keep track of all nodes created by show/for
+    this.nodesToDestroy = [];
+    // batch updates
+    this.pendingUpdate = false;
+
+    // Event handler for 'input' event on input-my-basic-component-1
+    this.onInputMyBasicComponent1Input = (event) => {
+      this.state.onChange;
```
PatrickJS, line 331: I will find a way to fix this later

packages/core/src/__tests__/__snapshots__/html.test.ts.snap
```diff
@@ -55,6 +56,9 @@ exports[`Html Basic 1`] = `
       if (el?.scope) {
         textNode.scope = el.scope;
       }
+      if (el?.context) {
```
decadef20, line 59: `renderTextNode` is never used here.

PatrickJS: I'll update `componentToHtml` can you check `componentToCustomElement`

PatrickJS: ok I fixed it https://github.com/BuilderIO/mitosis/pull/377/commits/de9698e9b396dbd05fe8294bbc5bb00ba502d538

decadef20 (approved):
LGTM

packages/core/src/parsers/jsx.ts
samijaber: This is causing us to duplicate context data every time. I think we should store either `value` or `ref`, depending on what `valueNode` is.

Pseudocode:

```ts
if (valueNode) {
  if (types.isObjectExpression(valueNode)) {
    return { value: parseStateObject(valueNode) }
  } else {
    return { ref: generate(valueNode).code }
  }
} else {
  return undefined
}
```

And the type of Context would then look something like:

```ts
type ContextSet = { name: string; } & ({ value: JSONObject } | { ref: any } | undefined);
```

samijaber (dismissed):

packages/core/src/parsers/jsx.ts
PatrickJS: yeah I'll do that until we figure out a better way to handle this

packages/core/src/parsers/jsx.ts
```diff
@@ -183,15 +183,21 @@ const componentFunctionToJson = (
...
-                // allow for ref injector
-                ref: generate(valueNode).code,
-                value:
-                  valueNode && types.isObjectExpression(valueNode)
-                    ? parseStateObject(valueNode)
-                    : undefined,
-              };
+              if (valueNode) {
+                if (types.isObjectExpression(valueNode)) {
+                  const value = parseStateObject(valueNode) as JSONObject;
```
samijaber, line 188: Isn't this cast unnecessary?

```suggestion
                  const value = parseStateObject(valueNode);
```

packages/core/src/parsers/jsx.ts
```diff
@@ -183,15 +183,21 @@ const componentFunctionToJson = (
...
+                  setContext[keyPath] = {
+                    name: keyNode.name,
+                    value,
+                  };
+                } else {
+                  const ref = generate(valueNode).code as string;
```
samijaber, line 194: Same here, unnecessary cast:

```suggestion
                  const ref = generate(valueNode).code;
```

packages/core/src/__tests__/__snapshots__/react.test.ts.snap
```diff
@@ -1069,12 +1119,20 @@ export default function MultipleOnUpdateWithDeps(props) {
...
-    console.log(\\"Runs once every update/rerender\\");
-  }, []);
+"import { useState } from \\"react\\";
 
-  return <div />;
+export default function OnInit(props) {
+  const [name, setName] = useState(() => \\"\\");
+
+  setName(defaultValues.name || props.name);
+  console.log(\\"set defaults with props\\");
```
samijaber, line 1128: Pretty sure your changes to `onInit` logic broke it: you're now setting state on every single render.

packages/core/src/generators/react.ts
```diff
@@ -635,17 +641,17 @@ const _componentToReact = (
           : ''
       }
       ${getContextString(json, options)}
-      ${getRefsString(json)}
       ${getInitCode(json, options)}
 
       ${
         json.hooks.onInit?.code
-          ? `useEffect(() => {
+          ? `${!onInitLocalVars ? '' : ';(function () {'}
```
samijaber, line 648: I don't really understand this change fully, but I know that removing `useEffect` breaks the implementation of `onInit`, as you're now running the code on every single render, not just on init.

Wanna revert this change and move it to a separate PR so it's not blocking the rest of this PR?

samijaber (changes requested):
Everything for context looks good now, but you broke `onInit` for React.

I suggest you move that out if you want to get the rest of the code merged without blocking it.

packages/core/src/generators/react.ts
PatrickJS: onMount is preferred. onInit is now only meant to work around some of the problems with useState. we can't set default values with props so this is the only place to do it

packages/core/src/parsers/jsx.ts
PatrickJS: `parseStateObject` returns `any`

packages/core/src/generators/react.ts
PatrickJS: https://github.com/BuilderIO/mitosis/issues/388#issuecomment-1136027557

PatrickJS: I put it under an experimental flag

packages/core/src/generators/react.ts
```diff
@@ -635,17 +642,29 @@ const _componentToReact = (
...
       }
       ${getContextString(json, options)}
-      ${getRefsString(json)}
       ${getInitCode(json, options)}
 
       ${
         json.hooks.onInit?.code
-          ? `useEffect(() => {
+          ? `${
+              options?.experimental?.localOnInit
```
samijaber, line 650: Just to re-iterate: having side-effect code run in the `render` body of a React component is a very bad idea, and not what we want to do. It's dangerous and should be avoided. Re-renders will cause your component to misbehave in many ways, and they will never work.

If I copy the snapshot example you had in the previous commit into Stackblitz, it crashes due to infinite re-renders: https://stackblitz.com/edit/react-ts-jisvkk?file=App.tsx

image


The `render` body should always be a pure function without side effects. Any side effect should be moved inside of a React hook of some kind. In the case of wanting to solve https://github.com/BuilderIO/mitosis/issues/388, we'd want to figure out a way to move the state initialization code inside of `useState(initValue)`.

On one hand, I'm OK keeping this code in since it's behind an experimental flag, on the other, it won't ever work as we expect so I'm not sure why we'd want to keep it

PatrickJS: we can change onInit to be useEffect with no deps until useState is refactored first.
(in this case being able to set `initValue` of `useState(initValue)` because we want to set default values in Mitosis useState).

onInit should only exist until Mitosis handles everything correctly 

The end goal is going towards having local variables and the dev not having to setup memo/useMemo/useCallback (perf step required by every react dev)

The original component was trying to avoid using `useState` but had to because of the current Mitosis limits

packages/core/src/generators/react.ts
```diff
@@ -520,6 +523,9 @@ const _componentToReact = (
 
   mapRefs(json, (refName) => `${refName}.current`);
 
+  const onInitLocalVars =
+    json.hooks?.onInit?.code && /var|const|let/.test(json.hooks?.onInit?.code);
+
```
samijaber, line 528: unused code

```suggestion
```

samijaber (approved):

PR #402, opened 2022-05-24
doc: support custom event name

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fixed by #383


PatrickJS:
I think @samijaber wanted to keep this still

samijaber (approved):
I think since we merged the #383, this is no longer relevant, so we can remove it 👍🏽

PR #408, opened 2022-05-26
fix(solid): dynamic this binding

samijaber:
## Description

Rendering dynamic tags in SolidJS causes a crash. Regression from #369

### DIFF
--- a/packages/core/src/generators/solid.ts
+++ b/packages/core/src/generators/solid.ts
@@ -40,7 +40,7 @@ function processDynamicComponents(
   traverse(json).forEach((node) => {
     if (isMitosisNode(node)) {
       if (node.name.includes('.')) {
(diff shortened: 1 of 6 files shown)


Dkendal (approved):
bye bye type casts

packages/core/src/generators/solid.ts
```diff
@@ -40,7 +40,7 @@ function processDynamicComponents(
   traverse(json).forEach((node) => {
     if (isMitosisNode(node)) {
       if (node.name.includes('.')) {
-        (node.bindings.component as { code: string }).code = node.name;
+        node.bindings.component = { code: node.name };
```
samijaber, line 43: This was the actual line causing the solidjs generator to crash.

PR #403, opened 2022-05-24
Feature/duplicate class attributes in output

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 
- any other context that you think might be helpful for someone to better understand what is contained in this pull request. 

This sort of information is useful for people reviewing the code, as well as anyone from the future trying to understand why changes were made or why a bug started happening.
 fix #282

PR #414, opened 2022-05-27
Fix: consolidate `class` & `className`

samijaber:
## Description

Reverts https://github.com/BuilderIO/mitosis/pull/403, which partially addressed the issue in some generators/situations, but not all.

- consolidates `class` & `className` properties/bindings in the JSX parser by merging `className` into `class` (or removing it if it's a dynamic binding). Now, the Mitosis JSON only has `class`, and never `className`.
- removes all references to `className` properties/bindings in generators (now that the Mitosis JSON only uses `class`

### DIFF
--- a/packages/core/src/parsers/jsx.ts
+++ b/packages/core/src/parsers/jsx.ts
@@ -741,6 +741,12 @@ function mapReactIdentifiers(json: MitosisComponent) {
         }
         delete item.properties.className;
       }
+
+      if (item.properties.class && item.bindings.class) {
+        console.warn(
+          `[${json.name}]: Ended up with both a property and binding for 'class'.`,
+        );
+      }
     }
   });
 }
--- a/packages/core/src/parsers/jsx.ts
+++ b/packages/core/src/parsers/jsx.ts
@@ -715,18 +715,31 @@ function mapReactIdentifiers(json: MitosisComponent) {
             item.bindings[key]!.arguments = value.arguments;
           }
         }
-        if (key === 'className') {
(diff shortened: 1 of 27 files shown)


PatrickJS (approved):

packages/core/src/parsers/jsx.ts
```diff
@@ -715,18 +715,37 @@ function mapReactIdentifiers(json: MitosisComponent) {
...
         }
+        delete item.bindings.className;
       }
-      for (const key in item.properties) {
-        if (key === 'class') {
-          const currentValue = item.properties[key];
-          delete item.properties[key];
-          item.properties.class = currentValue;
+
+      if (item.properties.className) {
```
PatrickJS, line 733: 🙌

PatrickJS (approved):

PatrickJS:
we need a linter for this

samijaber:
> we need a linter for this

To say that we prefer `class` over `className`? Or which part exactly

PatrickJS:
```
// TO-DO: it's too much work to merge 2 bindings, so just remove the old one for now.
```
^ when devs run into this case if there is a className binding and class bindings.

actually couldn't we solve this via 
```jsx
item.bindings.class = `( ${item.bindings.className} ) + ( ${item.bindings.class} )`;
```
since the expression has to resolve into a string or just too many edge cases

samijaber:
I didn't want to think through and account for all the edge cases of how to combine 2 bindings, so I left it out for now as it's an obscure edge case anyways.

PR #419, opened 2022-05-28
fix image lazy loading binding and escaping quotes for vue

steve8708:


packages/core/package.json
```diff
@@ -25,7 +25,7 @@
   "engines": {
     "npm": "99999999.9.9"
   },
-  "version": "0.0.56-11",
+  "version": "0.0.56-14",
```
steve8708, line 28: fyi @samijaber I tried again to publish but still can't

I'm running `yarn release:dev` from `packages/core`

first I had an issue with no npm auth token, so I added one. then I just get a 404 (says package not found in yarn registry). can you add some instructions somewhere on what the release flow should be (or point me to where they are if they do exist currently)?

packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -450,12 +451,22 @@ export const components: CompileAwayComponentsMap = {
...
-          muted: { code: node.bindings.muted?.code },
-          controls: { code: node.bindings.controls?.code },
-          playsinline: { code: node.bindings.playsInline?.code },
-          loop: { code: node.bindings.loop?.code },
+          poster: node.bindings.posterImage?.code && {
+            code: node.bindings.posterImage?.code,
+          },
+          autoplay: node.bindings.autoPlay?.code && {
+            code: node.bindings.autoPlay?.code,
+          },
```
steve8708, line 459: fyi this got broken in a recent contribution PR, didn't get the logic right here. we can add more tests to catch this in the future, the snapshot coverage of this logic looks to be mostly non existent so easy to fix at some point

packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -328,15 +328,14 @@ export const components: CompileAwayComponentsMap = {
       name: 'img',
       properties: noUndefined({
         $name: 'image',
-        loading: node.properties.lazy ? 'lazy' : undefined,
+        loading: 'lazy',
```
steve8708, line 331: fyi @mhevery the logic here wasn't right, fixed now tho. ultimately we should really move this and [this](https://github.com/builderio/mitosis/blob/native-10/packages/core/src/generators/qwik/directives.ts#L64:L136) (which I just discovered we had) to be generated by Mitosis to avoid easy human errors of implementation of these components

packages/core/package.json
samijaber: The only step to release is to `yarn release:dev`, so it must be an auth issue, which would explain the "package not found" issue you're seeing. It's hard for me to know why that's breaking on your end without pairing.

I believe I used `yarn npm login` to authenticate myself on npm. I did not add an npm auth token of any kind. Have you tried that? https://yarnpkg.com/cli/npm/login

packages/core/src/generators/vue.ts
```diff
@@ -35,6 +35,10 @@ import { BaseTranspilerOptions, TranspilerArgs } from '../types/config';
 import { GETTER } from '../helpers/patterns';
 import { methodLiteralPrefix } from '../constants/method-literal-prefix';
 
+function encodeQuotes(string: string) {
+  return string.replace(/"/g, '&quot;');
+}
```
samijaber, line 40: The Vue snapshots didn't update for this change. Can you give an example of what it's meant to fix, or what issue you ran into?

samijaber (approved):

packages/core/src/generators/vue.ts
steve8708: yeah, if you have something like `bindings: { foo : `'"'` }` you'd get `:foo="'"'"` which is broken, this makes sure it is now `:foo="'&quot;'"`

we should ultimately update the other generators with the same logic

Issue #372, opened 2022-05-20
Preserve Typescript types in output


CKGrafico:
Hello not sure if I'm understanding well the purpose of this library but when I create a component like:
```
export type ButtonProps = {
  name: string;
};

export default function Button(props: ButtonProps) {
  return <button>{props.name}</button>;
}

```

The result is:

```
export default function Button(props) {
  return <button>{props.name}</button>;
}
```
or:

```
import { Component, Input } from "@angular/core";

@Component({
  selector: "button",
  template: `
    <button>{{name}}</button>
  `,
})
export default class Button {
  @Input() name: any;
}
```

If I want to use mitosis to create my own components library, how I can get the types of my components to be able to expose them?

PatrickJS:
yeah thats a good point. @samijaber what do you think is the best way to transfer the types over?

samijaber:
> If I want to use mitosis to create my own components library, how I can get the types of my components to be able to expose them?

Hey @CKGrafico, that's a very good question. Unfortunately, at the moment, all types get stripped out when the component JSON is generated from the Mitosis source code. Preserving types is definitely an important feature that we want to incorporate into Mitosis though, so let me try to break down what this requires!

Our JSX parser already supports TypeScript syntax:

https://github.com/BuilderIO/mitosis/blob/29b38300268c88832fe0aebf878a87da7f744047/packages/core/src/parsers/jsx.ts#L790

However, we do not store any of these types yet. It will be a lot of work to store TS types, but we can definitely chip away at this problem incrementally.

I'm going to lay out what steps I think are needed to accomplish that. Sharing it in the open as notes for myself, but also in case you or anyone else is intrigued and would like to help in some capacity!

## Example 

If we were to use this very simple example:

https://astexplorer.net/#/gist/2ab2a4286ccc434c75e93310bfea78bd/d697bb79cffd991440d7bcb0e6cc87898541f73b

You can see: 
- `TypeAliasDeclaration` and `InterfaceDeclaration` which represent the type declarations `Props` and `Props2`
- `Parameter` objects have a `type: TypeReference` value which stores, e.g. `: Prop` within the `(prop: Prop)` argument declaration.


## 1. Store type information

We would need to start storing this information in the Mitosis JSON. We could have root-level `types` and `interfaces` keys in our `MitosisComponent` object that store such information as a start (this wouldn't account for types defined within the component body for now, but that's fine):
[...]

PR #420, opened 2022-05-30
feat: preserve react component props types

decadef20:
## Description

Add a short description of:
- what changes you made, 
   #372 
- why you made them, and 

##############

the PR is WIP, any suggestions are welcome. 

Now I am diving into `mitosis build`,  I am not much familiar with it.
### DONE
- [x] react framework support preserving props type and `export type/interface`

### TODO 
- [ ] context file type
- [ ] parse `'name' as string`
- [ ]  support other framework
- [ ]  building process

### DIFF
--- a/packages/core/src/parsers/jsx.ts
+++ b/packages/core/src/parsers/jsx.ts
@@ -163,23 +163,24 @@ const parseCodeJson = (node: babel.types.Node) => {
   return tryParseJson(code);
 };
 
-const getPropsTypeRef = (node: babel.types.FunctionDeclaration): string => {
-  const param = (node.params[0] || {}) as babel.types.Identifier;
-  let propsTypeRef = '';
+const getPropsTypeRef = (
+  node: babel.types.FunctionDeclaration,
+): string | undefined => {
+  const param = node.params[0];
   // TODO: component function params name must be props
   if (
+    babel.types.isIdentifier(param) &&
     param.name === 'props' &&
     babel.types.isTSTypeAnnotation(param.typeAnnotation)
   ) {
     const paramIdentifier = babel.types.variableDeclaration('let', [
       babel.types.variableDeclarator(param),
     ]);
-    propsTypeRef = generate(paramIdentifier)
+    return generate(paramIdentifier)
       .code.replace(/^let\sprops:\s+/, '')
       .replace(/;/g, '');
   }
-
-  return propsTypeRef;
+  return undefined;
 };
 
 const componentFunctionToJson = (
--- a/packages/core/src/parsers/jsx.ts
+++ b/packages/core/src/parsers/jsx.ts
@@ -163,6 +163,25 @@ const parseCodeJson = (node: babel.types.Node) => {
   return tryParseJson(code);
 };
 
+const getPropsTypeRef = (node: babel.types.FunctionDeclaration): string => {
+  const param = (node.params[0] || {}) as babel.types.Identifier;
+  let propsTypeRef = '';
+  // TODO: component function params name must be props
+  if (
+    param.name === 'props' &&
+    babel.types.isTSTypeAnnotation(param.typeAnnotation)
+  ) {
+    const paramIdentifier = babel.types.variableDeclaration('let', [
+      babel.types.variableDeclarator(param),
+    ]);
+    propsTypeRef = generate(paramIdentifier)
+      .code.replace(/^let\sprops:\s+/, '')
+      .replace(/;/g, '');
+  }
+
+  return propsTypeRef;
+};
+
 const componentFunctionToJson = (
   node: babel.types.FunctionDeclaration,
   context: Context,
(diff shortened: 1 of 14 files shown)


samijaber:
@decadef20 it does look like you need to update the Jest snapshots for the PR to pass ✅

decadef20:
> @decadef20 it does look like you need to update the Jest snapshots for the PR to pass ✅

Okay. I'll check test

decadef20:
> @decadef20 it does look like you need to update the Jest snapshots for the PR to pass ✅

All tests are passed @samijaber

PatrickJS:
@decadef20 can you pull latest and generate the snaps again

packages/core/src/types/mitosis-component.ts
```diff
@@ -62,4 +62,7 @@ export type MitosisComponent = {
   };
   children: MitosisNode[];
   subComponents: MitosisComponent[];
+  types?: string[];
+  interfaces?: string[];
```
decadef20, line 66: Maybe we don't need `types` and `interfaces` in two fields at the same time, just use one field to record both `type`  and `interface`  typing. What do you think? @samijaber @PatrickJS

samijaber: @decadef20 Do you mean something like:

```
typeReferences: {
  types?: string[];
  interfaces?: string[]
}
```

Or do you want to only have one array:

```
typesAndInterfaces?: string[]
```

We can't do the second option: I prefer if we keep types and interfaces separate so we can differentiate between them in the future if we need to (you never know!). But if you meant the first option, then I'm ok putting them under one object. Can also keep them as-is

decadef20: > @decadef20 Do you mean something like:
> 
> ```
> typeReferences: {
>   types?: string[];
>   interfaces?: string[]
> }
> ```
> 
> Or do you want to only have one array:
> 
> ```
> typesAndInterfaces?: string[]
> ```
> 
> We can't do the second option: I prefer if we keep types and interfaces separate so we can differentiate between them in the future if we need to (you never know!). But if you meant the first option, then I'm ok putting them under one object. Can also keep them as-is

I meant the second. But you are right, the first is easier to refactor

originswift-sys:
When merge? 🥺

PatrickJS:
@decadef20 are you going to add https://github.com/BuilderIO/mitosis/issues/421 to this PR or make another one?

decadef20:
> @decadef20 are you going to add https://github.com/BuilderIO/mitosis/issues/421 to this PR or make another one?

This PR has to take some time. I plan to make another PR for #421. Now I am on my vacation. I'll restart from tomorrow.

samijaber:
Reviewing this now @decadef20 ! 🙏🏽

samijaber (commented):
Looking very good!

I started reviewing before realizing you were not done. I left just a few high-level comments that you might find useful while you still work on this PR 👍🏽

samijaber:
@decadef20 What is missing from this PR by the way?

It does not need to include all the code for including types. We can merge them in small PRs: I actually prefer that more so that code is regularly being committed instead of creating a giant PR. The CLI code changes can come in a separate PR later on.

decadef20:
@samijaber thanks for your all CR comments which are really useful. I'll update later. 

My original intention is to implement all features listed in #420 in one PR. 

As you said, little PR is more proper. I can't agree with you more. I'll fix this PR later.

decadef20:
@samijaber @PatrickJS  I think it's ready, please help review

packages/core/src/parsers/jsx.ts
```diff
@@ -163,6 +163,25 @@ const parseCodeJson = (node: babel.types.Node) => {
...
+      babel.types.variableDeclarator(param),
+    ]);
+    propsTypeRef = generate(paramIdentifier)
+      .code.replace(/^let\sprops:\s+/, '')
+      .replace(/;/g, '');
+  }
+
+  return propsTypeRef;
+};
+
```
samijaber, line 184: Would be best to:
- avoid the type-cast, and check that `node.params[0]` is actually an Identifier
- return `undefined` when there is no type, instead of an empty string

```suggestion
const getPropsTypeRef = (
  node: babel.types.FunctionDeclaration,
): string | undefined => {
  const param = node.params[0];
  // TODO: component function params name must be props
  if (
    babel.types.isIdentifier(param) &&
    param.name === 'props' &&
    babel.types.isTSTypeAnnotation(param.typeAnnotation)
  ) {
    const paramIdentifier = babel.types.variableDeclaration('let', [
      babel.types.variableDeclarator(param),
    ]);
    return generate(paramIdentifier)
      .code.replace(/^let\sprops:\s+/, '')
      .replace(/;/g, '');
  }

  return undefined;
};
```

samijaber (commented):
Looks fantastic!! 🎉 

Just one comment on returning `undefined` values instead of empty strings when type refs aren't there

samijaber (approved):
Amazing, this is absolutely golden!

Really excited about the upcoming PRs to bring types to all generators

PR #436, opened 2022-06-03
Build basic and todo examples in CI

kylecordes:
Use yarn --immutable in CI per docs recommendations.

The previous improvement to include the examples in the workspace already ran them as part of PR build; this further runs them on any commit build.

Future PRs can work toward more complete CI, such as verifying the output code compiles under each supported target, verifying the resulting applications work via Playwright, etc.


samijaber (approved):
🎉 🎉 🎉

Issue #438, opened 2022-06-03
core: MitosisComponent.refs


PatrickJS:
add refs to MitosisComponent for type and argument

Issue #437, opened 2022-06-03
core: add compile away packages


PatrickJS:
remove custom mitosis package such as internal published versions of mitosis

Issue #400, opened 2022-05-24
core: allow for useRef to be used for values other than DOM refs


PatrickJS:
In react we need to use useRef to work around some features like `previousValue`. Currently we only use useRef for DOM references but we should separate the two. We can look for any useRef used in the jsx and determine anything not used will be used to maintain refs. 

In angular this is as simple as adding a prop to the class that isn't in the template.

In webcomponent it's the same as angular and right now we're using `this.` to create these types of refs. 

In react there's only two ways to use a ref outside of rendering loop. useRef or state that lives outside of the function component (can be an export).

Issue #186, opened 2022-02-27
Mitosis refs do not work when you compile to Vue


JLarky:
[...]

PR #439, opened 2022-06-03
fix(core/react/angular/webcomponent): useRef, compileAwayPackages, useRef with state

PatrickJS:
## Description
revert: https://github.com/BuilderIO/mitosis/pull/433
> it turned out `?.current` was invalid and the real issue was not having `useRef(null)` for react

fix: https://github.com/BuilderIO/mitosis/issues/438
fix: https://github.com/BuilderIO/mitosis/issues/437
fix: https://github.com/BuilderIO/mitosis/issues/400
fix: https://github.com/BuilderIO/mitosis/issues/186
fix: https://github.com/BuilderIO/mitosis/issues/440
fix: https://github.com/BuilderIO/mitosis/issues/442
fix: https://github.com/BuilderIO/mitosis/issues/455
fix: https://github.com/BuilderIO/mitosis/issues/457


packages/core/src/index.ts
```diff
@@ -4,7 +4,7 @@ export type Context<T> = {};
 
 // These compile away
 export const useState = <T>(obj: T) => obj;
-export const useRef = () => null as any;
+export const useRef = <T>(obj?: null | undefined | T) => obj as unknown as T;
```
PatrickJS, line 7: @samijaber I need some typing help here. I want useRef to be valid if no arguments or type params are added

`const something = useRef()`

packages/core/src/parsers/jsx.ts
```diff
@@ -101,7 +101,7 @@ export const createFunctionStringLiteral = (node: babel.types.Node) => {
   return types.stringLiteral(`${functionLiteralPrefix}${generate(node).code}`);
 };
 export const createFunctionStringLiteralObjectProperty = (
-  key: babel.types.Expression | babel.types.PrivateName,
+  key: babel.types.Expression | babel.types.PrivateName | any,
```
PatrickJS, line 104: for some reason this started to error without `any`. I'm not sure what typescript problem was added that can cause this

packages/core/src/generators/react.ts
```diff
@@ -521,7 +525,7 @@ const _componentToReact = (
   const refs = getRefs(json);
   let hasState = Boolean(Object.keys(json.state).length);
 
-  mapRefs(json, (refName) => `${refName}?.current`);
```
PatrickJS, line 524: it turned out `?.current` was invalid and the real issue was not having `useRef(null)`

packages/core/src/__tests__/data/basic-ref.raw.tsx
```diff
@@ -5,7 +5,8 @@ export interface Props {
 }
 
 export default function MyMasicRefComponent(props: Props) {
-  const inputRef = useRef();
+  const inputRef = useRef<HTMLInputElement>(null);
+  const inputNoArgRef = useRef(undefined);
```
PatrickJS, line 9: I'm not sure how to type this correctly

packages/core/src/index.ts
samijaber: Ugh, wrote a comment but it got erased somehow

Why do we want to make no-arguments a valid format? A reference needs to have some sort of default initial value, and it is far better for the end-user to provide that explicitly, rather than we do something implicit and then fiddle with types etc.

That would follow React's API as well: whatever type is provided to `useRef` is the type of the reference, and the user has to match that:

```suggestion
export const useRef = <T>(obj:T) => obj;
```

samijaber: It will also vastly simplify our parser and generators code, which in this PR involve a bunch of code to figure out if this is a dom ref or not, to then figure out that an initial `null` needs to be set. 

I think we should just do what React does and keep things simple.

samijaber (changes requested):
Requesting changes because I think that the implementation of `useRef` should not handle implicit initial default values, and instead require that information upfront explicitly from the user.

packages/core/src/index.ts
PatrickJS: React allows for undefined its just for DOM refs it should be set to null

samijaber: I missed the notification for this last comment.

Do you still want to allow no args if the type args are specified?

PatrickJS:
@samijaber can you review it now?

samijaber:
@PatrickJS Looking now

packages/core/src/helpers/get-props-ref.ts
```diff
@@ -0,0 +1,36 @@
...
+        const _prop = match?.[2];
+        if (_prop) {
+          prop = _prop;
+        }
+        if (shouldRemove) {
+          delete item.bindings.ref;
+        }
+        has = true;
+        this.stop();
+      }
```
samijaber, line 32: No need to loop over every binding here when we just need `ref`:

```suggestion
      const binding = item.bindings.ref;
      const regexp = /(.+)?props\.(.+)( |\)|;|\()?$/;
      if (binding && regexp.test(binding.code)) {
        const match = regexp.exec(binding.code);
        const _prop = match?.[2];
        if (_prop) {
          prop = _prop;
        }
        if (shouldRemove) {
          delete item.bindings.ref;
        }
        has = true;
        this.stop();
      }
```

samijaber (approved):

Issue #435, opened 2022-06-02
docs: update local developer docs to mention command yarn run fmt:prettier


chadalen:
- [x] I am interested in helping provide a fix!

The local developer docs found [here](https://github.com/BuilderIO/mitosis/tree/main/developer) should mention after the developer makes their changes to run `yarn run fmt:prettier` in the root of the project to make sure their code is formatted so the linter doesn't fail.

For reference see [here](https://github.com/BuilderIO/mitosis/pull/429#issuecomment-1145154278)

kylecordes:
In a recent change, I make that doc direct folks to run `yarn ci` before submitting; that include Prettier, tests, etc.

Maybe that chunk of text should explain what all checks `yarn ci` performs, i.e. Prettier and more?

samijaber:
@chadalen Are you still interested in doing this? Should be a quick doc change. Otherwise I can unassign you

chadalen:
@samijaber Here's the PR https://github.com/BuilderIO/mitosis/pull/528

samijaber:
Closed by #528

PR #441, opened 2022-06-03
feat(.github): always run prettier

PatrickJS:
## Description
always run prettier which will help for docs and first contribs

can help with https://github.com/BuilderIO/mitosis/issues/435

### DIFF
--- a/.github/workflows/pr-runner.yml
+++ b/.github/workflows/pr-runner.yml
@@ -27,6 +27,9 @@ jobs:
         run: yarn --immutable
 
       - name: Run prettier
+        run: yarn run fmt:prettier
+
+      - name: Run Lint
         run: yarn ci:lint
 
       - name: Build


.github/workflows/pr-runner.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         run: yarn --immutable
 
       - name: Run prettier
+        run: yarn run fmt:prettier
+
+      - name: Run Lint
         run: yarn ci:lint
```
PatrickJS, line 33: we can update this to `yarn run lint:eslint` and `yarn run lint:css` when the deps are fixed

.github/workflows/pr-runner.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         run: yarn --immutable
 
       - name: Run prettier
+        run: yarn run fmt:prettier
```
samijaber, line 30: This is going to format on-the-fly in the PR, which doesn't make sense. If a contributor forgets to prettify a document:

- this task will re-write it
- the next task `lint` will say that everything works fine


But then the un-prettified code gets committed, and if another dev saves the broken document on their machine, they will get a ton of new formatting applied, unrelated to their change.


The only way to "fix" this issue is to add a husky pre-commit hook to the repo, but that's a pain IMHO, I'd rather just have a line in the doc telling people to run `yarn run fmt:prettier` before they commit

samijaber (changes requested):
I don't think we can add `prettier --write` to the PR runner

Issue #430, opened 2022-06-02
End-to-end testing


samijaber:
## Problem

Our current snapshot testing in `packages/core` only does so much. It does not:

- always make sure that the generated code is even valid JS
- that the generated components are valid in regard to the rules of their frameworks
- that the generated components have the features they are meant to have (interactivity, state management, styling, lifecycle hooks logic, etc.)

Another big issue is that we do not have a good idea of what features are supported by which generators at any given point. We currently have about 12 generators, and they don't all support every single feature and edge case. It is unfeasible to manually gather that information and keep it up to date.

## Ideal process

The ideal process I have in mind to accomplish this type of code coverage is to have an end-to-end testing pipeline like so:

- write a Mitosis project containing many components in `packages/testing`
- run it through the Mitosis CLI to generate a directory for each framework we support
- render every generated version of every framework on some server (N servers for N frameworks)
- For each framework, run the same integration test suite on each of its generated components in Storybook using a tool like Playwright https://storybook.js.org/addons/storybook-addon-playwright

To solve the issue of identifying feature support, we would then add some logic that performs all of these tests against every output, but allows for certain failures. We then need to generate a matrix of feature support (e.g. each row is a feature, each column is a framework) that gets printed in a markdown file in the repo (where [this](https://github.com/BuilderIO/mitosis/blob/c701c0009893df07d56f5eabdc8910d81561b1d5/docs/feature-matrix-draft.md#L1) currently lives)

## Example
So, if we have a clickable button, say `packages/testing/src/button.lite.tsx`.

1 - We would run `mitosis build`, and end up with:
- `packages/testing/output/vue/button.vue`
- `packages/testing/output/svelte/button.svelte`
- `packages/testing/output/react/button.tsx`
- `packages/testing/output/solidjs/button.tsx`
- etc.

2- We would load all of those using parallel servers (each in its own Github Action)

3- We would then have a test, say `packages/testing/playwright/tests/button.ts`. This test would execute on each github action against a different server.

4- We would finally compile all of the results and generate a matrix based on what passed and what failed.

## Initial implementation

The quickest way to get an MVP of this idea is:
- Create a single, simple test (a div with text)
- Output it to just 1 framework (React to start would be good)
- import the generated components to a simple server that supports React 
- run a playwright test that checks that the div has the expected text
- do all of the above in a Github Action task

Once that baseline is there, next steps are to:
- add more frameworks support: Vue & Svelte
- update testing to handle multiple framework tests in parallel

[...]

PR #445, opened 2022-06-04
CI progress

### LINKED ISSUE #431: Add testing for CLI output
Baby step of https://github.com/BuilderIO/mitosis/issues/430, idea by @kylecordes 

- Update `examples/basic`'s dependency on CLI to be `workspace:*`
  - this entails adding that folder as a yarn workspace, and sorting any outstanding issues on that front
- add `yarn run build` in `examples/basic` as a CI test step

kylecordes:
Starting E2E implementation, progress on #430 and closes #431.

This implementation **assumes Mitosis is intended to emit files ready to ship/consume as an NPM package**, perhaps by a sibling project in a monorepo. That wasn't written down anywhere, but I saw hints this direction.

See the commit comments for more details.

I first started with a single workspace to test multiple targets, but found I was creating a multi-project mechanism within a single workspace... which is Yarn's job. So I switched to one workspace per target.

Free payoff of multiple workspaces: the yarn foreach mechanism parallelizes e2e execution!

This PR has several of eventually N workspaces, one for each target, or occasionally (as with React) JS and TS variations, where both kinds of consumer code are common. This also serves as an example for users on how to have a monorepo, define components in a Mitosis project, and consume them from a project based on React/Vue/Solid/etc.

I have in mind Playwright to verify that the output runs, shows a correct GUI, responds to clicks, etc. Mitosis working **e2e** would mean a pile of components are created once, compiled to N targets, and all N targets pass the same web-automation test suite. (Plus later other consideration for non-web target ideas.)

To try the progress so far: `yarn ci`

I have no idea how close this is to what the Mitosis team has in mind for E2E, feel free to savage it, I will not be mortally offended.


packages/e2e-vue3-ts/vite.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite';
+import vue from '@vitejs/plugin-vue';
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [vue()],
+});
```
samijaber, line 7: PS: Our vue 3 output is actually broken in Mitosis, but that's something that's going to be fixed in the _very_ near future. So we might want a vue 2 server here instead for now, or just drop the vue 3 package.

samijaber: oh wow...or it's not broken! 😄 This works for me locally 🎉

kylecordes: The only output is under “nuxt” and appears to work with Vue 3.

It barely works - can't even do two components because the inter-component import doesn’t work.

But at least this will lock in this first increment of working. Enable the ratchet of progress.

kylecordes:
I believe this is ready to go, except for one problem:

**We treat the Vue output as Vue3**, and it works for this simple case. I've attempted to switch to Vue 2, but not been able to get all the wiring to work. Perhaps we can proceed with this as it is, and seek a Vue 2 expert to take a look?

Avenues of further work:

* Cover more targets; but some of them are waiting for mitosis fixes.
* Split the trivial e2e-app into 2 components - this need a Vue fix first.
* Store the E2E output reports from Playwright, in case a developer wants to look at them without running locally.
* Split apart the results in GitHub actions somehow, so that failure details are more easily spotted without studying lengthy build output.
* Consider a more powerful build tool, either with or instead of yarn workspaces; currently every build rebuilds everything, which is unnecessary.
* Consider CI caching or a CI container which has the needed browser pre-installed, to speed up CI.

packages/e2e-solid/playwright.config.ts
```diff
@@ -0,0 +1,6 @@
+import { configFor } from '@builder.io/e2e-app-spec/shared.config';
+
+export default configFor(
+  'yarn workspace @builder.io/e2e-solid run serve --port 7501',
+  7501,
```
samijaber, line 5: I wonder if we can change `configFor`'s logic so that we don't have to include the `--port $PORT` in the command, and instead append it in `configFor`. That would remove the duplication we're seeing each time.

packages/e2e-solid/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@builder.io/e2e-solid",
+  "private": true,
+  "version": "0.0.1",
+  "scripts": {
+    "build": "vite build",
+    "dev": "vite",
+    "e2e": "playwright test",
+    "playwright-install": "playwright install --with-deps",
```
samijaber, line 9: Is there a particular reason why this command has to exist within one of the `e2e` packages, and not in the root `package.json`?

packages/e2e-solid/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "serve": "vite preview"
+  },
+  "dependencies": {
+    "@builder.io/e2e-app": "workspace:*",
+    "@builder.io/e2e-app-spec": "workspace:*",
+    "solid-js": "^1.4.2",
+    "solid-styled-components": "^0.28.1"
+  },
+  "devDependencies": {
+    "typescript": "^4",
```
samijaber, line 19: One thing: could we add `playwright` as a devDependency to each `e2e` project as well? If we don't, it's impossible to run `yarn run e2e` within just one of the e2e packages, which I think will be needed when debugging individual outputs 🤔. What do you think? 

Can make it a very wide version range like you did with TS here, so that they all use the same version.

samijaber (commented):
I am _so_ excited for this PR. Absolutely fantastic effort, thank you so much @kylecordes.

I have left a few small comments, but once they're addressed this thing is getting shipped ASAP.

packages/e2e-app/src/components/my-component.lite.tsx
```diff
@@ -0,0 +1,60 @@
+import '@builder.io/mitosis/dist/src/jsx-types';
```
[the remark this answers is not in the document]
samijaber: (This is more for my own out-loud thoughts, nothing actionable in this PR!)

package.json
```diff
@@ -13,18 +13,21 @@
...
     "fmt": "yarn fmt:prettier",
     "fmt:prettier": "yarn lint:prettier --write",
     "lint": "run-p -c lint:*",
     "lint:css": "stylelint 'packages/**/*.ts{,x}'",
     "lint:eslint": "eslint 'packages/**/*.ts{,x}'",
-    "lint:prettier": "yarn prettier --check",
+    "lint:prettier": "yarn run prettier --check",
     "prerelease": "yarn build",
-    "prettier": "prettier '{packages,docs}/**/*.{js,jsx,ts,tsx,json,md,html}'",
+    "prettier": "prettier --loglevel warn '{packages,docs}/**/*.{js,jsx,ts,tsx,json,md,html}'",
```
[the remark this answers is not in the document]
kylecordes: The CI output is noisy - this is a tiny step in a less noisy direction. Noone seems to know why prettier emits so much output when it takes no action this setting makes it not do that, many prettier users think this should be the default. I apologize for the scope creep toward "make the CI better generally"...

packages/e2e-solid/playwright.config.ts
kylecordes: Yes, there are def ways to make this better. Important to do now? Or followup?

packages/e2e-solid/package.json
kylecordes: I put it here because this package already had the dependency. Because of a VERY long story around how Typescript looks up types, this project uses Yarn in a mode that doesn't consolidate/hoist, so this is slightly faster. Running it from one of the points where we execute Playwright seemed expedient, though a bit ugly.

package.json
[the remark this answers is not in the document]
samijaber: No worries! Was just curious about the impact. Fine by me, thanks for cleaning it up 👍🏽

packages/e2e-app/src/components/my-component.lite.tsx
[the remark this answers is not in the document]
samijaber: Yeah for sure, let's not creep scope this!

packages/e2e-solid/playwright.config.ts
samijaber: Followup is totally fine!

packages/e2e-solid/package.json
samijaber: Love a long story around Typescript & types 😄.

Not a big deal at all, let's do this in a follow up if at all

packages/e2e-solid/package.json
samijaber: This is also ok in a followup PR @kylecordes. Let's merge this thing since it's all good and working!!

samijaber (approved):

packages/e2e-solid/package.json
samijaber: Oh, just saw your answer right before mine. Adding some docs would be good to clarify that, as people might try to do what I did (cd into a package, then run playwright there and see it fail).

PR #448, opened 2022-06-05
fix(qwik): correct regression with bindings inside of loops

mhevery:


packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -251,7 +251,8 @@ export class SrcBuilder {
         Object.prototype.hasOwnProperty.call(bindings, rawKey) &&
         !ignoreKey(rawKey)
       ) {
-        let { code: binding } = bindings[rawKey];
+        let binding = bindings[rawKey];
+        binding = (binding && binding.code) || binding;
```
samijaber, line 255: Old habits die hard, but you can use the optional chaining operator today for this!
```suggestion
        binding = binding?.code || binding;
```

samijaber (approved):

packages/core/src/generators/qwik/src-generator.ts
PatrickJS: We also got `||=`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment

mhevery: fixed

Issue #421, opened 2022-05-31
core: include local file "export"


PatrickJS:
We should include `export`s within the same file.

https://mitosis.builder.io/?outputTab=react&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+const+defaultValues+%3D+%7B%0A++name%3A+%27hello%27%0A%7D%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+state+%3D+useState%28%7B%0A++++name%3A+%27%27%0A++%7D%29%3B%0A%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%3Cinput%0A++++++++css%3D%7B%7B%0A++++++++++color%3A+%22red%22%2C%0A++++++++%7D%7D%0A++++++++value%3D%7BdefaultValues.name+%7C%7C+name%7D%0A++++++++onChange%3D%7B%28event%29+%3D%3E+state.name+%3D+event.target.value%7D%0A++++++%2F%3E%0A++++++Hello%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D

in Angular we can detect if it's being used then add it to the class. This should help with default values

PatrickJS:
@decadef20 do you think you can do this one? We want to collect all exposed `export` variables probably in MitosisComponent.exportRef what do you think @samijaber?

steve8708:
ah there isn't actually, the original intent was to just straight up copy everything above and below the component into the output, so if you wrote

```
whatever()

export default function MyComponent() { ... }

whatever()
```

we would just copy the added code above/below in the output generator

I think I removed this at some point to not cause confusion, as it's important not to do this as in the past it wouldn't work as expected. now tho it actually might be perfectly ok to support (we'd just have to handle those local vars being forwarded to being available in the component, e.g. we have some logic for this right now with imports)


```
const foo = someSetupCode();

export default function MyComponent() { 
 }
```


tl;dr yeah I think we actually can support, now, arbitrary code above and below components and just copy them over to the same spot in the generators

so the above example can easily just be

```
<template></template>
<script>
econst const foo = someSetupCode()
export default {
  name: "my-component",
};
</script>
```

and if that export is used in the template, it'll need to be forwarded in some cases, like

```
<template></template>
<script>
econst const foo = someSetupCode()
export default {
[...]

PR #449, opened 2022-06-06
Feature/include local export file

decadef20:
## Description

Add a short description of:
- what changes you made, 
   #421 
- why you made them, and 

Please help review react and vue framework test snap. If passed,  I'll fully support other frameworks.


decadef20:
@samijaber  help review this PR

packages/core/src/__tests__/vue.test.ts
```diff
@@ -31,6 +31,7 @@ describe('Vue', () => {
     const output = componentToVue()({ component, path });
     expect(output).toMatchSnapshot();
   });
+  return;
```
samijaber, line 34: This line delete the snapshots of all the tests below.
```suggestion
```

decadef20: Sorry. This is debug code. Forget to delete.

decadef20: Fixed.

PatrickJS (approved):

packages/core/src/parsers/jsx.ts
```diff
@@ -853,6 +878,42 @@ export function parseJsx(
...
+                  isFunction = babel.types.isFunction(
+                    (node.declaration as any).declarations[0].init,
+                  );
+                }
+                if (babel.types.isVariableDeclaration(node)) {
+                  name = (node.declarations as any)[0].id.name;
+                  isFunction = babel.types.isFunction(
+                    (node.declarations as any)[0].init,
+                  );
+                }
```
samijaber, line 903: This allows a bunch of nodes that will cause crashes. We should add checks for `isIdentifier` to the inner node


```suggestion
                if (
                  babel.types.isExportNamedDeclaration(node) &&
                  babel.types.isVariableDeclaration(node.declaration) &&
                  babel.types.isIdentifier(node.declaration.declarations[0].id)
                ) {
                  name = node.declaration.declarations[0].id.name;
                  isFunction = babel.types.isFunction(
                    node.declaration.declarations[0].init,
                  );
                } else if (
                  babel.types.isVariableDeclaration(node) &&
                  babel.types.isIdentifier(node.declarations[0].id)
                ) {
                  name = node.declarations[0].id.name;
                  isFunction = babel.types.isFunction(
                    node.declarations[0].init,
                  );
                }
```

samijaber (commented):
The most important issue to fix is the one about the `jsx` parser: removing the `any` casts and adding more node type checks to avoid crashes

packages/core/src/generators/vue.ts
```diff
@@ -490,6 +509,13 @@ export const componentToVue =
       );
     }
 
+    if (localVarAsFunc.length) {
+      functionsString = functionsString.replace(
+        /}\s*$/,
+        `${localVarAsFunc.join(',')}}`,
+      );
+    }
```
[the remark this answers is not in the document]
decadef20: I have tried to modify `component.state` directly but failed to put it in right place.

The plugin looks great, I'll try it later.

packages/core/src/parsers/jsx.ts
decadef20: I'll fix it later

packages/core/src/generators/vue.ts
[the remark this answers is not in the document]
decadef20: I tried to use plugin, 
```bash
-   data: () => ({ name: "Steve", DEFAULT_VALUES }),
+   data: () => ({ name: "Steve", DEFAULT_VALUES: "DEFAULT_VALUES" }),
```
Obviously, it's not what we want .
Or we could copy value to data, like this 
```bash
-   data: () => ({ name: "Steve", DEFAULT_VALUES }),
+   data: () => ({ name: "Steve", DEFAULT_VALUES: { name: "Steve",} }),
```

PR #453, opened 2022-06-06
Feature/local exports

samijaber:
## Description

Re-instate #449 , including a bug-fix in the JSX parser


packages/core/src/parsers/jsx.ts
```diff
@@ -889,16 +889,22 @@ export function parseJsx(
...
-                if (babel.types.isVariableDeclaration(node)) {
-                  name = (node.declarations as any)[0].id.name;
+                } else if (
+                  babel.types.isVariableDeclaration(node) &&
+                  babel.types.isIdentifier(node.declarations[0].id)
+                ) {
+                  name = node.declarations[0].id.name;
                   isFunction = babel.types.isFunction(
-                    (node.declarations as any)[0].init,
+                    node.declarations[0].init,
```
samijaber, line 907: cc @decadef20 , wanted to point this out to you. It's always much safer to figure out what the correct type is, and narrow down to the type, instead of casting `any`. This can avoid crashes if an unexpected value is consumed by this logic.

samijaber:
cc @PatrickJS this is merged back into `main` now.

packages/core/src/parsers/jsx.ts
decadef20: Thanks. Remember it! It's a really good practice

PR #460, opened 2022-06-08
docs(hooks): forwardRef for React

PatrickJS:
## Description
how to use forward for React

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -11,11 +11,11 @@
 
 Use the `useRef` hook to hold a reference to a rendered DOM element.
 
-```jsx
+```typescript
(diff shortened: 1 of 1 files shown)


docs/hooks.md
```diff
@@ -25,7 +26,7 @@ export default function MyComponent() {
     },
   });
 
-  const inputRef = useRef();
+  const inputRef = useRef < HTMLInputElement > null;
```
samijaber, line 29:
```suggestion
  const inputRef = useRef<HTMLInputElement>(null);
```

PatrickJS: prettier actually did this haha so I just removed it

Issue #462, opened 2022-06-08
react: transfer types forwardRef


PatrickJS:
We need to parse the type in the Props and transfer the type to the forwardRef

mitosis input
```typescript
export interface Props {
  inputRef: MyType;
}

export default function MyComponent(props: Props) {
  return <input ref={props.inputRef} />
}
```

react output
```typescript
export interface Props {
  inputRef: MyType;
}

export default forwardRef<MyType>(function MyComponent(props: Props, inputRef) {
  return <input ref={inputRef} />
})
```

PatrickJS:
@samijaber we need a parser to parse the types for local exports. We can't transfer types for Angular output either without that

samijaber:
@PatrickJS The path of least resistance to get this working without any additional parsing is to simply access `Props["inputRef"]`. We know the name of the props type from `json.propsTypesRef`, and the name of the prop through the `getRefsString`.

Will put up a WIP PR for you to see what I mean

Issue #424, opened 2022-05-31
webcomponent: correctly export webcomponent file


PatrickJS:
we may want to export everything within the webcomponent file to be used include the webcomponent itself

PatrickJS:
fixed by https://github.com/BuilderIO/mitosis/pull/449 and https://github.com/BuilderIO/mitosis/pull/464

PR #464, opened 2022-06-08
Feat type props fix forward ref type

PatrickJS:
fix: https://github.com/BuilderIO/mitosis/issues/462
fix: https://github.com/BuilderIO/mitosis/issues/424


samijaber (approved):

packages/core/src/__tests__/__snapshots__/angular.test.ts.snap
```diff
@@ -358,6 +387,11 @@ export default class MyBasicOutputsComponent {
 exports[`Angular className to class 1`] = `
 "import { Component } from \\"@angular/core\\";
 
+type Props = {
+  [key: string]: string | JSX.Element;
+  slotTesting: JSX.Element;
```
PatrickJS, line 392: we might need to remove some props from types in the future

Issue #389, opened 2022-05-23
core: allow destructure assignment of state and props


PatrickJS:
we should undo the destructure assignment so the rest of the code generators still work. In some parts of the generators we are relying on props. or state.

samijaber:
> we should undo the destructure assignment

What do you mean by that exactly? Which de-structure assignment?

PatrickJS:
react developers do de-structure assignments all the time for props. A lot of the generator code relies on `props.` so it would be better to undo their de-structured props

```jsx
function MyComponent({children}) {

  return (
    <div>
      {children}
    <div>
  )
}
```
convert it to props before the generator runs

```jsx
function MyComponent(props) {

  return (
    <div>
      {props.children}
    <div>
  )
}
```

samijaber:
Oh I see. You're saying that the JSX Parser should include logic that, if `props` or `state` is de-structured, then it should be smart enough to swap the restructure with `props`/`state`, and the assignments/uses of the variables with `props.children` and `state.foo` so that the Mitosis syntax is still able to rely on de-structuring for better dev UX.

Yeah, that would be neat and not a particularly difficult thing to implement either!

PatrickJS:
💯 Yeah that would be great

decadef20:
the de-structured assignment may do alias or set a default value. If we undo it,  we need to handle the default value of props

```js
export default function MyComponent({children: c, type, color = 'blue'}) {

  return (
    <div>
      {c}
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
}
```

samijaber:
Closing as I believe #468 implemented this!

PR #468, opened 2022-06-10
Feature/props destructure

decadef20:
## Description

Add a short description of:
- what changes you made, 
     #389 
- why you made them, and 

### DIFF
--- a/packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
+++ b/packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
@@ -1,0 +1,15 @@
+import { useState } from '@builder.io/mitosis';
+
+// @ts-ignore
+export default function MyBasicComponent({children: c, type}) {
+  const state = useState({
+    name: 'Steve',
+  });
+
+  return (
+    <div>
+      {c} {type}
+      Hello! I can run in React, Vue, Solid, or Liquid!
+    </div>
+  );
+}
--- a/packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
+++ b/packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
@@ -1,9 +1,9 @@
 import { useState } from '@builder.io/mitosis';
 
 // @ts-ignore
-export default function MyBasicComponent({children: c, type}) {
+export default function MyBasicComponent({ children: c, type }) {
   const state = useState({
-    name: 'Steve',
+    name: 'Decadef20',
   });
 
   return (
--- a/packages/core/src/__tests__/data/basic-props.raw.tsx
+++ b/packages/core/src/__tests__/data/basic-props.raw.tsx
@@ -1,14 +1,14 @@
 import { useState } from '@builder.io/mitosis';
 
 // @ts-ignore
-export default function MyBasicComponent(props) {
+export default function MyBasicComponent({ children: c, type }) {
   const state = useState({
     name: 'Decadef20',
   });
 
   return (
     <div>
-      {props.children} {props.type}
+      {c} {type}
       Hello! I can run in React, Vue, Solid, or Liquid!
     </div>
   );
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -11,6 +11,8 @@
   - [Can't assign "params" to "state"](#cant-assign-"params"-to-"state")
   - [Can't assign function output to "state"](#cant-assign-function-output-to-"state")
   - [Can't destructure assignment from state](#cant-destructure-assignment-from-state)
+  - [Can't set default props value with destructuring](#can't-set-default-props-value-with-destructuring)
+  - [Can't destructure props as ...rest](#can't-destructure-props-as-...rest)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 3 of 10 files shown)


packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
```diff
@@ -0,0 +1,15 @@
+import { useState } from '@builder.io/mitosis';
+
+// @ts-ignore
+export default function MyBasicComponent({ children: c, type }) {
```
decadef20, line 4:
```js
function MyBasicComponent({ children: c, type, color = 'blue', ...rest }) {}
```


- props default value
we could recommend  defining a local variable to support props default value

- rest parameter
  I find it tricky to support rest props parameter in all framework.
 If there is something  I missed, please let me know.

PatrickJS: we can do default value and rest parameter in another PR 👍 can you create a Gotcha to say we can't do these

PatrickJS:
@decadef20 this is amazing 👍 do you think you can make another PR for `state`?

packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
decadef20: > we can do default value and rest parameter in another PR 👍 can you create a Gotcha to say we can't do these

Ok

PatrickJS (approved):

packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
PatrickJS: actually I think if we rename the rest parameter `rest` in the code to `props` then it should work for all frameworks

decadef20: You mean
 - input
  ```js
export default function MyBasicComponent({ children: c, type, ...rest }) {
  const state = useState({
    name: 'Decadef20',
  });

  return (
    <div {...rest}>
      {c} {type}
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
}
```

- output

```js
export default function MyBasicComponent(props) {
  const state = useState({
    name: 'Decadef20',
  });

  return (
    <div {...props}>
      {props.children} {props.type}
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
}

```

I also think this way. But there may be types error

decadef20:
> @decadef20 this is amazing 👍 do you think you can make another PR for `state`?

@PatrickJS  Ok. I'll try it later.

samijaber:
Reviewing this now @PatrickJS

packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
samijaber: @decadef20 your example output here would be incorrect, because `...props` will also contain '`props.children` and `props.type`, which the user does not want to provide to the `div`

decadef20: @samijaber  Yeah. agree with you. So I think it a little tricky

PatrickJS: oh yeah this does get really tricky

packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
```diff
@@ -0,0 +1,15 @@
...
+    name: 'Decadef20',
+  });
+
+  return (
+    <div>
+      {c} {type}
+      Hello! I can run in React, Vue, Solid, or Liquid!
+    </div>
+  );
+}
```
samijaber, line 15: Can we add a test to the JSX parser that compares the JSON output of parsing this component and one without de-structuring?

```tsx
import { useState } from '@builder.io/mitosis';

// @ts-ignore
export default function MyBasicComponent(props) {
  const state = useState({
    name: 'Decadef20',
  });

  return (
    <div>
      {props.children} {props.type}
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
```

You can add the test to `packages/core/src/__tests__/parse-jsx.test.ts`. Something like:

```js
expect(parseJsx(basicPropsRaw)).toEqual(parseJsx(basicxPropsDestructureRaw));
```

samijaber (commented):
Would just love a test that compares the 2 JSONs, to make sure the JSON itself is equal to that of a component without de-structuring. That will be a more powerful assertion.

packages/core/src/__tests__/data/basic-props-destructure.raw.tsx
decadef20: > Can we add a test to the JSX parser that compares the JSON output of parsing this component and one without de-structuring?
> 
> 
> 
> ```tsx
> 
> import { useState } from '@builder.io/mitosis';
> 
> 
> 
> // @ts-ignore
> 
> export default function MyBasicComponent(props) {
> 
>   const state = useState({
> 
>     name: 'Decadef20',
> 
>   });
> 
> 
> 
>   return (
> 
>     <div>
> 
>       {props.children} {props.type}
> 
>       Hello! I can run in React, Vue, Solid, or Liquid!
> 
>     </div>
> 
>   );
> 
> ```
> 
> 
> 
> You can add the test to `packages/core/src/__tests__/parse-jsx.test.ts`. Something like:
> 
> 
> 
> ```js
> 
> expect(parseJsx(basicPropsRaw)).toEqual(parseJsx(basicxPropsDestructureRaw));
> 
> ```

That's so thoughtful. I'll add it later

samijaber: @decadef20 if we can add to this PR before it's merged, that'd be great so we don't forget/lose track of it

decadef20: @samijaber yeah. I understand. It's just almost 2 A.M. I will add it in current PR when I wake up.

samijaber: Oh sounds good! No rush at all, this PR is not urgent 😄. I keep forgetting about the timezone differences too. 😴

decadef20: @samijaber  parse jsx test is added

packages/core/src/__tests__/data/basic-props.raw.tsx
```diff
@@ -0,0 +1,15 @@
+import { useState } from '@builder.io/mitosis';
```
samijaber, line 1: btw: you swapped the names of the files: this one should be `basic-props-destructure.raw.tsx`, not the other one.

PR #470, opened 2022-06-10
E2E Improvements

kylecordes:
NO need to merge yet

Revising, improving, expanding the E2E and related CI.


PatrickJS:
it would be great to have angular and webcomponent included

kylecordes:
This adds another target (Svelte), and the next few are awaiting fixes to Mitosis.

Therefore I'm marking this ready to review.

kylecordes:
@PatrickJS I want those too (I want all targets!), see above comment about maybe E2E to consuming Mitosis output as source (rather than via node_modules)  to get there faster.  Also see #473 and #476.

samijaber (approved):

samijaber:
Amazing @kylecordes ! 🏆

PR #480, opened 2022-06-13
better support both forms of state and functions in Mitosis

steve8708:
currently, we have a "useState" hook that supports two different formats, based on how it is used. the logic is funky - unlike most code, the behavior actually changes based on how it's written, which is weird and has needed to change

this changes the default name for the object format to useStore({}) and keeps useState() to be used just for the react format. this also fixes types accordingly and allows you to intermix both types of hooks and functions as you please

so now you can do:

```ts
export default function MyCompnent(props) {
  const state = useStore({
    myProperty: 'foo',
    myMethod() { /* ... */ },
    get myGetter() { /* ... */ },
 })
 function myFunction() { /* ... */ }
 const [foo, setFoo] = useState('foobar')

 return <>/*...*/</>
}
```

pros:
 - one name does one thing
 - avoid confusing of react's useState semantics and (part of) ours
 - this is backwards compatible, you can still use useState({...}) for now so you don't have to refactor old code
 - fixes/improves TS types

 cons:
  - `const state = useStore(...)` may cause confusion. state object vs state hooks

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,7 +20,7 @@ Mitosis uses a static subset of JSX, inspired by [Solid](https://www.solidjs.com
 
 ```tsx
 export function MyComponent() {
-  const state = useState({
+  const state = useStore({
     name: 'Steve',
   });
 
--- a/docs/components.md
+++ b/docs/components.md
@@ -18,10 +18,10 @@ Mitosis is inspired by many modern frameworks. You'll see components look like R
 An example Mitosis component showing several features:
 
 ```javascript
-import { For, Show, useState } from '@builder.io/mitosis';
+import { For, Show, useStore } from '@builder.io/mitosis';
 
 export default function MyComponent(props) {
-  const state = useState({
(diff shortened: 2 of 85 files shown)


PatrickJS:
yeah this is needed

steve8708:
got everything passing - ready for review @samijaber

samijaber (approved):

steve8708:
oh I do have those, I have snapshot tests with all forms, see [here](https://github.com/BuilderIO/mitosis/pull/480/files#diff-74091d2e238a82add688c4ad8deca814ca08bd8a57317477efb736645da7f9d5L32-L51) - more never hurts tho

ohkimur:
@steve8708 @samijaber Is this added to the `docs`, if not let me know and I'll make a PR.

steve8708:
pretty sure it either isn't there at all or at minimum could use some touching up, so would love a PR improving this @ohkimur!

PR #484, opened 2022-06-14
Refactor: Snapshot tests

samijaber:
## Description

Improve the infrastructure of snapshot tests. Moving forward, every snapshot test will by default be added to every generator. If the test should be ignored for certain frameworks, we can easily omit it from them.

- create a `__tests__/shared.ts` file that holds:
- - dictionaries of fixtures, and creates the jest tests on the fly for each generator.
- - a mapping of generators to fixtures, to figure out which fixtures should be used for each output.


packages/core/src/__tests__/shared.ts
```diff
@@ -0,0 +1,220 @@
...
+      Object.keys(tests).forEach((key) => {
+        test(key, () => {
+          const component = parseJsx(tests[key]);
+          const output = generator({ component, path });
+          expect(output).toMatchSnapshot();
+        });
+      });
+    });
+  }
+};
```
samijaber, line 220: ☝🏽  This is the file that matters

PatrickJS:
Sounds good

Dkendal (approved):

Issue #485, opened 2022-06-14
`propsTypesRef` not working with destructured props


samijaber:
If we add the same Props types to these two snapshots:

https://github.com/BuilderIO/mitosis/blob/da3e2671f5d76bf3a002ebee4f397493d3ea6996/packages/core/src/__tests__/data/basic-props-destructure.raw.tsx#L4
https://github.com/BuilderIO/mitosis/blob/da3e2671f5d76bf3a002ebee4f397493d3ea6996/packages/core/src/__tests__/data/basic-props.raw.tsx#L4

```ts
type Props = {
  children: any;
  type: string
}
```

Then the `parser-jsx` test will fail due to the de-structured props not receiving their types reference. CC @decadef20

decadef20:
I'll fix it later @samijaber

PR #486, opened 2022-06-15
fix: propsTypesRef not working with destructured props

decadef20:
## Description

Add a short description of:
- what changes you made, 
#485 
- why you made them, and 

### DIFF
--- a/packages/core/src/parsers/jsx.ts
+++ b/packages/core/src/parsers/jsx.ts
@@ -922,7 +922,7 @@ function undoPropsDestructure(
         {
           ...babel.types.identifier('props'),
           typeAnnotation: param.typeAnnotation,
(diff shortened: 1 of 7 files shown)


packages/core/src/parsers/jsx.ts
```diff
@@ -908,6 +917,16 @@ function undoPropsDestructure(
       return pre;
     }, {} as Record<string, string>);
 
+    if (param.typeAnnotation) {
+      node.params = [
+        {
+          ...babel.types.identifier('props'),
+          typeAnnotation: param.typeAnnotation,
+        } as babel.types.Identifier,
```
samijaber, line 925: This type-cast is unnecessary and can be removed

```suggestion
        },
```

samijaber (approved):

Issue #483, opened 2022-06-14
Component exports not including Functions


samijaber:
https://github.com/BuilderIO/mitosis/blob/7b5d4c9396c975b3af5ab83734622347cd546862/packages/core/src/parsers/jsx.ts#L1009

We are hitting this warning in our snapshot tests for React when running for this snapshot:
https://github.com/BuilderIO/mitosis/blob/7b5d4c9396c975b3af5ab83734622347cd546862/packages/core/src/__tests__/data/basic-ref-usePrevious.raw.tsx#L7

Looking at the logic:
https://github.com/BuilderIO/mitosis/blob/7b5d4c9396c975b3af5ab83734622347cd546862/packages/core/src/parsers/jsx.ts#L982-L1010

This is not accounting for `node.declaration` being a `FunctionDeclaration`

samijaber:
We should change the warning to throw an error, or at least improve the warning: "Could not parse export: ignoring..."

PR #487, opened 2022-06-15
fix: Component exports not including Functions

decadef20:
## Description

Add a short description of:
- what changes you made, 
#483 
- why you made them, and 


samijaber:
Taking a look now!

packages/core/src/__tests__/__snapshots__/webcomponent.test.ts.snap
```diff
@@ -1721,6 +1721,19 @@ exports[`webcomponent BasicRefPrevious 1`] = `
   showInput: boolean;
 }
 
+export function usePrevious<T>(value: T) {
+  // The ref object is a generic container whose current property is mutable ...
+  // ... and can hold any value, similar to an instance property on a class
+  let ref = useRef<T>(null); // Store current value in ref
```
samijaber, line 1727: Hmm, `useRef` is not being replaced in this export, so this will only work in React where Mitosis' `useRef` has the same name of React's `useRef`. 

Probably need to run this code through a processor to handle these imports 🤔  @PatrickJS

PatrickJS: yeah we only process the component function. we can fix this later but I added this for when we get to fix it in the future. if the function is left here this would almost work if we add `.current`

PR #497, opened 2022-06-17
CLI: Add Vue2 & Vue3 targets

samijaber:
## Description

Depends on #495 and #496

- big refactoring to CLI logic in `build`
- export `componentToVue2` and `componentToVue3` from core
- add `vue2` and `vue3` as possible targets in CLI


kylecordes:
re: allows for outputting vue 2 and/or vue 3

For maximum consistency, `vue2` and `vue3` would be separate targets. That way someone using both downstream gets them both from one run of mitosis - just as they get react and solid from one run of mitosis. But... this is of ~0 urgency :-)

samijaber:
@kylecordes yeah, this PR accounts for that already: https://github.com/BuilderIO/mitosis/pull/497/files#diff-c7b884310f6800dc7df19c24d0d186a55a872b36839b11c45e605a030b6df65cR208-R227

I went back-and-forth on having two separate targets `vue2` and `vue3`, but I do think I'll end up with that as it cleans things up a bit.

PR #500, opened 2022-06-18
E2E: Add Vue 2, add Angular, Mitosis output is source not package

kylecordes:
Revamp the E2E test to assume that Mitosis emits "source", not packaged output. This simplifies most of the harnesses, and much more closely matches the current capability of the project.

Add Vue 2 E2E.

Rename file outputs to enable further extension of the E2E.


kylecordes:
This seems like a big enough chunk to be worthy of merge.

Next step will be a way to have some tests fail and show the status somehow, rather than fail the build.

samijaber:
@kylecordes with #497 merged, you should be able to point to `vue2` and `vue3` by just using the new targets I added to the CLI

kylecordes:
@samijaber Excellent! I will do that and also catch up with the merge conflicts.

kylecordes:
Vue 2 and 3 are now tested, both pass.

They also both accommodate one component using another; but that doesn't work in Angular yet, so it can't be in this simple single-example version of the E2E.

In the next PR I will work on accommodating partial failures to populate a "what works for what target" test status grid.

packages/e2e-app/mitosis.config.js
```diff
@@ -1,18 +1,9 @@
...
-    'vue',
-    'html',
-    'svelte',
-    'solid',
-    'angular',
-    'webcomponent',
+    // 'qwik',  // CLI does not support target: qwik
+    // 'builder', // CLI does not support target: builder
+    'html', // HTML output in a JS file
+    'webcomponent', // TS output in a JS file
```
samijaber, line 7: If I understand correctly, we are now moving towards `e2e-app` simply storing our mitosis source code and not emitting anything? Can you elaborate a bit on why you preferred to sync the directory into each individual e2e app and run `mitosis build` there instead of building once in this location? Is the primary limitation not being able to provide a relative path for the destination? It sounds like your plan is to move back to the original system once the CLI becomes more powerful, so I want to have a clearer idea of what's blocking that. 


Also, in this case, can we delete this `mitosis.config.js` file and the `mitosis build` command from `e2e-app/package.json`? Or are they still being used under some circumstances?

samijaber: Thank you for the thoughtful write-up! I want to add as much context as possible based on our experience building Builder SDKs with Mitosis.

> First, it meant that partial failure caused total failure. If mitosis had an error for any target, no other targets could be tested; we want to get away from that and toward partial success being reported in a test results grid.

Makes perfect sense 👍🏽 

> Instead, with these libraries the expectation is that a package will run its own compilation/bundling/etc process and publish the output as a npm package

I am not certain of how true that might be. For Vue and Svelte code, you don't need to run any additional compilation steps: you can provide an npm package that contains `.svelte` and `.vue` files and they will be consumed just fine.

In Svelte's case, it's actually problematic to pre-compile them before publishing, due to Svelte version mismatches. So you _need_ to provide the original `.svelte` file: https://github.com/sveltejs/svelte/issues/6584#issuecomment-1019578529

[Our Mitosis-generated Builder SDKs](https://github.com/BuilderIO/builder/tree/main/packages/sdks/) are currently outputted in Vue, React-Native, SolidJS and Svelte, and all of them work out-of-the-box after `mitosis build` runs, without additional build steps (save for very few add-ons, like [this solidjs index.jsx](https://github.com/BuilderIO/builder/blob/0cccd30e372e8d0784ce2452fec65cc2889b87ee/packages/sdks/overrides/solid/src/solid-index.jsx)). All we need to do is provide the correct `package.json` `exports`/`main`/`module`/etc. fields for each output.

We have not yet outputted the SDK to Angular/Qwik, so what I'm saying might not apply to that, but I do want to point out that so far, our experience is that most of these modern frameworks (react-native, solid, svelte, vue2, vue3) do not need additional steps to work on top of `mitosis build`!

Curious if you've encountered any limitations/edge cases with Vue for e2e tests that might suggest otherwise.

> Mitosis output is “source”

Agreed 💯! Mitosis output is fundamentally just source code for that particular output 👍🏽

samijaber (approved):
Merging this since it's a net positive, discussed w/ @kylecordes and will have a followup PR accounting for the discussion in the comments (moving some source stuff back to being built in `e2e-app`)

PR #505, opened 2022-06-21
fix embedded Builder preview URL

steve8708:
when we took down the firebase project it broke the
URL the preview was loading

note that the Builder preview still isn't getting content updates, so that's a separate issue


samijaber (approved):

samijaber:
> note that the Builder preview still isn't getting content updates, so that's a separate issue

Weird, I see it working for me! It takes a few seconds, but it definitely works when switching between the templates, and making changes: https://www.loom.com/share/5127d019cff54477bc83271fd05f1f15

What are you seeing?

samijaber:
Oh just understood what you're saying: the preview in the bottom isn't updating. Will take a look

PR #513, opened 2022-06-23
Generate and bundle-for-package in e2e-app

kylecordes:
The goal of this PR is to revamp things so that we prepare output for all supported targets, in e2e-app, ready to be consumed as packages from the test harnesses.

For several of the targets this is close to a no-op, for others there will be some code that will probably eventually move into Mitosis core.


kylecordes:
@samijaber reminder to look when you get a chance

samijaber:
@kylecordes I'm surprised by the difficulties you have with `vue-3`. I have an upcoming PR here for our Builder SDK: https://github.com/BuilderIO/builder/pull/1092/files

And we are generating a vue-3 SDK that is imported just fine in example source code, without any additional bundling/compilation step.

It might very well be that the right way to do things for Vue 2/3 is to add a bundling step, but I'm surprised that you're hitting that for such a simple scenario here yet I haven't 🤔.

kylecordes:
@samijaber Not a difficulty - rather, a next step. Your PR for the builder SDK does the same thing as Papasani - it bundles and ships the JS output, it doesn’t only ship the Vue files.

samijaber:
> Your PR for the builder SDK does the same thing as Papasani - it bundles and ships the JS output, it doesn’t only ship the Vue files.
Hmm, I might be misunderstanding what you mean by "bundles", because the process for building a new version of the builder SDK is to do:

- `mitosis build`
- followed by `npm version prerelease &&  npm publish --tag dev` for each individual SDK.

There is no additional bundling step done after `mitosis build`. What are you referring to when you say "bundles and ships the JS output"? the fact that we have `exports` entry points in our `package.json`?

kylecordes:
@samijaber Ah, I was looking at your changes in the examples rather than at the packages/sdks/output/vue stuff. I will have to study this more, to understand why the downstream consumers of this code get good results with just the .vue files, contrary to what research suggested.

PR #536, opened 2022-07-03
fix(qwik): output generation

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 


packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -59,9 +58,12 @@ export const componentToQwik =
             [component.propsTypeRef || 'any'],
           ),
         ]),
+        true,
+        true,
       );
       file.exportDefault(component.name);
       emitStyles(file, css);
+      file.exportConst('COMPONENT', JSON.stringify(component, null, 2));
```
samijaber, line 66: It doesn't quite make sense to me that we are exporting the Mitosis JSON object for each qwik object. What is the purpose of doing this? Is it for debugging purposes? Can we remove it?

mhevery: Opps, yes that is debugging, Let me remove it.

samijaber (commented):
Just need to update the snapshots again @mhevery

mhevery:
done

packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap
```diff
@@ -235,16 +235,12 @@ import {
   useScopedStyles$,
   useStore,
 } from \\"@builder.io/qwik\\";
-function underscore_fn_name(props, state) {
+export const underscore_fn_name = function underscore_fn_name(props, state) {
```
samijaber, line 238: Not a big problem, but it's weird that we're generating `export const fnName = function fnName(){}`. Seems like a mistake. Don't we just want `export function fnName(){}`?

mhevery: ohh, that is because it is sitting on code that can generate both CJS and MJS so it has to be `export const fnName =` The fact that `fnName` is there twice, we could fix that. 

But the way we are handling these functions is one big hack, but we need better support from Qwik and should be redone.

Could we do all of that once Qwik handles it better?

samijaber: Sounds good, thanks for the context. Yeah we can handle this later, the current code is valid and works just fine, I was just curious to understand more 👍🏽

samijaber (approved):

PR #537, opened 2022-07-04
Feature/jsx runtime types

### LINKED ISSUE #523: Fix/implement JSX Runtime types
Currently, Mitosis components are not able to get correct JSX runtime types. We are doing things the old school way: exporting a global JSX interface. This conflicts with react types.

We should do things the new way, by providing a JSX interface that is exported at the `jsx-runtime` `exports` of `package.json`, and then consumers will need to add `jsxImportSource: '@builder.io/mitosis'` to `tsconfig.json`

https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
https://www.typescriptlang.org/tsconfig#jsxImportSource

Inspiration from other frameworks:
https://preactjs.com/guide/v10/typescript/#typescript-configuration
https://unpkg.com/browse/@builder.io/qwik@0.0.25/jsx-runtime.d.ts
https://unpkg.com/browse/solid-js@1.4.4/types/jsx.d.ts
https://unpkg.com/browse/solid-js@1.4.4/jsx-runtime.d.ts

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fixed #523 

### DIFF
--- a/packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
+++ b/packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
@@ -1,3 +1,4 @@
+/** @jsxImportSource @builder.io/mitosis */
 import { useStore, useMetadata, onMount } from '@builder.io/mitosis';
 
 useMetadata({
--- a/packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
+++ b/packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
@@ -1,4 +1,3 @@
-/** @jsxImportSource @builder.io/mitosis */
 import { useStore, useMetadata, onMount } from '@builder.io/mitosis';
 
 useMetadata({
--- a/packages/core/jsx-runtime.d.ts
+++ b/packages/core/jsx-runtime.d.ts
@@ -1,1 +1,1 @@
-export * from "./src/jsx-types"
+export * from './src/jsx-types';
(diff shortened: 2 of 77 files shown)


packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
```diff
@@ -1,3 +1,4 @@
+/** @jsxImportSource @builder.io/mitosis */
```
samijaber, line 1: Is it not possible to add `jsxImportSource` once in the TSConfig of `packages/core`: https://github.com/BuilderIO/mitosis/blob/6e0a1ff7f090ba9d1b9a0a952c1b45275094e438/packages/core/tsconfig.json#L1-L21

This would avoid having to provide it manually for every single test fixture!

samijaber (commented):
Absolutely fantastic! 

The only thing I want to look into is setting up the JSX import source in our snapshot tests as well to avoid needing to add the line for each test manually, but if that is hard to do then we can merge without it for now 🙏🏽

packages/core/src/__tests__/data/basic-outputs-meta.raw.tsx
decadef20: I'll try it.

samijaber (approved):
@decadef20 You're amazing! Thank you 🙏🏽

samijaber:
Just realized we need to add some docs for this https://github.com/BuilderIO/mitosis/issues/545

PR #547, opened 2022-07-08
Node 16

kylecordes:
Node 16 became LTS in Oct 2021.

In other work, an eslint error occurs with Node 14, and moving to current LTS seems wiser than tracking down lint issues on older Node.


kylecordes:
(missed .tool-versions the first time, forgot about the asdf users among us!)

samijaber (approved):
Thank you!

PR #554, opened 2022-07-08
Move all E2E support under /e2e/

kylecordes:
First step in making the E2E less tangled with the production code.


kylecordes:
There are a bunch more ways to keep going on the E2E…   But since this moves ~all of the files involved, it's probably good to merge this one as is, zero semantic change.  @samijaber

.prettierignore
```diff
@@ -6,12 +6,13 @@ __snapshots__
...
-packages/e2e-qwik/server/**
-packages/e2e-app-qwik-output/src/e2e-app/**
-packages/e2e-app-qwik-output/lib/**
+e2e/**/output/**
+e2e/e2e-app-vue3-output/vue3/src/**
+e2e/e2e-angular/angular_src/app/lib/**
+e2e/e2e-qwik/server/**
+e2e/e2e-app-qwik-output/src/e2e-app/**
+e2e/e2e-app-qwik-output/lib/**
+packages/e2e-*/**
```
samijaber, line 15: This path no longer matches anything! Do we actually need it?

samijaber (approved):

PR #557, opened 2022-07-11
E2E matrix testing with regression detection

kylecordes:
Goals:

* Multiple examples apps (cases), so we can see progress per-target
* Matrix test all example apps on all targets
* Have a list of what should pass
* Fail build only on regression - so we can have test cases for future Mitosis work

For upcoming PRs:

* Single Playwright invocation to do all the tests, more efficient than one per case
* Custom reporter to produce a table of results (target vs feature) for easier reading


e2e/e2e-app/package.json
```diff
@@ -4,21 +4,22 @@
...
     "mitosis-svelte": "mitosis build svelte",
     "mitosis-swift": "mitosis build swift",
-    "mitosis-template": "echo NOT SUPPORTED YET mitosis build template",
     "mitosis-vue2": "mitosis build vue2",
     "mitosis-vue3": "mitosis build vue3",
-    "build": "concurrently \"npm:mitosis-*\" --max-processes 3 --group || echo continuing"
+    "mitosis-webcomponent": "mitosis build webcomponent",
+    "build": "concurrently \"npm:mitosis-*\" --max-processes 3 --group || echo continuing",
+    "e2e": "../../node_modules/.bin/playwright test",
+    "report": "../../node_modules/.bin/playwright show-report"
```
samijaber, line 22: 🔮

kylecordes: BTW, if I am successful in realizing the e2e vision, I might be emboldened to try again at getting yarn+workspaces+typescript to cooperate in such a way to enable the more efficient yarn mode that would get rid of these awful ../../node_modules.

samijaber: Having that Matrix will be such a huge win, I can't wait

e2e/e2e-angular/generate_component_list.js
```diff
@@ -0,0 +1,22 @@
...
+// GENERATED FILE
+
+${components.map((p, index) => `import Comp${index} from '${p}';`).join('\n')}
+
+export const components = [${components.map((p, index) => 'Comp' + index).join(', ')}];
+`;
+
+writeFileSync('angular_src/app/mitosis-component-list.ts', output);
+
+// mitosis-component-list;
```
samijaber, line 22: It feels like this should be added to `mitosis build`'s logic, as an additional step if Angular was one of the outputs! What do you think?

e2e/e2e-app/overall-result-must-pass.json
```diff
@@ -0,0 +1,16 @@
...
+  { "caseName": "01-one-component", "target": "e2e-svelte", "ok": true },
+  { "caseName": "01-one-component", "target": "e2e-vue2", "ok": true },
+  { "caseName": "01-one-component", "target": "e2e-vue3", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-angular", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-qwik", "ok": false },
+  { "caseName": "02-two-components", "target": "e2e-react", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-solid", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-svelte", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-vue2", "ok": true },
+  { "caseName": "02-two-components", "target": "e2e-vue3", "ok": true }
```
samijaber, line 15: Neat!

I think having this be a JS dictionary might be more convenient: type-checking wise and simplicity-wise:

```ts
enum Case {
  One = '01-one-component',
  Two = '02-two-components',
}

enum TestTarget {
  Angular = 'e2e-angular',
  Qwik = 'e2e-qwik',
  React = 'e2e-react',
  Solid = 'e2e-solid',
  Svelte = 'e2e-svelte',
  Vue2 = 'e2e-vue2',
  Vue3 = 'e2e-vue3',
}

interface Entry {
  ok: boolean;
}

export const CASES: Record<Case, Record<TestTarget, Entry>> = {
  '01-one-component': {
    'e2e-angular': { ok: true },
    'e2e-qwik': { ok: true },
    'e2e-react': { ok: true },
    'e2e-solid': { ok: true },
    'e2e-svelte': { ok: true },
    'e2e-vue2': { ok: true },
    'e2e-vue3': { ok: true },
  },
  '02-two-components': {
    'e2e-angular': { ok: true },
    'e2e-qwik': { ok: false },
    'e2e-react': { ok: true },
    'e2e-solid': { ok: true },
    'e2e-svelte': { ok: true },
    'e2e-vue2': { ok: true },
    'e2e-vue3': { ok: true },
  },
};
```

Benefits:
- guarantees no typos in case names or targets
- guarantees no missing combinations of target x test

Could also have the nesting swapped (first targets, and then the cases). Doesn't really matter too much.

We could even have the list of cases flow from the mitosis app/playwright test, and the list of targets flow from the build config, but that's an overkill optimization at this moment IMO!

kylecordes: Typed data is great! I started with the simplest way to get a must-pass list off the ground. I deferred more thought about it until the matrix is:

1. A bit larger, perhaps at least 4 cases
2. Up and running in CI
3. Displaying a results table somehow

The list of must-pass could be a bit simpler - just the list of what must pass, omitted the ok flag.

I'm hoping to reverse the logic of this. Instead of a long list of must-pass, a short list of temporarily-allowed-to-fail. I think I'll do that next step in the next PR.

kylecordes:
I think this will resolve #430 !

(Lots of followup items in the E2E project here on Github)

samijaber (approved):
👌🏽

Issue #544, opened 2022-07-06
Angular and Svelte style properties are not transformed


CKGrafico:
### I am interested in helping provide a fix!

No

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [X] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://codesandbox.io/s/hungry-cache-trlkvp

### Expected Behaviour

in angular `[style]="obj"` should be `[ngStyle]="obj"` and in svelte is more complex https://svelte.dev/tutorial/style-directive

### Actual Behaviour

not doing anything

### Additional Information

not sure if you want to cover this in mitosis

CKGrafico:
Sorry to many days without creating issues in your repo 😅 I’m on my phone but should be something like this.

https://mitosis.builder.io/?outputTab=react&code=import+%7B+useStore+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+state+%3D+useStore%28%7B%0A++++name%3A+%22Steve%22%2C%0A++++customStyles%3A%0A++++%7B+color%3A+%27red%27%7D%0A++%7D%29%3B%0A++%0A++return+%28%0A++++%3Cdiv+styles%3D%7Bstate.customStyles%7D%3E%0A++++++%3Cinput%0A++++++++css%3D%7B%7B%0A++++++++++color%3A+%22red%22%2C%0A++++++++%7D%7D%0A++++++++value%3D%7Bstate.name%7D%0A++++++++onChange%3D%7B%28event%29+%3D%3E+%7B+state.name+%3D+event.target.value+%7D%7D%0A++++++%2F%3E%0A++++++Hello%0A++++++%7Bstate.name%7D%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D


basically you’re doing nothing (as usual in any property) but the property `style` has some special scenarios as I’ve explained above.

samijaber:
- For Angular: it should be a 1-line addition to this map to map `style` to `ngStyle`:

[...]

PR #559, opened 2022-07-12
Feature/angular svelte style

decadef20:
## Description

Add a short description of:
- what changes you made, 
#544 
- why you made them, and 


decadef20:
It's so weird. All tests pass on my machine and I pushed. I don't know why CI failed.  Any suggestions here? @samijaber

samijaber:
@decadef20 You might need to run `yarn` to get the newest version of the svelte `prettier` plugin

packages/core/src/generators/svelte.ts
```diff
@@ -299,6 +309,19 @@ const FUNCTION_HACK_PLUGIN: Plugin = () => ({
...
+    const child = children[i];
+    if (child.bindings.style || child.properties.style) {
+      return true;
+    } else if (child.children.length) {
+      return hasStyleObject(child.children);
+    }
+  }
+
+  return false;
+};
```
samijaber, line 323: We use `traverse` for this type of logic in our codebase. See this example:

https://github.com/BuilderIO/mitosis/blob/9fc4c5d26eb02ba9c933ee9726eada8699b1659d/packages/core/src/helpers/styles/helpers.ts#L16-L29

It's simpler and avoids the need for correctly performing recursion.

samijaber: NOTE: The function I'm sharing is checking for node.bindings.css, so you can't reuse it, but you can write a similar one with a different check

samijaber (approved):
Lovely!

packages/core/src/generators/svelte.ts
decadef20: Okay

Issue #530, opened 2022-06-30
BUG: string props defined like {"$VAL"} are considered bindings


samijaber:
**Scope**
- [x] This impacts ALL Mitosis outputs

**To Reproduce**
A link to a https://mitosis.builder.io/ fiddle containing the bug: [link](https://mitosis.builder.io/?outputTab=json&code=export+default+function+RenderBlocks%28props%29+%7B%0A++return+%28%0A++++%3Cdiv%0A++++++val1%3D%22ok%22%0A++++++val2%3D%7B%22ok%22%7D%0A%2F%3E%0A++%29%3B%0A%7D%0A)

**Expected behavior**
`val2` should be under `properties`, not `bindings`

PR #560, opened 2022-07-12
fix: string props defined like {"$VAL"} are considered bindings

decadef20:
## Description

Add a short description of:
- what changes you made, 
#530 
- why you made them, and 


samijaber:
@decadef20 You also need to update prettier with `yarn` here to fix the tests

samijaber (approved):
🥇

PR #563, opened 2022-07-13
refactor: add nodeHasAnyStyles helper

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/helpers/styles/helpers.ts
+++ b/packages/core/src/helpers/styles/helpers.ts
@@ -7,18 +7,39 @@ import json5 from 'json5';
 import { pickBy } from 'lodash';
 import { dashCase } from '../dash-case';
 
-export const nodeHasStyles = (node: MitosisNode) => {
+export const nodeHasBindingStyles = (node: MitosisNode) => {
   return Boolean(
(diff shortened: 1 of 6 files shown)


packages/core/src/helpers/styles/helpers.ts
```diff
@@ -7,18 +7,39 @@ import json5 from 'json5';
...
+      if (nodeHasBindingStyles(item)) {
+        hasStyles = true;
+        this.stop();
+      }
+    }
+  });
+  return hasStyles;
+};
+
+export const hasAnyStyles = (component: MitosisComponent) => {
```
samijaber, line 37: The problem here is now we use the word `styles` to mean `node.bindings.css` and also `node.bindings.style`/`node.properties.style`, which is confusing.

I think we should rename:

- `hasAnyStyle` -> `hasStyle`
- `nodeHasAnyStyles` -> `nodeHasStyle`
- `hasBindingsStyles` -> `hasCss`
-  `nodeHasBindingStyles` -> `nodeHasCss`

😅

decadef20: Agree with you. Updated

PR #567, opened 2022-07-14
Fix: Svelte onUpdate deps

samijaber:
## Description

missing some pre-processing for onUpdate deps array

### DIFF
--- a/packages/core/src/generators/svelte.ts
+++ b/packages/core/src/generators/svelte.ts
@@ -65,18 +65,16 @@ const mappers: {
     }
 
     return `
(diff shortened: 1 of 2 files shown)


packages/core/src/generators/svelte.ts
```diff
@@ -472,7 +454,7 @@ export const componentToSvelte =
                     function ${fnName}() {
                       ${hookCode}
                     }
-                    $: ${fnName}(...${deps})
+                    $: ${fnName}(...${stripStateAndProps(deps, options)})
```
samijaber, line 457: this is the fix

PR #571, opened 2022-07-14
Qwik 0.0.36 E2E, now uses build/serve

kylecordes:
Using the updated Qwik allows the E2E to use the Qwik library output, a closer match to intended downstream use.


kylecordes:
I'm pretty sure the fiddle error is unrelated to this pr... hmmm.

PR #574, opened 2022-07-18
fix(qwik): fixes to better support Qwik SDK

mhevery:
- Don't flatten state getters into state. This allows the state object to be removed in most cases, so that Qwik does not need to serialize it unnecessarily.
- Support metadata flag to generate light components. This allows us to flag which components are light. In our case it will be the `<RenderBlock>` components.
- Minor bug-fixes around the way JSX is generated.


samijaber (approved):

mhevery:
Closing in favor of https://github.com/BuilderIO/mitosis/pull/600

PR #586, opened 2022-07-22
Pr qwik add ids

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/generators/qwik/jsx.ts
+++ b/packages/core/src/generators/qwik/jsx.ts
@@ -165,22 +165,24 @@ function rewriteHandlers(
   const outBindings: { [key: string]: { code: string; arguments?: string[] } } = {};
   for (let key in bindings) {
     if (Object.prototype.hasOwnProperty.call(bindings, key)) {
-      let { code: binding } = bindings[key]!;
+      let bindingExpr: string | undefined = bindings?.[key]?.code;
       let handlerBlock: string | undefined;
-      if (binding != null) {
+      if (bindingExpr != null) {
         if (key == 'css') {
           continue;
-        } else if ((handlerBlock = handlers.get(binding))) {
-          key = `${key}Qrl`;
-          binding = invoke(file.import(file.qwikModule, 'qrl'), [
+        } else if ((handlerBlock = handlers.get(bindingExpr))) {
+          key = `${key}$`;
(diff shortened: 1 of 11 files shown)


packages/core/src/generators/qwik/jsx.ts
```diff
@@ -165,22 +165,25 @@ function rewriteHandlers(
   const outBindings: { [key: string]: { code: string; arguments?: string[] } } = {};
   for (let key in bindings) {
     if (Object.prototype.hasOwnProperty.call(bindings, key)) {
-      let { code: binding } = bindings[key]!;
+      let bindingExpr: string | undefined =
+        typeof bindings[key] == 'object' ? bindings[key]?.code : (bindings[key] as any);
```
samijaber, line 169: Weird: why do we need this change? `bindings[key]` should _always_ be a `{ code: string; arguments?: string[] } | undefined`. 

To avoid the `!` assertion, you can do this instead:

```suggestion
      let bindingExpr: string | undefined = bindings?.[key]?.code
```

samijaber (approved):

packages/core/src/generators/qwik/jsx.ts
mhevery: > Weird: why do we need this change? bindings[key] should always be a { code: string; arguments?: string[] } | undefined.

Yes, it should be, but it is not in some cases. 🤷 

Will incorporate your suggestion.

mhevery: Actually, your suggestion changes the meaning of the code so I don't think I can take it as is.

samijaber: > Yes, it should be, but it is not in some cases. 🤷

I guess that's what I'm trying to dig into: can you give an example of when this is happening? That should be impossible with the current version of Mitosis JSX parser.

Only way you're encountering this is if you're reading some old Mitosis component JSON that you've hardcoded/stored somewhere. The old format was `bindings: string`, but we've completely replaced that everywhere. 

My guess is you have some test specs that are stuck with the old format?

mhevery: OK switched to your case, and all tests pass. I think I was just being paranoid.

samijaber (approved):

PR #588, opened 2022-07-24
E2E runs in a single invocation of Playwright

kylecordes:
Uses a Playwright improvement that just shipped in 1.24.0.

This speeds E2E slightly, and brings it a step closer to the intended vision.


samijaber (approved):
Amazing! This testing framework keeps getting cleaner

kylecordes:
Yes - I keep looking for small ways to make it cleaner, to compensate for the inevitable big bump in complexity to support the matrix (N cases * M targets) testing coming up.

PR #589, opened 2022-07-24
E2E with 2 components - WIP don't merge

kylecordes:
Works correctly with all supported targets except Qwik.

Angular output sometimes has uninitialized inputs, which Angular accepts at runtime - needs a tsconfig setting to accept at compile time.


kylecordes:
I see Misko has a PR with Qwik output fixes - I wonder if those will fix the error here.

kylecordes:
With the latest Qwik and mentioned PRs from Misko merged, problem still happens.

The first (perhaps not all) is now in issue #592

kylecordes:
Closing this, because the matrix E2E (which can track the progress of problems like this without breaking the build) is ~ready.

PR #593, opened 2022-07-26
fix: consistent `bindings: {code: string}`

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/parsers/builder.ts
+++ b/packages/core/src/parsers/builder.ts
@@ -62,12 +62,15 @@ const verifyIsValid = (code: string): { valid: boolean; error: null | Error } =>
   return { valid: false, error: null };
 };
 
-const getActionBindingsFromBlock = (block: BuilderElement, options: BuilderToMitosisOptions) => {
+const getActionBindingsFromBlock = (
+  block: BuilderElement,
+  options: BuilderToMitosisOptions,
+): MitosisNode['bindings'] => {
   const actions = {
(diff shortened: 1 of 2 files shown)


packages/core/src/parsers/builder.ts
```diff
@@ -872,3 +879,20 @@ export const builderContentToMitosisComponent = (
 
   return componentJson;
 };
+function mapBuilderBindingsToMitosisBindingWithCode(
+  bindings: { [key: string]: string } | undefined,
+): MitosisNode['bindings'] {
+  const result: MitosisNode['bindings'] = {};
+  bindings &&
+    Object.keys(bindings).forEach((key) => {
+      const value: string | { code: string } = bindings[key] as any;
```
samijaber, line 888: Surprised that we needed this: I thought `BuilderElement` bindings will always be `bindings: { [key: string]: string }`, so how could the value be a `{ code: string }`? Did you encounter that possibility?

samijaber (approved):
Types to the rescue!

packages/core/src/parsers/builder.ts
samijaber: If my understanding is correct, this is the only possibility here:

```suggestion
      const value: string = bindings[key]
```

mhevery: Yes the tests fail without this.

PR #597, opened 2022-07-27
Add return outside of function.

alujs:
onUpdate(() => {
    console.log("dddhi");
    return () => {
      console.log("hello");
    };
  }, [rootRef, otherRef]);

Causes an error  SyntaxError: unknown: 'return' outside of function. (3:2) 

This "fixes" it. Angular will need to be updated.


samijaber (approved):
@alujs In a separate PR, do you mind updating one of the hook tests to have a return function to show this new feature?

PR #600, opened 2022-07-30
fix(qwik): fixes to better support Qwik SDK

mhevery:
- Don't flatten state getters into state. This allows the state object
  to be removed in most cases, so that Qwik does not need to serialize
  it unnecessarily.
- Support metadata flag to generate light components. This allows us
  to flag which components are light. In our case it will be the
  `<RenderBlock>` components.
- Minor bug-fixes around the way JSX is generated.

## Description

Add a short description of:
- what changes you made, 
- why you made them, and 


packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -6,17 +6,35 @@ import { renderJSXNodes } from './jsx';
...
 import { babelTransformExpression } from '../../helpers/babel-transform';
 import { BaseTranspilerOptions, Transpiler } from '../../types/transpiler';
-import { MitosisNode } from '../../types/mitosis-node';
 
 Error.stackTraceLimit = 9999;
 
 // TODO(misko): styles are not processed.
 
-const DEBUG = false;
+const DEBUG = true;
```
samijaber, line 14: You forgot to switch this back

packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -317,22 +387,24 @@ function stateToMethodOrGetter(state: Record<string, any>): Record<string, 'meth
...
-    }
-    if (key.startsWith('on') && binding?.code) {
-      const args = binding?.arguments || [];
-      binding.code = `(${args.join(',')}) => ${binding.code}`;
-      delete node.bindings[key];
-      node.bindings[key + '$'] = binding;
+/**
+ * Return a top-level element for the component.
+ *
+ * WHAT: If the component has a single root element, than this returns the element name.
```
samijaber, line 393:
```suggestion
 * WHAT: If the component has a single root element, then this returns the element name.
```

packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -317,22 +387,24 @@ function stateToMethodOrGetter(state: Record<string, any>): Record<string, 'meth
...
+ *
+ * WHY: This is useful to pull the root element into the component's host and those saving unnecessary wrapping.
```
samijaber, line 395: Did you mean this instead?

```suggestion
 * WHY: This is useful to pull the root element into the component's host and thus save unnecessary wrapping.
```

samijaber (approved):

packages/core/src/generators/qwik/component-generator.ts
mhevery: fixed

packages/core/src/generators/qwik/component-generator.ts
mhevery: fixed

packages/core/src/generators/qwik/component-generator.ts
mhevery: fixed

PR #614, opened 2022-08-05
Fix/angular this

alujs:
Updated types for aria purposes ripped directly from the react types.

Updated stripStateAndPropsRefs to properly deal with embedded functions and state functions in Angular so they have the correct 'this.' when being used. 

Also updated the regex to find more instances of state variables being used and prepend 'this'.

### DIFF
--- a/packages/core/src/helpers/strip-state-and-props-refs.ts
+++ b/packages/core/src/helpers/strip-state-and-props-refs.ts
@@ -4,6 +4,7 @@ export type StripStateAndPropsRefsOptions = {
   includeState?: boolean;
   contextVars?: string[];
   outputVars?: string[];
+  stateVars?: string[];
   context?: string;
   domRefs?: string[];
 };
@@ -26,6 +27,7 @@ export const stripStateAndPropsRefs = (
   const outputVars = options?.outputVars || [];
   const context = options?.context || 'this.';
   const domRefs = options?.domRefs || [];
+  const stateVars = options?.stateVars || [];
 
   if (contextVars.length) {
     contextVars.forEach((_var) => {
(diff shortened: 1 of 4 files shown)


packages/core/src/helpers/strip-state-and-props-refs.ts
```diff
@@ -66,7 +68,19 @@ export const stripStateAndPropsRefs = (
...
+        '$1' + 'this.' + _var + '$2',
+      );
+    });
+  }
+  if (stateVars.length) {
+    stateVars.forEach((_var) => {
+      newCode = newCode.replace(
+        // determine expression edge cases - https://regex101.com/r/iNcTSM/1
+        new RegExp(
+          `(?!^${_var}|^ ${_var})(?<!function|get)(^|\\n|\\r| |;|\\(|\\[|!|,)${_var}(\\?\\.|,|\\.|\\(| |;|\\)|\\]|$)`,
```
samijaber, line 81: Can we share some of the Regex between this and the above condition, and explain a bit how this one is different? Want to avoid increasing the number of unwieldy regex expressions in our codebase, and if they're necessary then:
- share the logic between them
- document them

Something like this (but with better names than `regex1` and `regex2`):
```ts
// determine expression edge cases - https://regex101.com/r/iNcTSM/1
const regex1 = `(^|\\n|\\r| |;|\\(|\\[|!|,)${_var}(\\?\\.|,|\\.|\\(| |;|\\)|\\]|$)`;

// domRefs replace
newCode = newCode.replace(new RegExp(regex1, 'g'), '$1' + 'this.' + _var + '$2');

// stateVars replace
const regex2 = `(?!^${_var}|^ ${_var})(?<!function|get)${regex1}`;
newCode = newCode.replace(new RegExp(regex2, 'g'), '$1' + 'this.' + _var + '$2');
```

samijaber (approved):
👍🏽

PR #615, opened 2022-08-06
Store short allowed-failure list for E2E

### LINKED ISSUE #602: E2E: replace must-pass with allow-to-fail
The initial E2E matrix implementation has a must-pass list, to prevent regressions.

That list will grow, be unwieldily, warrant more complexity.

Replace it with a much shorter allow-to-fail list - which is likely to remain small even as Mitosis as its E2E suite grow.

kylecordes:
Store only the short list of allowed failures instead of the long list of required-to-pass. This makes it possible to merge cases where Mitosis doesn't yet work perfectly on all targets, while keeping the stored list very short.

Closes #602


e2e/e2e-app/e2e.ts
```diff
@@ -50,6 +60,25 @@ async function readSummary(caseName: string): Promise<Entry[]> {
...
+
+  const didntPass = allResults.filter((e) => !e.ok);
+  const regressions = didntPass.filter(
+    (dp) => !allowFailures.find((af) => dp.caseName == af.caseName && dp.target == af.target),
+  );
+
+  if (regressions.length > 0) {
+    console.error('E2E regressions, these formerly passed!', regressions);
+    throw new Error('E2E regressions: ' + regressions.length);
+  }
```
samijaber, line 79: I think we'll also want to explicitly _throw_ if a previously allowed failure starts to pass! That way, we require/remind devs to remove the entry from `allow-failures.json`

samijaber (approved):

PR #621, opened 2022-08-09
Marko support (beta)

steve8708:


packages/core/src/generators/marko/generate.ts
```diff
@@ -0,0 +1,308 @@
...
+    str += ` ${key}="${value}" `;
+  }
+  for (const key in json.bindings) {
+    const { code, arguments: cusArgs = ['event'] } = json.bindings[key]!;
+    if (key === '_spread' || key === '_forName') {
+      continue;
+    }
+
+    if (key === 'ref') {
+      str += ` ref=((el) => this.${code} = el) `;
```
DylanPiercey, line 84: I think this part needs to be tweaked to use the [key](https://markojs.com/docs/class-components/#key) attribute which is the way to gather references in Marko.

It's string based keys (in Marko 5) so you have something like:

```marko
class {
  onMount() {
    console.log("got el", this.getEl("my-key");
  }
}

<div key="my-key"/>
```

packages/core/src/generators/marko/generate.ts
```diff
@@ -0,0 +1,308 @@
...
+    let jsString = dedent`
+    ${renderPreComponent({ component: json, target: 'marko' })}
+
+    class {
+        ${methodsString}
+
+        ${
+          !hasState
+            ? ''
+            : `onCreate() {
```
DylanPiercey, line 170: The `onCreate` method receives `input` into it, so you could change this to:

```
`onCreate(${thisHasProps ? 'input' : ''}) {`
```

packages/cli/src/build/build.ts
```diff
@@ -185,6 +186,8 @@ const getGeneratorForTarget = ({
       return componentToSvelte(options.options.svelte);
     case 'qwik':
       return componentToQwik(options.options.qwik);
+    case 'marko':
+      return componentToMarko(options.options.qwik);
```
samijaber, line 190:
```suggestion
      return componentToMarko(options.options.marko);
```

samijaber (approved):
👍🏽 

just the options error I commented on

packages/core/src/generators/marko/generate.ts
steve8708: oh perfect, will do that, thanks @DylanPiercey!

packages/core/src/generators/marko/generate.ts
steve8708: ah! perfect, thanks @DylanPiercey 

I'll remove this old ref logic for now and circle back on a full refs implementation based on this

steve8708: nvm went ahead and implemented, huge thanks for the tip @DylanPiercey this made it very easy to implement

PR #625, opened 2022-08-10
allow angular to reference external imported values in templates

steve8708:
also fix an import parsing bug getting the names and values backwards when using the `as` syntax

before: (would error that `links` is not defined on the component)
image

after: (fixed)
image


steve8708:
dang, these e2e tests are amazing @kylecordes! Just caught a bug that none of our other tests caught

steve8708:
fixed a bunch of other random stuff here too, all found from working on the benchmarking tool

PatrickJS (approved):

PatrickJS:
good enough

PR #626, opened 2022-08-11
fix(vue): use named slots

Szymon-dziewonski:
## Description
**Why change is needed?**
`slotIcon={<svg></svg>}` is compiled into `{{slotIcon}}` which isn’t slot, this result in not being able to pass HTML element into slot, it will be displayed as `&lt;svg&gt;...&lt;/svg&gt;`. Named slots should work same as default slot (`props.children`) which is compiled to `<slot></slot>` , therefore named slot should be transpiled into `<slot name="icon"></slot>`. 

**What have been changed?**
I changed parsing prefixed `slotIcon` prop so it will be rendered as `<slot name="icon"/>`
Minor change, self closing `<slot/>` tag instead of `<slot></slot>`

Thank you :)

### DIFF
--- a/packages/core/src/generators/vue.ts
+++ b/packages/core/src/generators/vue.ts
@@ -36,6 +36,7 @@ import { methodLiteralPrefix } from '../constants/method-literal-prefix';
 import { OmitObj } from '../helpers/typescript';
 import { pipe } from 'fp-ts/lib/function';
 import { getCustomImports } from '../helpers/get-custom-imports';
+import { isSlotProperty, stripSlotPrefix } from '../helpers/slots';
 
 function encodeQuotes(string: string) {
   return string.replace(/"/g, '&quot;');
@@ -157,7 +158,7 @@ const NODE_MAPPERS: {
         ${
           isMitosisNode(json.meta.else)
             ? `
-            <template ${SPECIAL_PROPERTIES.V_ELSE}> 
+            <template ${SPECIAL_PROPERTIES.V_ELSE}>
               ${blockToVue(json.meta.else, options)}
             </template>`
             : ''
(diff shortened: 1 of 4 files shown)


packages/core/src/generators/vue.ts
```diff
@@ -362,6 +362,9 @@ export const blockToVue: BlockRenderer = (node, options, scope) => {
   }
 
   if (node.bindings._text?.code) {
+    const slotPrefixPhrase = 'props.slot';
+    if (node.bindings._text?.code?.substring(0, slotPrefixPhrase.length) === slotPrefixPhrase)
+      return `<slot name="${node.bindings._text?.code?.substring(slotPrefixPhrase.length).toLowerCase()}"/>`;
     return `{{${stripStateAndPropsRefs(node.bindings._text.code as string)}}}`;
```
samijaber, line 368: One small thing: we have `isSlotProperty` that you can reuse, and `stripStateAndPropsRefs` will remove `props.`. So I recommend an approach like this one:

```suggestion
  const textCode = node.bindings._text?.code;
  if (textCode) {
    const strippedTextCode = stripStateAndPropsRefs(textCode);
    if (isSlotProperty(strippedTextCode)) {
      return `<slot name="${textCode.substring('slot'.length).toLowerCase()}"/>`;
    }
    return `{{${strippedTextCode}}}`;
  }
```

We can actually also export SLOT_PREFIX from https://github.com/BuilderIO/mitosis/blob/78a6d882cd6a4a022d5910831d71b194a5b30c1d/packages/core/src/helpers/slots.ts#L1 to reuse it here, or add a `stripSlotPrefix` helper to the `slots.ts` helper and import it here :)

PatrickJS (approved):

samijaber:
@Szymon-dziewonski you also wanna make sure to lint by running `yarn run fmt:prettier` at the root of the monorepo. Looks like the linter is failing

packages/core/src/generators/vue.ts
Szymon-dziewonski: @samijaber thank you for your suggestions, code looks much better, I just simply wasn't aware of those utils. Nevertheless I've added `stripSlotPrefix` utils inside `slots.ts` helper file

PR #630, opened 2022-08-11
add a test for sub component path name for marko

steve8708:


packages/core/src/__tests__/marko.test.ts
```diff
@@ -33,4 +34,15 @@ describe('Marko format', () => {
 
 describe('Marko', () => {
   runTestsForTarget('marko', componentToMarko());
+
+  const exampleSubComponent = `
+    import SubComponent from './foo.lite'
+    export default function Example() {
+      return <SubComponent />
+    }
+  `;
```
samijaber, line 43: I'm trying to get in the habit of always adding every test to every target by default. That way we always get into the habit of adding coverage throughout the whole codebase.

If you can add it to this https://github.com/BuilderIO/mitosis/blob/d19939361a50ff422ba7546e0a7ec08b3f4692ad/packages/core/src/__tests__/shared.ts#L68

then it'll end up tested everywhere by default! you can just provide a string instead of adding a whole `.tsx` file, that should work too

steve8708: so the problem I ran into with this is ti must include '.lite' in the path, but our tests use .raw

I think we likely can/should replace our tests what have .raw to all use .lite to catch this stuff in the future. just requires changing the jest config to use raw loader for .lite I'm pretty sure\

I found a workaround in my other open PR, so will apply this there

PR #631, opened 2022-08-12
Preact support

steve8708:
also added more tests, some other bug fixes, and support for type imports


packages/core/src/generators/react.ts
```diff
@@ -51,10 +51,18 @@ export interface ToReactOptions extends BaseTranspilerOptions {
   stateType?: 'useState' | 'mobx' | 'valtio' | 'solid' | 'builder';
   format?: 'lite' | 'safe';
   type?: 'dom' | 'native';
+  preact?: boolean;
```
samijaber, line 54: wouldn't it make more sense to add `preact` to `type`?

```suggestion
  type?: 'react' | 'react-native' | 'preact';
```

steve8708: That was my initial impulse but actually it would’ve made the code more complicated and harder to maintain given preact is almost identical to dom with just a couple small exceptions regarding imports

steve8708: Also fwiw you can use preact with react native, so they aren’t mutually exclusive options

packages/core/src/parsers/jsx/jsx.ts
```diff
@@ -535,6 +536,20 @@ export function parseJsx(
...
+            for (const statement of path.node.body) {
+              if (isTypeImport(statement)) {
+                const importDeclaration = statement as babel.types.ImportDeclaration;
+                // Remove .lite from path if exists, as that will be stripped
+                if (importDeclaration.source.value.endsWith('.lite')) {
+                  importDeclaration.source.value = importDeclaration.source.value.replace(
+                    /\.lite$/,
+                    '',
+                  );
+                }
```
samijaber, line 548: I think this import path needs to be replaced with the correct extension for each framework, using `getFileExtensionForTarget`: https://github.com/BuilderIO/mitosis/blob/6d99f9bcf704067ef6a2965ec94fc6f66226a4de/packages/core/src/helpers/render-imports.ts#L26

samijaber: Actually nevermind, this is too complicated. Will deal with it when trying to add types to other generators...

samijaber (approved):

packages/core/src/parsers/jsx/jsx.ts
steve8708: yeah agreed, also I am leaning on the fact that I've seen bundlers allow importings './foo' when foo is actually foo.vue etc, based on the configuration, so this may or may not be an issue for some or all, so figured we can address it when needed

PR #632, opened 2022-08-14
581/svelte generate typescript code

### LINKED ISSUE #581: Svelte: Generate Typescript code

We would have to repeat step 3, but for Svelte: https://github.com/BuilderIO/mitosis/issues/372#issuecomment-1133471614

NOTE: `svelte` needs a `lang=ts` attribute added to its `<script>` to handle TS code

raymondmuller:
Closes #581 

## Description

Inject type information in the Svelte generator.

As the types and interfaces arrays (can) contain `export `, I have chosen to add a context="module" script block, as in Svelte you can't export types or interfaces in the normal script block. 

Screenshot 2022-08-14 at 15 45 47

This also avoids having to check and add lang="ts" to the regular script block, as the svelte compiler already detects it from the context="module" script block.

The prop declarations will look like this
``` 
export let text: ButtonProps["text"];
export let link: ButtonProps["link"];
export let openLinkInNewTab: ButtonProps["openLinkInNewTab"];
```

This is not how I would normally write my Svelte files I suppose, as I would just do 
```
export let text: string;
export let link: string;
export let openLinkInNewTab: boolean;
```

But that would require more complex code in the generator file.

And this seems to work and adds typescript autocompletion when working with the Svelte components.

### Possible TODO
If propsTypeRef contains multiple definitions (e.g. Person | never), should we detect so and get the type from the corresponding interface or do we just accept this constraint? 
Right now it defaults to the first definition (Person in the case of this example). Seems acceptable to me.


packages/core/src/generators/svelte.ts
```diff
@@ -401,7 +426,18 @@ export const componentToSvelte =
           if (name === 'children') {
             return '';
           }
-          return `export let ${name};`;
+
+          let propDeclaration = `export let ${name}`;
+
+          if (propsTypeRef && propsTypeRef !== 'any') {
+            propDeclaration += `: ${
+              isString(propsTypeRef) ? propsTypeRef.split(' |')[0] : propsTypeRef
```
samijaber, line 434: Since `propsTypeRef` is guaranteed to be a string here, we should be able to just do:
```suggestion
              propsTypeRef.split(' |')[0]
```

packages/core/src/generators/svelte.ts
```diff
@@ -384,11 +384,36 @@ export const componentToSvelte =
...
+    if (json.propsTypeRef?.startsWith('{')) {
+      propsTypeRef = `${json.name}Props`;
+
+      if (json.interfaces?.length) {
+        json.interfaces.push(`export interface ${propsTypeRef} ${json.propsTypeRef};`);
+      } else {
+        json.interfaces = [`export interface ${propsTypeRef} ${json.propsTypeRef};`];
+      }
+    }
+
```
samijaber, line 398: Awesome! For consistency across generated code, I think that we should move this logic into the JSX parser!

https://github.com/BuilderIO/mitosis/blob/3caff5b890260896e3cd916afeac147465b8f16e/packages/core/src/parsers/jsx/component-types.ts#L7-L23

We would change:

```ts
    return generate(paramIdentifier)
      .code.replace(/^let\sprops:\s+/, '')
      .replace(/;/g, '');
```

to:

```ts
    const generatedTypes = generate(paramIdentifier)
      .code.replace(/^let\sprops:\s+/, '')
      .replace(/;/g, '');

    if (generatedTypes.startsWith('{')) {
      const propsTypeRef = `${node.id?.name}Props`;

      context.builder.component.interfaces = [
        ...(context.builder.component.interfaces || []),
        `export interface ${propsTypeRef} ${generatedTypes}`,
      ];
    }

    return generatedTypes;
```

You would have to provide `context: Context` as an additional argument in that fn.

Let me know if this makes sense!

samijaber (commented):
Welcome to Mitosis! 😄 

I just have 1 suggestion around moving the logic you build into the general JSX parser, so that every generator benefits from your contribution!

raymondmuller:
Thanks @samijaber . I became aware of Mitosis last week and am super excited about this project (I'll expect to make some more pull requests soon).

Regarding this one, I have refactored the code based on your review (thanks for that). It all made perfect sense.

raymondmuller:
Forgot to run prettier again (updated now)

samijaber:
@raymondmuller You'll want to update snapshots for the tests now that you moved the props logic to the JSX parser.'

raymondmuller:
Done, sorry for that :)

samijaber (approved):
Absolutely fantastic. Thank you for your first contribution to Mitosis!! 🏆

samijaber:
@raymondmuller You have one more snapshot update to do 😄

raymondmuller:
@samijaber That is odd. I did that! I just ran `yarn snapupdate` in the core folder again and nothing is changing

raymondmuller:
@samijaber Figured out what was wrong. This branch wasn't up to date with the main branch. Should work now

PR #641, opened 2022-08-16
Fix/svelte

samijaber:
## Description

fix svelte TS code


packages/core/src/generators/svelte.ts
```diff
@@ -401,7 +401,7 @@ export const componentToSvelte =
     }
 
     str += dedent`
-      <script>
```
steve8708, line 404: I think we need this too, and potentially context=module (idk what that is)

https://github.com/BuilderIO/mitosis/pull/641/files#diff-33b0fc92f8d3ef8a691c9b46bf035c05827776f7271c39bb17fad5f52f9dd2c1L395

Seems like the logic up there is a little bit off

steve8708: but if we get that line in we will output types too

steve8708 (approved):

PR #646, opened 2022-08-17
Feat/default props

raymondmuller:
## Description

This will allow you to set default values for the props. The defaultProps object can be defined as a property on the component itself (as per the React docs : https://reactjs.org/docs/react-without-es6.html#declaring-default-props).

I've added 
- A method in props.ts that collects these defaultProps by checking for ExpressionStatements, doing some type checking, and then assigning those to a newly defined defaultProps JSONObject on the component. (note: I have added 'any' types there and could use some help defining the right types. I am not very experienced with Typescript so having a bit of a hard time there).
- DefaultProps to the react generator
- DefaultProps to the svelte generator
- DefaultProps to the vue generator
- A test (default-props.raw.tsx)


packages/core/src/generators/react.ts
```diff
@@ -711,6 +711,12 @@ const _componentToReact = (
       );
     }${isForwardRef ? ')' : ''}
 
+    ${
+      !json.defaultProps
+        ? ''
+        : `${json.name || 'MyComponent'}.defaultProps = ${json5.stringify(json.defaultProps)};`
+    }
```
samijaber, line 718: Hmm, this isn't entirely future-proof. See https://twitter.com/dan_abramov/status/1133878326358171650

However, I don't mind it as a v1. One long-term solution would be to generate the code like this:

```ts
//BEFORE
export default MyComponent(props: PropType) {
  //...rest of component
}

//AFTER
export default MyComponent(__props: PropType) {
  const props = { ...__props, ...json.defaultProps }
  //...rest of component
}
```

But I'm slightly concerned about how this could possibly cause re-renders.

Let's stick with your solution for now 👍🏽

samijaber (approved):
@raymondmuller This is huge! 🎉

PatrickJS:
Experian needs a better way to handle default props so this will help

PatrickJS:
@raymondmuller can you add angular support? if not I can have someone add it after this is merged
https://github.com/BuilderIO/mitosis/issues/648

raymondmuller:
Sure. Might be able to add it tonight. I think it would be good to get this merged already though. I have another branch that could use some of the code from this PR (Vue Types).  Can make a separate PR with default props in Angular.

PatrickJS:
@raymondmuller separate PR is fine, thanks 👍

samijaber:
We actually discussed this internally and we might opt for a `useDefaultProps` compile-away hook, as that is much more consistent with the Mitosis style than the `MyComponent.defaultProps` React style.

Will fire up a separate PR to make the change

PatrickJS:
yeah, the benefit of mitosis is that you can compile away everything for the ideal API rather than being restricted by the frameworking having to work without a compile step

steve8708:
Yeah def would want to use hooks here, for consistency/extensibility and ultimately (one day) composability 

Something like:

```tsx
import { useDefaultProps } from '@builder.io/mitosis'

useDefaultProps({ ... })

export default function MyComponent(props) { }
```

Which can play nicely with types:
```tsx
import { useDefaultProps } from '@builder.io/mitosis'

export type MyProps {}

useDefaultProps<MyProps>({ ... })

export default function MyComponent(props: MyProps) { }
```

Basically trying to use the idea of hooks that live outside of the component for metadata style stuff

PR #649, opened 2022-08-18
Feat/648 use default props for angular

### LINKED ISSUE #648: useDefaultProps for Angular
we need to support defaultProps for Angular
https://github.com/BuilderIO/mitosis/pull/646

this can be as simple as
`mitosis`
```typescript
export interface Props {
  myProp: string;
}

useDefaultProps<Props>({
  myProp: 'hi'
});
default export function MyComponent(props: Props) {
}

```

`angular`
```typescript
export interface Props {
  myProp: string;
}

const defaultProps = {
  myProp: 'hi'
};

@Component({
  selector: 'my-component'
})
export default class MyComponent {
  @Input() myProp: Props["myProp"] = defaultProps["myProp"];
}
```

raymondmuller:
## Description

Adds default props to the angular generator.
Closes #648

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/lit.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/lit.test.ts.snap
@@ -2440,6 +2440,10 @@ const spread = (properties) =>
 
 @customElement(\\"my-button\\")
 export default class Button extends LitElement {
+  createRenderRoot() {
+    return this;
+  }
+
   @property() link: any;
   @property() attributes: any;
   @property() openLinkInNewTab: any;
@@ -2447,6 +2451,8 @@ export default class Button extends LitElement {
 
   render() {
     return html\`
+        
+      }
         <my-fragment >
             
       \${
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/__snapshots__/lit.test.ts.snap
```diff
@@ -2440,13 +2440,19 @@ const spread = (properties) =>
 
 @customElement(\\"my-button\\")
 export default class Button extends LitElement {
+  createRenderRoot() {
+    return this;
+  }
```
samijaber, line 2445: Weird how this showed up in this PR's diff... 🤔 are you on the latest `main`?

raymondmuller: Yes, but the snapshots are not updated on the main branch :)

steve8708: ya current main is actually not passing build, prob bc two separate PRs that were each valid on their own got merged, so merging this one will fix it https://github.com/BuilderIO/mitosis/commit/48371deef473537fbf13e830d39840911f2c1956

That is the state we want the snapshots in (with the `createRenderRoot`) ^

samijaber (approved):
LGTM 👍🏽 

Tagging @PatrickJS in case he sees something

PatrickJS (approved):

PR #656, opened 2022-08-21
fix(svelte): no line-breaks between children

### LINKED ISSUE #310: Spacing issues around tags with dynamic text
## **Describe the bug**

I made an [extremely simple example of running Mitosis with Astro](https://github.com/arijs/astro-mitosis). The same Mitosis component has different rendering between frameworks. I guess we'd need a test suite for this. But below is an example:

The text of the input is rendered like this:
```jsx
    return <div>
        <input
            css={{
            color: "red",
            }}
            value={name}
            onChange={(event) => setName(event.target.value)}
        />
        Hello <span>{name}</span>!
        I can run in React, Vue, Solid, or Liquid!
    </div>;
```

Note that the dynamic content is rendered on a `<span/>` tag. I could have rendered it directly on the parent text but I wanted to specifically set it apart. I think the same issue happens without the `<span/>` tag.

## **To Reproduce**

1. Just render dynamic content on a tag with text around it, and see if the text respects the spaces in the source.

### If possible, a link to a https://mitosis.builder.io/ fiddle containing the bug:
There's not much point in this because the "rendered" output is a Builder editor and not rendered by the selected framework.

[Link to fiddle](https://mitosis.builder.io/?outputTab=react&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+%5Bname%2C+setName%5D+%3D+useState%28%22Steve%22%29%3B%0A%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%3Cinput%0A++++++++css%3D%7B%7B%0A++++++++++color%3A+%22red%22%2C%0A++++++++%7D%7D%0A++++++++value%3D%7Bname%7D%0A++++++++onChange%3D%7B%28event%29+%3D%3E+setName%28event.target.value%29%7D%0A++++++%2F%3E%0A++++++Hello+%3Cspan%3E%7Bname%7D%3C%2Fspan%3E%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D)

## **Expected behavior**

First, I expect all frameworks to render the same. Second, I expect the spacing around texts to represent what I wrote in the source. Specifically, to have a space if there is one in the source and to not have a space if there isn't one in the source.

## **Screenshots**

image

## **Additional context**

- **React and Solid:**
  No space after "Hello". Here the rendered source JSX has the word "Hello" on its own line. For it to work we need to output one of these options:
  - A space character when text is alone on the line:
    ```
      Hello{" "}
    ```
  - Make the text on the same line as the tag. I guess this is the easier option.
    ```
      Hello <span>{name}</span>!
    ```
    _Edit:_ Also, I almost overlooked it, but the text "Hello" also does not have a space _before_ it, having no space with the text input just before it.
- **Svelte:**
  This one is actually putting a space between the `<span/>` and the exclamation mark `!`. This space needs to be removed.
  ```
    Hello <span>{name}</span> !
  ```
- **Vue:**
  This one actually gets it right. No complaints here

micmro:
## Description

Add a short description of:
The line-breaks causes the browser to render spaces that were not intended to be there. Other generators (like React) also don't add these.

fixes #310

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/svelte.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/svelte.test.ts.snap
@@ -46,15 +46,11 @@ exports[`Svelte AdvancedRef 1`] = `
 
     <select name=\\"cars\\" id=\\"cars\\">
       <option value=\\"supra\\">GR Supra</option>
-
       <option value=\\"86\\">GR 86</option>
     </select>
   {/if}
-
   Hello
-
-  {lowerCaseName()}
-  ! I can run in React, Qwik, Vue, Solid, or Web Component!
+  {lowerCaseName()}! I can run in React, Qwik, Vue, Solid, or Web Component!
 </div>
 
 <style>
@@ -91,7 +87,6 @@ exports[`Svelte Basic 1`] = `
       name = myEvent.target.value;
     }}
   />
-
   Hello! I can run in React, Vue, Solid, or Liquid!
 </div>
 
@@ -154,9 +149,7 @@ exports[`Svelte Basic Context 1`] = `
 
 <div>
   {myService.method(\\"hello\\") + name}
-
   Hello! I can run in React, Vue, Solid, or Liquid!
-
   <input
     on:change={(event) => {
       onChange;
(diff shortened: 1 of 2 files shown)


samijaber (approved):
Thanks for the fix @micmro!

Issue #653, opened 2022-08-19
Vue: Generate Composition API


samijaber:
Add a `'api': 'options' | 'composition'` option to the Vue generator, and allow the generation of Composition API code.

samijaber:
@raymondmuller Are you interested at all in giving this a shot? If so, let me know and I can assign this to you, and provide any guidance you may need.

PR #665, opened 2022-08-24
Feat/653 vue generate composition api

### LINKED ISSUE #582: Vue: Generate Typescript code

We would have to repeat step 3, but for Vue: https://github.com/BuilderIO/mitosis/issues/372#issuecomment-1133471614

raymondmuller:
## Description

Closes issue #653 

Add composition API as an **option** to the Vue generator. It still defaults to the options API for now on this branch as this probably needs some more testing, though I think it is in a good state to get out there for others to try it and give feedback. It is my first time writing (part of a) generator here so I might have missed something :)

This also fixes #582 , at least for the composition API script. 
While Vue does support TypeScript usage with Options API, it is recommended to use Vue with TypeScript via Composition API as it offers simpler, more efficient and more robust type inference, and you get type based declarations.

This looks like: 

- Types only

``` 
interface Props {
  conditionA: boolean;
  conditionB: boolean;
}
const props = defineProps<Props>();
```

- With default values
```
export interface ButtonProps {
  attributes?: any;
  text?: string;
  link?: string;
  openLinkInNewTab?: boolean;
}
const props = withDefaults(defineProps<ButtonProps>(), {
  text: "default text",
  link: "https://builder.io/",
  openLinkInNewTab: false,
});
```

I chose to use the <script setup> as (apart from my own preference) it is the recommended syntax per the Vue docs 

**<script setup> is a compile-time syntactic sugar for using Composition API inside Single-File Components (SFCs). It is the recommended syntax if you are using both SFCs and Composition API. It provides a number of advantages over the normal <script> syntax:**
- More succinct code with less boilerplate
- Ability to declare props and emitted events using pure TypeScript
- Better runtime performance (the template is compiled into a render function in the same scope, without an intermediate proxy)
- Better IDE type-inference performance (less work for the language server to extract types from code)

_source: https://vuejs.org/api/sfc-script-setup.html_

Note that defineProps and defineEmits are compiler macros only usable inside <script setup>. **They do not need to be imported,** and are compiled away when <script setup> is processed ,which is why they are not added the the `vueImports` array.


samijaber (commented):
This is huge! Thanks @raymondmuller. Will take a look at this very soon and give feedback :)

packages/core/src/generators/vue.ts
```diff
@@ -772,6 +926,7 @@ const componentToVue =
           parser: 'vue',
           plugins: [
             // To support running in browsers
+            require('prettier/parser-typescript'),
```
samijaber, line 929: Would you mind adding this in a separate PR first? It caused a huge (highly positive!) change in the Vue snapshot tests, but it makes it hard to see if the rest of this PR introduced changes to the Options API output

raymondmuller: Here you go: https://github.com/BuilderIO/mitosis/pull/679

samijaber: Sweet! Merged it. Whenever you can, merge `main` into this branch 🙏🏽

raymondmuller: Done

packages/core/src/generators/vue.ts
```diff
@@ -752,7 +770,143 @@ const componentToVue =
...
+        )})`;
+      } else if (component.propsTypeRef && component.propsTypeRef !== 'any') {
+        str += `defineProps<${component.propsTypeRef}>()`;
+      } else {
+        str += `defineProps(${json5.stringify(Array.from(elementProps))})`;
+      }
+      return str;
+    };
+
+    function generateCompositionApiScript() {
```
samijaber, line 791: Would it be possible to move `generateCompositionApiScript` and `generateOptionsApiScript` outside of the scope of the main generator function? Explicitly providing props to them will make the code a lot less tangled together 🙏🏽

raymondmuller: Done (https://github.com/BuilderIO/mitosis/pull/665/commits/83b2e9eb7182bedbaa05badd5305551d73e88e57)

samijaber (changes requested):
I see from the snapshots that function generation is broken. 

One good way to test this is to tweak our e2e tests here: https://github.com/BuilderIO/mitosis/blob/2d760644b9b91b22486a9c3c6349403d14f8d993/e2e/e2e-app/mitosis.config.js#L48-L49

Such that Vue 3 uses this new composition API (or Vue 2, doesn't matter!) Then the e2e test will make sure the code is valid, compiles, and can actually pass e2e playwright tests

samijaber (commented):
@raymondmuller This looks great! Thanks for the updates.

Looks like e2e tests are still failing for the vue3 Composition API code though.

raymondmuller:
> @raymondmuller This looks great! Thanks for the updates.
> 
> Looks like e2e tests are still failing for the vue3 Composition API code though.

Passing now!

packages/core/src/generators/vue.ts
```diff
@@ -752,8 +722,246 @@ const componentToVue =
...
+  } else {
+    str += `defineProps(${json5.stringify(Array.from(props))})`;
+  }
+  return str;
+};
+
+function appendValueToRefUses(input: string, refKeys: string[], assignmentOnly = false) {
+  let output = input;
+  refKeys.forEach((ref) => {
+    let regexpStr = `((this\\.)?${ref})\\b`;
```
samijaber, line 746: We want to avoid using regexes for manipulating properties, as they're hard to maintain and have caused us lots of issues in the past. It's better to use an AST like babel's to transform the appropriate nodes.

We have 2 helpers to handle this type of stuff:

- `replaceIdentifiers` for simple use-cases: https://github.com/BuilderIO/mitosis/blob/2d760644b9b91b22486a9c3c6349403d14f8d993/packages/core/src/helpers/replace-identifiers.ts#L4

- `babelTransformExpression` for more complex ones (used internally by `replaceIdentifiers`): https://github.com/BuilderIO/mitosis/blob/2d760644b9b91b22486a9c3c6349403d14f8d993/packages/core/src/helpers/babel-transform.ts#L87


https://astexplorer.net/ is a useful tool to test out your transform and make sure it behaves as expected, or to learn what types of nodes you want to change. There are also a handful of examples in the codebase of using these helpers, that should give you an idea of how they use.

packages/core/src/generators/vue.ts
```diff
@@ -752,8 +729,243 @@ const componentToVue =
...
+};
+
+function appendValueToRefs(input: string, component: MitosisComponent, options: ToVueOptions) {
+  const refKeys = Object.keys(pickBy(component.state, (i) => i?.type === 'property'));
+
+  let output = processBinding(input, options, component, false);
+
+  return babelTransformExpression(output, {
+    Identifier(path: babel.NodePath<babel.types.Identifier>) {
+      console.log({ path });
```
samijaber, line 757:
```suggestion

```

samijaber (approved):
Absolutely fantastic contribution. Thank you @raymondmuller 🏆 🏆 🏆

Issue #529, opened 2022-06-30
Fiddle: add TS Support


samijaber:
Our Fiddle still does not support TS, even though Mitosis itself knows how to handle it!

<img width="843" alt="image" src="https://user-images.githubusercontent.com/1393142/176569542-29269529-7459-4501-9dc2-41611cf7b489.png">

This is weird because we set `language="typescript"`:

https://github.com/BuilderIO/mitosis/blob/7bc921c4ef53fdaf1b902618aab34f8398c1ad93/packages/fiddle/src/components/Fiddle.tsx#L800
https://github.com/BuilderIO/mitosis/blob/7bc921c4ef53fdaf1b902618aab34f8398c1ad93/packages/fiddle/src/components/CodeEditor.tsx#L146

Not sure where the bug is, but would be neat to get  type coverage in the fiddle!

samijaber:
Ideas from other issue: 

https://github.com/BuilderIO/mitosis/issues/183#issuecomment-1050129752

raymondmuller:
Passing `path="mitosis-file.tsx"` as a prop to <MonacoEditor might bring us one step closer.
Getting other errors now though
<img width="499" alt="Screenshot 2022-08-19 at 15 42 26" src="https://user-images.githubusercontent.com/3808818/185635160-d8dc518c-bf41-4b60-b3c7-5e606bbef501.png">

<img width="517" alt="Screenshot 2022-08-19 at 15 41 28" src="https://user-images.githubusercontent.com/3808818/185635302-8d6cfa76-738b-4ce2-a908-e9da3098de1d.png">

<img width="517" alt="Screenshot 2022-08-19 at 16 01 02" src="https://user-images.githubusercontent.com/3808818/185635575-652cbeb7-810b-4ca4-8d11-e70fa2940de1.png">

.


Missing types.

samijaber:
My thoughts:

- From my NextJS memories, we'd have to add the raw-loader config to nextjs's webpack config to extend it
and from there try to do the `raw-loader` import
- alternatively, looks like the string might be able to be a relative path to the node modules as well: https://stackoverflow.com/questions/43058191/how-to-use-addextralib-in-monaco-with-an-external-type-definition

- also since we have `@builder.io` imports 🤔, we might need to update to a new version of Monaco Editor if ours is before this fix was pushed https://github.com/microsoft/monaco-editor/issues/2295

raymondmuller:
That seems to work :)

[...]

PR #668, opened 2022-08-25
Fix/529 add TS support to fiddle

raymondmuller:
## Description

Adds Typescript support to fiddle (solves #529).

Screenshot 2022-08-25 at 23 19 08

Screenshot 2022-08-25 at 23 18 22

Screenshot 2022-08-25 at 23 19 22

Also replaces createMuiTheme with createTheme per the deprecation notice in the console and updates **next**.


raymondmuller:
Not sure why the build fails saying it can't find the types import @samijaber  . Works fine for me locally (running yarn ci:fiddle-build).

samijaber:
> Not sure why the build fails saying it can't find the types import @samijaber . Works fine for me locally (running yarn ci:fiddle-build).

I assume that's because the path to the node modules is different in the Vercel server itself. 🤔

Maybe we should look at https://webpack.js.org/guides/asset-modules/ instead: seems like a replacement for the raw-loader, and maybe that will work better 🤔

raymondmuller:
@samijaber 
JsxRuntimeTypes is found (@builder.io/mitosis/jsx-runtime) however the other types aren't. I've added index.d.ts to the exports in package.json, but that didn't seem to help either (https://github.com/BuilderIO/mitosis/pull/668/files#diff-0b810c38f3c138a3d5e44854edefd5eb966617ca84e62f06511f60acc40546c7R35). Is this cached and does the dependency (mitosis in this case) need to be updated? 

I see this in the GH Action (  ➤ YN0013: │ 2490 packages were already cached)

samijaber (approved):
🏆 🏆 🏆 🏆

PR #674, opened 2022-08-28
fix(svelte): output more natural prop defs

raymondmuller:
## Description

Outputs a more 'natural way' of Svelte code.
As described in https://github.com/BuilderIO/mitosis/pull/632 , you probably won't write the prop declarations in Svelte by
1. Defining the props type in a context module script
2. Then defining the type like this for example: Props['name']

You would more likely write `export let name: string;` instead

Examples:

Before:
Screenshot 2022-08-28 at 16 18 55

Screenshot 2022-08-28 at 16 19 25

After:

Screenshot 2022-08-28 at 16 26 52

Screenshot 2022-08-28 at 16 27 03

### DIFF
--- a/packages/core/src/helpers/prop-types-to-json.ts
+++ b/packages/core/src/helpers/prop-types-to-json.ts
@@ -1,4 +1,4 @@
-import { JSONObject } from 'src/types/json';
+import { JSONObject } from '../types/json';
 import { MitosisComponent } from '..';
 
 export const propTypesToJson = (
--- a/packages/core/src/helpers/prop-types-to-json.ts
+++ b/packages/core/src/helpers/prop-types-to-json.ts
@@ -1,0 +1,28 @@
+import { JSONObject } from 'src/types/json';
+import { MitosisComponent } from '..';
+
+export const propTypesToJson = (
+  props: string[], // collected props
+  json: MitosisComponent,
+  strip = true, // strip the prop types from json.types after collecting them
+): JSONObject => {
+  let obj: JSONObject = {};
+
+  if (json.types?.length && json.propsTypeRef) {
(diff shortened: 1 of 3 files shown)


packages/core/src/helpers/prop-types-to-json.ts
```diff
@@ -0,0 +1,28 @@
...
+        const matches = propTypes[0].match(regexp);
+        if (matches?.length) {
+          obj[prop] = matches[0];
+        }
+      });
+    }
+  }
+
+  return obj;
+};
```
samijaber, line 28: Hesitant to go down this path...Although it'll work nicely for simple examples like the one in our snapshot, won't it completely fall apart for more complex scenarios including type imports, extending other types/interfaces, or anything involving type arithmetic?

Take for example:

```tsx
import type SomeType from '../types/my-component-types.ts';

Interface Props extends SomeType {
  otherProps: boolean;
}

// OR
type Props = Pick<SomeType<'propName'>;
```

To be perfectly honest, I don't think Mitosis as a project is concerned with the output looking exactly like how a human would write them, so I'm not worried with keeping type references, i.e. `export let prop: Props['prop']: '';`. That looks fairly acceptable to me.

raymondmuller: It will indeed probably not work for those two examples. I am pretty new to Typescript so hadn't thought about that.

Cool, let's skip this!

samijaber: No problem.

Indeed, the only way to do some sort of type simplification is to run the `tsc` TypeScript Compiler on the entire Mitosis project, store the results somewhere, and then parse them to know what to output. A large undertaking

PR #686, opened 2022-09-01
Fix: TS imports

samijaber:
## Description

- fixes issue where JSX parser was dropping certain import declarations, by setting this field to `true` https://babeljs.io/docs/en/babel-preset-typescript#onlyremovetypeimports
- add snapshot tests
- remove redundant JSX snapshot tests


packages/core/src/parsers/jsx/jsx.ts
```diff
@@ -53,7 +53,19 @@ export function parseJsx(
...
+    presets: [
+      [
+        tsPreset,
+        {
+          isTSX: true,
+          allExtensions: true,
+          // If left to its default `false`, then this will strip away:
+          // - unused JS imports
+          // - types imports within regular JS import syntax
+          onlyRemoveTypeImports: true,
```
samijaber, line 65: Who doesn't love a single-line obscure config fix that takes forever to find

PR #695, opened 2022-09-04
feat: support configuring a custom parser and file extensions

raymondmuller:
## Description

This allows configuring a custom parser (a function returning MitosisJSON) and configuring a different file extension.

I am writing a custom parser, which is why I need this :)

### DIFF
--- a/packages/core/src/types/config.ts
+++ b/packages/core/src/types/config.ts
@@ -1,3 +1,5 @@
+import { MitosisComponent } from './mitosis-component';
+
 export type Format = 'esm' | 'cjs';
 export type Language = 'js' | 'ts';
 interface TranspilerOptions {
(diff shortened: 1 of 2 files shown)


packages/core/src/types/config.ts
```diff
@@ -51,4 +53,14 @@ export type MitosisConfig = {
...
+  /**
+   * Configure the extension of the files you want to compile
+   * Default value: 'lite.tsx'
+   */
+  extension: string;
+  /**
+   * Configure a custom parser function which takes a string and returns MitosisJSON
+   * Defaults to the JSXParser of this project (src/parsers/jsx)
+   */
+  parser: (code: string) => MitosisComponent;
```
samijaber, line 65: I would make these optional:

```suggestion
  extension?: string;
  /**
   * Configure a custom parser function which takes a string and returns MitosisJSON
   * Defaults to the JSXParser of this project (src/parsers/jsx)
   */
  parser?: (code: string) => MitosisComponent;
```

samijaber (approved):

Issue #688, opened 2022-09-02
project example basic build vue output with the keyword class


PengBoUESTC:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [X] Vue
- [ ] Web components

### Reproduction case

[...]

PR #698, opened 2022-09-05
feat: avoid using JavaScript keyword as property name

PengBoUESTC:
## Description
#688 
Add a short description of:
- throw an Error when js keywords used as props, 
-  js keywords should not be used as var

I get the `prohibitedKeywordRE` from [vue/core](https://github.com/vuejs/core/blob/fa6556a0d56eeff1fec4f948460351ccf8f99f35/packages/compiler-core/src/validateExpression.ts)


samijaber (approved):
Thanks @PengBoUESTC ! Appreciate the contribution :)

PR #702, opened 2022-09-05
fix: #700 vue3 composition api does not add computed

### LINKED ISSUE #700: Vue 3 composition-api  do not use computed

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [X] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKNroDRwQB2AqmACYCGM6cAvnAGZQQhwBEAAgEbLAA25dFAB0wCAHoQwHKmCp2AbgBQy9AA9IsOEMaVk%2FeI2REAxjHFE4AWQCeAYVaQi6IjAAUYFmFQBKRMpwcKbEqPBh1LQAvCgY2LjuCIFBcESUIOgAXBzY6ABu6Ox4yUEA5ujwaOgACt7u%2FkkpKVAVyFBWXhA%2BImkZcAA%2B%2FXAA5Lr6hgBqlPzI6MMl9Ml0vipByS0wbVbuCwA85MB5AHwLQbvARGDIMKcppqioUQiNTa8h%2FNDZ7C3kRbdBdDo%2FzgeWmsyeERoPXS6CBrxSxHsAAtKERyk93PlXDB%2FFEjog4JD0NC%2BjEsW4RDBKFByjARKCZrRAbcJCd4QAJdD8D63BBEkRVWpdOgAQjgAElgqi4FATHALnAAEroSjmAgTWYETAQfjAcgEaBwAAywAAjnxyCK9hIDsdkitlHQgA

### Expected Behaviour

when using `useStore` `get` should produce `computed` https://vuejs.org/api/reactivity-core.html#computed reactive variable

### Actual Behaviour

Not render at all `get` state

### Additional Information

_No response_

raymondmuller:
## Description
closes #700

### DIFF
--- a/packages/core/src/generators/vue.ts
+++ b/packages/core/src/generators/vue.ts
@@ -815,6 +815,8 @@ function generateCompositionApiScript(
     } `;
   }
 
+  const getterKeys = Object.keys(pickBy(component.state, (i) => i?.type === 'getter'));
+
   let str = dedent`
     ${props.size ? getCompositionPropDefinition({ component, props, options }) : ''}
     ${refs}
@@ -851,12 +853,29 @@ function generateCompositionApiScript(
           )}})`
     }
     ${
+      !getterKeys
+        ? ''
+        : getterKeys
+            .map((key) => {
+              const code = component.state[key]?.code?.toString();
+              return !code
+                ? ''
+                : `const ${key} = computed(${appendValueToRefs(
+                    code.replace(key, '').replace('get ()', '() =>'),
+                    component,
+                    options,
+                  )})`;
+            })
+            .join('\n')
+    }
+    ${
       !onUpdateWithoutDeps?.length
         ? ''
         : onUpdateWithoutDeps.map((hook) => {
             return `onUpdated(() => ${appendValueToRefs(hook.code, component, options)})`;
           })
     }
+
     ${
       !onUpdateWithDeps?.length
         ? ''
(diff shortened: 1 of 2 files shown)


packages/core/src/generators/vue.ts
```diff
@@ -929,6 +950,7 @@ const componentToVue: TranspilerGenerator<ToVueOptions> =
       component.hooks.onMount?.code && vueImports.push('onMounted');
       component.hooks.onUnMount?.code && vueImports.push('onUnMounted');
       onUpdateWithoutDeps.length && vueImports.push('onUpdated');
+      size(getterKeys) && vueImports.push('computed');
```
samijaber, line 953: Shouldn't this be a vue3 import only? i.e. wrapped in a `if (options.vueVersion >= 3)` check?

I think that might be the case for some other imports as well like `watch` and `ref`.

raymondmuller: Composition API is available in Vue 2 (since 2.7) as well and this is wrapped in a `if (options.api === 'composition)`

samijaber:
Just need to update snapshots after we merged your other PR @raymondmuller

samijaber (approved):

PR #724, opened 2022-09-12
fix(vue2): Show

raymondmuller:
## Description

Fixes Vue 2 Show / v-if.
Right now when the content is pure text, it doesn't work

**Current*
https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN0A0cmAFhAO5wC%2BcAZlBCHAEQACARssADYAm6UAHTAIAehDAYEVMFTMA3AChF6AB6RYcfrRzJu8WsgB2AYxgijcALIBPAMKNIR9EZgAKMAzCoAlIkVwcCYQRqjwANqoZOREGDCkFAC6cAC8KBjYeOhuMFDI6D5KAXBQ6DDIUJZuxYG1cAA8vMAAbgB8NXW19QmU5CQuKQhRFFTtneO1%2FdzcEB3j9aI9YxMLTW3FhYpUQA%3D%3D%3D

**After**
Screenshot 2022-09-12 at 22 19 21

Also, some content doesn't render currently

**Current**
Screenshot 2022-09-12 at 22 33 37

**After**
Screenshot 2022-09-12 at 22 33 40


packages/core/src/generators/vue.ts
```diff
@@ -125,7 +125,10 @@ type BlockRenderer = (json: MitosisNode, options: ToVueOptions, scope?: Scope) =
 const NODE_MAPPERS: {
   [key: string]: BlockRenderer | undefined;
 } = {
-  Fragment(json, options) {
+  Fragment(json, options, scope) {
+    if (options.vueVersion === 2 && scope?.isRootNode) {
+      return `<div>\n${json.children.map((item) => blockToVue(item, options)).join('\n')}\n</div>`;
+    }
```
samijaber, line 131: that's a bit of a hack unfortunately...because it means that we will add an extraneous `div` compared to other generators' code.

I would rather we throw and alert the user that Vue 2 cannot handle fragments at the root node. Then the user has a choice:

- manually add a `div` (that way all outputs will be consistent)
- do not generate code for Vue 2

raymondmuller: Makes sense. Will make the change tomorrow 👍

samijaber:
Made some tweaks to avoid the need to bypass the multi-root-node error you added in our tests.

Issue #732, opened 2022-09-15
Mitosis strips out boolean attribute unless you pass in value


chadalen:
### I am interested in helping provide a fix!

Sure

### Which generators are impacted?

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=IYOw5grgNsBOQ%3D%3D%3D&code=FAUwHgDg9gTgLgAgCYgGYEMCuAbRrMB2AxnAJZQEICyAngMJQC20BIBcAFBDFBAM4BKBAG9gCBDBBxMMShzHiEAHgB8CxctoNmFNolJ866GEhVEKcPQgAWISUoD0Wpiz1qN4pc52t2CA0YmALzCcDCYIAC%2BZhZWtvZO9C667O6KjmkCANzAkcBAA

### Expected Behaviour

This mitosis code
```jsx
<MyComponent isCard>content here</MyComponent>
```

Should keep the `isCard` prop in any output
e.g React
```jsx
<MyComponent isCard>content here</MyComponent>
```

### Actual Behaviour

This mitosis code
```jsx
<MyComponent isCard>content here</MyComponent>
```

Strips out the `isCard` prop in all outputs

e.g react output
```jsx
<MyComponent>content here</MyComponent>
```

### Additional Information

To get around this problem we have to set the value of the prop
e.g this mitosis code works
```jsx
<MyComponent isCard={true}>content here</MyComponent>
```

react output
```jsx
<MyComponent isCard={true}>content here</MyComponent>
```

However coming from React world it's more common to do this

```jsx
<MyComponent isCard>content here</MyComponent>
```

_No response_

PR #733, opened 2022-09-15
fix(jsx): boolean attribute

PatrickJS:
## Description
fix https://github.com/BuilderIO/mitosis/issues/732


packages/core/src/parsers/jsx/element-parser.ts
```diff
@@ -190,6 +190,12 @@ export const jsxElementToJson = (
       if (types.isJSXAttribute(item)) {
         const key = item.name.name as string;
         const value = item.value;
+
+        // boolean attribute
+        if (!item.hasOwnProperty('value')) {
```
samijaber, line 195: I think we can reuse `value` from above
```suggestion
        if (value === null) {
```

samijaber (commented):
Probably need to `yarn snapupdate`

samijaber (approved):
Can we add one of these props to a snapshot to make sure it works

PatrickJS:
I'm adding a snapshot because the preview deployment didn't show it fixed

PatrickJS:
snaps show it works and even in the preview now

Screen Shot 2022-09-15 at 10 47 54 AM

PR #735, opened 2022-09-19
fix: replace file extension based on config

raymondmuller:
## Description

Replace file extension with target's extension based on the config.

### DIFF
--- a/packages/cli/src/build/build.ts
+++ b/packages/cli/src/build/build.ts
@@ -250,8 +250,10 @@ const replaceFileExtensionForTarget = ({
   target: Target;
   path: string;
   options: MitosisConfig;
-}) =>
-  path.replace(/\.lite\.tsx$/, getFileExtensionForTarget({ type: 'filename', target, options }));
+}) => {
+  let regex = new RegExp(`.${options.extension}$`);
+  return path.replace(regex, getFileExtensionForTarget({ type: 'filename', target, options }));
+};
 
 /**
  * Transpiles and outputs Mitosis component files.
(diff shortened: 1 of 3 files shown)


packages/cli/src/build/build.ts
```diff
@@ -250,8 +250,10 @@ const replaceFileExtensionForTarget = ({
   target: Target;
   path: string;
   options: MitosisConfig;
-}) =>
-  path.replace(/\.lite\.tsx$/, getFileExtensionForTarget({ type: 'filename', target, options }));
+}) => {
+  let regex = new RegExp(`.${options.extension}$`);
```
samijaber, line 254: `options.extension` does not have a default value, so if not set the regex will be `.$` here.

We should:

- make it required here (remove `?`): https://github.com/BuilderIO/mitosis/blob/a2ce09c470ee6aac7634c62282320cb19333682e/packages/core/src/types/config.ts#L60

- Add its default value here as `lite.tsx`: https://github.com/BuilderIO/mitosis/blob/a2ce09c470ee6aac7634c62282320cb19333682e/packages/cli/src/build/build.ts#L35-L41

packages/cli/src/build/helpers/transpile.ts
```diff
@@ -20,6 +20,10 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
       // afterwards, we replace all `.lite` imports with the correct file extension
       /\.lite(['"][;\)])/g,
       `${getFileExtensionForTarget({ type: 'import', target, options })}$1`,
+    )
+    .replace(
+      `.${options.extension}`,
```
samijaber, line 25: Heads up: you will need to make this change in other places as well, including some logic in `core`:
image'

Although I am debating moving the `render-imports` that each generator uses into the `cli`, so that it runs on component & non-component files (which we realized later on needs to happen anyways).

samijaber (approved):

PR #741, opened 2022-09-20
Feat/add ts js toggle to fiddle

raymondmuller:
## Description

- Adds a toggle to the fiddle where you can select Typescript or Javascript output (whenever its added to the array returned by the function `hasBothTsAndJsSupport `
- default to typescript for Svelte and Vue
Screenshot 2022-09-20 at 22 17 58
Screenshot 2022-09-20 at 22 18 06


packages/core/src/generators/svelte.ts
```diff
@@ -311,6 +311,7 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
       stateType: 'variables',
       prettier: true,
       plugins: [FUNCTION_HACK_PLUGIN, ...plugins],
+      typescript: true,
```
samijaber, line 314: I would rather keep typescript support OFF by default for all generators because outputting typescript requires additional build tools after Mitosis.

packages/core/src/generators/vue.ts
```diff
@@ -536,6 +536,7 @@ const BASE_OPTIONS: ToVueOptions = {
   plugins: [],
   vueVersion: 2,
   api: 'options',
+  typescript: true,
```
samijaber, line 539: same here
```suggestion
```

packages/core/src/generators/svelte.ts
samijaber:
```suggestion
```

PR #742, opened 2022-09-20
fix: remove slots from prop definitions in vue

raymondmuller:
## Description
Remove slots from the prop definition in Vue

Before:
Screenshot 2022-09-20 at 22 32 25

After:
Screenshot 2022-09-20 at 22 32 11

### DIFF
--- a/packages/core/src/generators/vue.ts
+++ b/packages/core/src/generators/vue.ts
@@ -579,7 +579,7 @@ function generateOptionsApiScript(
   options: ToVueOptions,
   path: string | undefined,
   template: string,
-  props: Set<string>,
+  props: string[],
   onUpdateWithDeps: extendedHook[],
   onUpdateWithoutDeps: extendedHook[],
 ) {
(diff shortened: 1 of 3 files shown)


packages/core/src/generators/vue.ts
```diff
@@ -966,7 +966,7 @@ const componentToVue: TranspilerGenerator<ToVueOptions> =
 
     const getterKeys = Object.keys(pickBy(component.state, (i) => i?.type === 'getter'));
 
-    const elementProps = getProps(component);
+    const elementProps = Array.from(getProps(component)).filter((prop) => !prop.startsWith('slot'));
```
samijaber, line 969: you can reuse `packages/core/src/helpers/slots.ts`'s `isSlotProperty` here

packages/core/src/generators/vue.ts
```diff
@@ -966,7 +966,7 @@ const componentToVue: TranspilerGenerator<ToVueOptions> =
 
     const getterKeys = Object.keys(pickBy(component.state, (i) => i?.type === 'getter'));
 
-    const elementProps = getProps(component);
+    const elementProps = Array.from(getProps(component)).filter(isSlotProperty);
```
samijaber, line 969: You want to _exclude_ those that are slots though, so you need to negate this filter!

raymondmuller: Yeah, noticed that right after :)

samijaber (approved):

PR #743, opened 2022-09-20
fix(svelte): slots

raymondmuller:
## Description

Improves slots in Svelte

Before:
Screenshot 2022-09-20 at 23 07 46
Screenshot 2022-09-20 at 23 09 05

After:
Screenshot 2022-09-20 at 23 07 33
Screenshot 2022-09-20 at 23 08 59

### DIFF
--- a/packages/core/src/generators/svelte.ts
+++ b/packages/core/src/generators/svelte.ts
@@ -30,6 +30,7 @@ import { gettersToFunctions } from '../helpers/getters-to-functions';
 import { babelTransformCode } from '../helpers/babel-transform';
 import { pipe } from 'fp-ts/lib/function';
 import { hasContext } from './helpers/context';
+import { isSlotProperty, stripSlotPrefix } from '../helpers/slots';
 import { VALID_HTML_TAGS } from '../constants/html_tags';
 import { isUpperCase } from '../helpers/is-upper-case';
 import json5 from 'json5';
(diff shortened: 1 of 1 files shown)


packages/core/src/generators/svelte.ts
```diff
@@ -372,7 +397,7 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
 
     const hasData = dataString.length > 4;
 
-    const props = Array.from(getProps(json));
+    const props = Array.from(getProps(json)).filter((prop) => !prop.startsWith('slot'));
```
samijaber, line 400: same advice here of using `isSlotProperty`

packages/core/src/generators/svelte.ts
```diff
@@ -95,6 +97,23 @@ ${json.children.map((item) => blockToSvelte({ json: item, options, parentCompone
...
 {/if}`;
   },
+  Slot({ json, options, parentComponent }) {
+    if (!json.bindings.name) {
+      const key = Object.keys(json.bindings).find(Boolean);
+      if (!key) return '<slot />';
+
+      return `
+        <span #${key}>
+        ${json.bindings[key]?.code}
```
samijaber, line 107: You might want to wrap this with `stripStateAndPropsRefs` as well, in case you are relying on state or props within this slots code 🤔

samijaber (approved):

PR #748, opened 2022-09-22
Feature: inject vue & angular config

### LINKED ISSUE #726: Plugins do not receive options

No

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

Can not

### Expected Behaviour

When you write custom plugins, judging from API of the plugin function, you should get options, but you do not, because all of the generators are just ignoring the third parameter.

Also, the whole point of having plugins that have some kind of order of execution, you should have a result of the previous execution.

My case is that I want to add `styleUrls` to the angular component decorator, but I can not know exactly what component file contents I am working on because I do not have access to the previous `json.post` result and in the `code.pre` I have to parse the content to understand what was the JSON for it.
I hope you get the idea 

### Actual Behaviour

I do not get options in plugins

### Additional Information

_No response_

samijaber:
## Description

- Closes #726
- adds Vue & Angular config overriders via metadata hook or plugins
- fix overrides path logic in cli'
-

### DIFF
--- a/packages/core/src/generators/angular.ts
+++ b/packages/core/src/generators/angular.ts
@@ -373,6 +373,9 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       ],`
           : ''
       }
+      ${Object.entries(json.meta.angularConfig || {})
+        .map(([k, v]) => `${k}: ${v}`)
+        .join(',')}
     })
     export default class ${json.name} {
       ${localExportVars.join('\n')}
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/mitosis-cli",
-  "version": "0.0.22",
+  "version": "0.0.23",
   "description": "mitosis CLI",
   "types": "build/types/types.d.ts",
   "bin": {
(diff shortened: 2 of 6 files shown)


packages/core/src/generators/angular.ts
```diff
@@ -373,6 +373,9 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       ],`
           : ''
       }
+      ${Object.entries(json.meta.angularConfig || {})
+        .map(([k, v]) => `${k}: ${v}`)
+        .join(',')}
```
g-cheishvili, line 378: This is not safe. Users will encounter strange issues. Imagine case:
user added to `meta.angularConfig` property `selector`, then they would face duplication of that property. Instead what should be done is that the configs, which are already in there hardcoded, moved to some object, then spread and unified with `meta.angularConfig` and then done 
```ts
thatUnifiedObjectWhereAllComponentMetadataInformationIs.map(([k, v]) => `${k}: ${v}`)
        .join(',')
```
Without that, user will be forced to look into the generator code, in order to understand what went wrong and why

PR #750, opened 2022-09-22
refactor(angular): angular component metadata generation

g-cheishvili:
## Description

Add a short description of:
- Instead of having values in string moved them to Object
- Now it is in Object and can be more or less better overridden by the user
- Also, touched a bit `imports` generation

### DIFF
--- a/packages/core/src/generators/angular.ts
+++ b/packages/core/src/generators/angular.ts
@@ -341,7 +341,7 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       template: `\`
         ${indent(template, 8).replace(/`/g, '\\`').replace(/\$\{/g, '\\${')}
         \``,
-      ...(css.length ? { styles: `\`${indent(css, 8)}\`` } : {}),
+      ...(css.length ? { styles: `[\`${indent(css, 8)}\`]` } : {}),
       ...(options.standalone
         ? // TODO: also add child component imports here as well
           {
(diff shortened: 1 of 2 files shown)


samijaber (approved):
Thanks for this!

samijaber:
you'll need to run `fmt:prettier` at the repo root to fix your PR

g-cheishvili:
Yeah, noticed. I am off from PC now, its 2:30 am here, will do tomorrow

packages/core/src/generators/angular.ts
```diff
@@ -335,7 +335,25 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
         }),
     });
-
+    // Preparing built in component metadata parameters
+    const componentMetadata: Record<string, any> = {
+      selector: `'${kebabCase(json.name || 'my-component')}'`,
+      template: `\`
+        ${indent(template, 8).replace(/`/g, '\\`').replace(/\$\{/g, '\\${')}
+        \``,
+      ...(css.length ? { styles: `\`${indent(css, 8)}\`` } : {}),
```
samijaber, line 344: You removed the `[]` brackets in this string, this breaks the rendering of the styles property. you can see that the snapshot tests are all shifted.

```suggestion
      ...(css.length ? { styles: `[\`${indent(css, 8)}\`]` } : {}),
```

Make sure to run `yarn snapupdate` in `packages/core` locally to update test snapshots and see if there are changes there

samijaber (changes requested):

samijaber (approved):
Awesome, welcome to Mitosis! ⚡

PR #751, opened 2022-09-23
refactor(angular): refactored angular related imports generation

g-cheishvili:
## Description

- added `populatedWithAngularImports` and moved adding Angular related imports to be BEFORE `json.post` call.
- added one helper for adding an import to `MitosisComponent`. I do not know if something like it existed, could not find it, so added it. if there is any, please refer me to it.

## Reasoning
I added these because otherwise in `json.post` you can not see which imports are gonna end up in file completely. Also, output code's structure is now better because it follows a widely accepted pattern:
- imports
- types
- anything else
- class declaration

### DIFF
--- a/packages/core/src/helpers/add-import.ts
+++ b/packages/core/src/helpers/add-import.ts
@@ -1,0 +1,26 @@
+import { MitosisComponent, MitosisImport } from '@builder.io/mitosis';
+
+/**
+ * Add an import to a component, if it doesn't already exist.
+ * Returns modified MitosisComponent, which is NOT a new object
+ * @param json The component to add the import to
+ * @param theImport The import to add
+ */
+export function addImportToMitosisComponent(
+  json: MitosisComponent,
+  theImport: MitosisImport,
+): MitosisComponent {
+  if (!json.imports) {
+    json.imports = [];
+  }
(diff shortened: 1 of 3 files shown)


packages/core/src/helpers/add-import.ts
```diff
@@ -0,0 +1,26 @@
+import { MitosisComponent, MitosisImport } from '@builder.io/mitosis';
+
+/**
+ * Add an import to a component, if it doesn't already exist.
+ * Returns modified MitosisComponent, which is NOT a new object
+ * @param json The component to add the import to
+ * @param theImport The import to add
+ */
+export function addImportToMitosisComponent(
```
samijaber, line 9: We don't have this yet. Thanks for putting it together!

samijaber (approved):

packages/core/src/helpers/add-import.ts
g-cheishvili: Aand I need to update snapshots here as well

g-cheishvili:
@samijaber fixed everything. Take a look

samijaber:
@g-cheishvili I believe this is still pending updating snapshots?

PatrickJS:
@g-cheishvili any updates on this?

PR #752, opened 2022-09-23
fix svelte mitosis_styling fn

samijaber:
## Description

- fix svelte style injector to handle null/undefined prop

### DIFF
--- a/packages/core/src/generators/svelte.ts
+++ b/packages/core/src/generators/svelte.ts
@@ -463,7 +463,7 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
         hasStyle(json)
           ? `
         function mitosis_styling (node, vars) {
-          Object.entries(vars).forEach(([ p, v ]) => { node.style[p] = v })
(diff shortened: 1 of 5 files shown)


packages/core/src/generators/svelte.ts
```diff
@@ -463,7 +463,7 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
         hasStyle(json)
           ? `
         function mitosis_styling (node, vars) {
-          Object.entries(vars).forEach(([ p, v ]) => { node.style[p] = v })
+          Object.entries(vars || {}).forEach(([ p, v ]) => { node.style[p] = v })
```
Dkendal, line 466:
```suggestion
          Object.entries(vars ?? {}).forEach(([ p, v ]) => { node.style[p] = v })
```

samijaber: hmmm, would rather keep the `||` operator since it works the same way for this particular scenario at least, and this code is injected into the final Svelte component of our users. So better off generating code that has 100% support over 92% if the latter doesn't give us an advantage.

mhevery (approved):

Issue #747, opened 2022-09-22
Angular external component selectors are not converted


ocombe:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=IYOw5grgNsBOQ%3D%3D%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFChIsRHACSAYwgA7IqXKUA5DRABPZeBwq9PYMpVbJE9AA8Z8NiRzIm8EshWKY3NQFk9AGEKSBV0FRgACjByMFQASkQJODgLVHgAbRUcEHQAGjgMGAA5HPQAXTgAXhQMbDx0SKpsdAA3dCp46xSodBhkKDVI5JS4AB4WYFaAPhHR8aVVHln5sZ5JmZGuiUIgA%3D

### Expected Behaviour

Mitosis should convert the components names into the correct angular selectors when importing from external modules

### Actual Behaviour

I import from an external mitosis module, but it doesn't convert the components into their correct angular selectors.

The mitosis code:
```tsx
export default function Icon(props: IconProps) {
  return (
    <i
      data-testid="icon"
  ></i>
  );
}
```
gets turned into this angular code
```ts
@Component({
    selector: "icon",
    template: `
    <i data-testid="icon"></i>
  `,
})
class Icon {
}
```

and the mitosis button code that imports Icon:
```tsx
export default function Button(props: ButtonProps) {
  return (
    <button
      data-testid="button"
    >
        <Icon />
    </button>
  );
}
```

is transformed into this angular code:
```
@Component({
    selector: "button",
    template: `
    <button
      data-testid="button"
    >
          <Icon></Icon>
    </button>
  `,
})
class Button {
}
```

Notice that in the last angular template it's using `Icon` instead of `icon` (lowercase) which is the correct selector.
Whereas when I'm importing from a local mitosis file, it does use the selectors correctly (probably thanks to https://github.com/BuilderIO/mitosis/pull/357)

### Additional Information

[...]

Issue #746, opened 2022-09-22
Angular generated code misses default imports


ocombe:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFChIsRHDYkcyJvByo4AIWQxBAOyKlylKhx4MdEXeIkSAxpdQQm6DkwgBzABRaLugJQ26AAeMvAKSiqkyLq2MNz6ALIAngDCFJC66LownmDkYKh%2BiBJwcPa6qPAA2ro4IOgANHAYMABydegAunAAvCgY2HjonlTY6ABu6FR%2BkiVwUOgwyFD6nnOlADw%2BegB866VwABLoTG4AhHAAkmU4%2BlDRcMD6AEroOLFNAGrIjXCYTsAWE1oHAADLAACOjBYZ32Gx420se1KMwkhCAA

### Expected Behaviour

Import default from another file should work with the Angular generator

### Actual Behaviour

Whenever I add a default import from another file, it generates the code just fine in react but not in Angular. The following imports will be removed in Angular:
```ts
import { default as Button } from "./button";
import Button from "./button";
import * as Button from "./button";
```

Whereas the following import works just fine and is present in the generated code:
```ts
import { Button } from "./button";
```

It seems to work on mitosis.builder.io but it doesn't work for me with the latest version of mitosis.
If I set "standalone" to true in the generator options then the import it correctly retained in the generated file.

### Additional Information

This is particularly annoying because Mitosis requires default exports, which means that if I want one mitosis file to import from another mitosis file it will not work and I will have to create a non default export in an intermediate file (that imports default from the first and re-exports it as a named export).

ocombe:
After looking at the code, it seems that mitosis does this on purpose: https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L352
[...]

PR #754, opened 2022-09-23
Angular/746 747

alujs:
#746 
Preserved imports by following https://github.com/BuilderIO/mitosis/issues/746

#747 
and added json.name to the selector for https://github.com/BuilderIO/mitosis/issues/747


packages/core/src/__tests__/angular.test.ts
```diff
@@ -11,3 +11,14 @@ describe('Angular', () => {
...
+describe('Angular with Preserve Imports', () => {
+  runTestsForTarget({ options: {}, target: 'angular', generator: componentToAngular });
+  runTestsForTarget({
+    options: {
+      preserveImports: true,
+    },
+    target: 'angular',
+    generator: componentToAngular,
+  });
+});
```
samijaber, line 24: Thanks for adding these! Moving this to a separate test file would be better so that the snapshots are in their own files. Otherwise the `angular.test.ts.snap` becomes super enormous.

packages/core/src/__tests__/angular.import.test.ts
```diff
@@ -0,0 +1,13 @@
+import { componentToAngular } from '../generators/angular';
+import { runTestsForTarget } from './shared';
+
+describe('Angular with Preserve Imports', () => {
+  runTestsForTarget({ options: {}, target: 'angular', generator: componentToAngular });
```
samijaber, line 5: This one is identical to the original tests in `angular.test.ts`. Can we remove them?

samijaber (approved):

Issue #757, opened 2022-09-23
core: proposal for useStyle hook


raymondmuller:
I think it would be great to be able to easily add extra CSS to a component and therefore have created a draft PR with a proof of concept of how this could look.

### Proof of concept (PR: https://github.com/BuilderIO/mitosis/pull/758)
A new hook has been created: useStyle, which can be used to add CSS as followed:

<code>useStyle(\` 
input:focus {
outline: 1px solid blue;
}
\`)</code>

Which would then be appended to where the css is defined by each generator.
The PR contains examples for Vue, React, Angular and Svelte of how that could look like.

It seems like an easy, but powerful addition to Mitosis as this allows you to add
- pseudo selectors (e.g. input:focus)
- nested selectors (e.g. .card > .card-title) 
- media queries 

and well, anything you would like to put in your CSS really.

This would be really helpful for Sveltosis too, as it would allow me to just copy whatever you would put in the <style> section, directly to that property.

Svelte
<img width="1375" alt="Screenshot 2022-09-23 at 23 01 27" src="https://user-images.githubusercontent.com/3808818/192057138-532f979d-dc00-4cb2-900e-4c1475fc296a.png">

Vue
<img width="1491" alt="Screenshot 2022-09-23 at 23 01 24" src="https://user-images.githubusercontent.com/3808818/192057146-a03f832b-226e-4669-b884-1e7c5d69b658.png">

Angular
<img width="1499" alt="Screenshot 2022-09-29 at 09 12 04" src="https://user-images.githubusercontent.com/3808818/192964505-28133466-8f93-49f1-b179-b5e1b852a329.png">

React
<img width="1342" alt="Screenshot 2022-09-23 at 23 00 27" src="https://user-images.githubusercontent.com/3808818/192057179-1961d81e-5308-4856-875f-0d5e83577ac5.png">

Opinions?

g-cheishvili:
> Opinions?

That's not angular on the screenshot 😄 
Other than that, would be nice to have

raymondmuller:
Accidentally uploaded the wrong screenshot :) (editing the original one as well)

Here you go
<img width="1499" alt="Screenshot 2022-09-29 at 09 12 04" src="https://user-images.githubusercontent.com/3808818/192964340-74a4c907-5278-44da-835f-9acdd2a81578.png">

samijaber:
[...]

PR #758, opened 2022-09-23
feat: useStyle poc

raymondmuller:
## Description
See feature idea discussion here: https://github.com/BuilderIO/mitosis/issues/757

### DIFF
--- a/packages/core/src/constants/hooks.ts
+++ b/packages/core/src/constants/hooks.ts
@@ -4,4 +4,5 @@ export const HOOKS = {
   CONTEXT: 'useContext',
   REF: 'useRef',
   DEFAULT_PROPS: 'useDefaultProps',
+  STYLE: 'useStyle',
 } as const;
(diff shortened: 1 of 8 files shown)


samijaber (commented):
I think you're better off making the change once inside of `collectCss`, the helper responsible for aggregating all of the styles into a CSS string. You could append `json.style` to its output, which ought to give us support in all generators that support CSS instantly. 

https://github.com/BuilderIO/mitosis/blob/45c48a5f5c322f3995eb7db8da28bb0f925d75fd/packages/core/src/helpers/styles/collect-css.ts#L80-L85

Mind giving that a shot and see if it works without hiccups?

packages/core/src/helpers/styles/helpers.ts
```diff
@@ -21,7 +21,7 @@ export const nodeHasStyle = (node: MitosisNode) => {
 };
 
 export const hasCss = (component: MitosisComponent) => {
-  let hasStyles = false;
+  let hasStyles = !!component.style?.length;
```
samijaber, line 24: might be worth returning early if it's already true, to avoid traversing the entire component:

```ts

```
```suggestion
  let hasStyles = !!component.style?.length;
  
  if (hasStyles) {
    return true
  }
```

samijaber (commented):
Love this! Can we add 2 component snapshot tests:

- 1 that uses only `useStyle`
- 1 that uses `useStyle` _and_ the `css` prop, showing how they get combined.

samijaber (approved):
Excellent @raymondmuller . Thank you for this new feature!! 🏆

PR #762, opened 2022-09-26
fix: add svg specific tags to VALID_HTML_TAGS

cmgriffing:
## Description

A library I am porting from React to Mitosis uses inlined SVG elements.
When converting the Mitosis component to Svelte, various tags are changed into something like this:

```
<svelte:component this={line} ...
```

This does not build. After digging through the source code, it seems that the `svelte:component` for the SVG tags is because they are not present in VALID_HTML_TAGS.

My PR currently is missing the "Filter Expression" tags such as:
- [feBlend](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feBlend)
- [feColorMatrix](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feColorMatrix)
- [feComponentTransfer](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feComponentTransfer)
- [feComposite](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feComposite)
- etc...

I can add them to this PR if needed.


cmgriffing:
Oh sorry. I should have looked for an existing issue and PR more. Thanks!

samijaber:
No worries! This is the first time 2 folks raise a PR to fix the same issue. Mitosis is gaining traction!! 😄

PR #771, opened 2022-09-27
allow useMetaData

### LINKED ISSUE #769: ESLint: `only-default-function-and-imports` does not allow `useMetadata` hook
Example of valid code marked as invalid:

```ts
// eslint-disable-next-line @builder.io/mitosis/only-default-function-and-imports
useMetadata({
  qwik: {
    component: {
      isLight: true,
    },
  },
});

export default function RenderComponent(props) {
  return (
<div>Text</div>
  );
}
```

sahilmob:
fixes #769

### DIFF
--- a/packages/eslint-plugin/src/rules/only-default-function-and-imports.ts
+++ b/packages/eslint-plugin/src/rules/only-default-function-and-imports.ts
@@ -1,6 +1,7 @@
 import { Rule } from 'eslint';
 import * as types from '@babel/types';
 import isMitosisPath from '../helpers/isMitosisPath';
+import { HOOKS } from '../constants/hooks';
 
 // ------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/only-default-function-and-imports.ts
```diff
@@ -51,7 +51,11 @@ const rule: Rule.RuleModule = {
             !(
               types.isExportNamedDeclaration(child) &&
               types.isTSTypeAliasDeclaration(child.declaration)
-            )
+            ) &&
+            (!types.isExpressionStatement(child) ||
+              !types.isCallExpression(child.expression) ||
+              !types.isIdentifier(child.expression.callee) ||
+              child.expression.callee.name !== 'useMetadata')
```
samijaber, line 58: I think we can/should add this hook to https://github.com/BuilderIO/mitosis/blob/a7e54bb7d38d671a2c52fb272fab9c933d684173/packages/eslint-plugin/src/constants/hooks.ts#L1

instead of hardcoding it in this file

sahilmob: Yes we should, but only `useMetadata` is allowed, right?

samijaber (approved):
Thank you!

Issue #782, opened 2022-09-28
Parser is not getting any options


g-cheishvili:
https://github.com/BuilderIO/mitosis/blob/ad117b1ef745bf470eb1b5a7b7f3120b64ba6073/packages/cli/src/commands/compile.ts#L121

Parser is not getting any options. There is no way currently to override them

PR #783, opened 2022-09-28
feat(cli): automatic ts output detection

g-cheishvili:
## Description

- In cli compile run, none of the options were passed to the parseJSX function.
- Previously you would have missing types in outputs, if your types were declared outside of the component code
- This is not ideal solution. It should be overridable by user through options. Someone, more well versed in source code should look at it to say from where that option should come

closes https://github.com/BuilderIO/mitosis/issues/782 by the way

### DIFF
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -118,7 +118,7 @@ const command: GluegunCommand = {
 
         switch (from_) {
           case 'mitosis':
-            json = parseJsx(data, { typescript: !!path.match(/\.tsx?$/) });
+            json = parseJsx(data, { typescript: generatorOpts.typescript });
             break;
 
           case 'builder':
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/compile.ts
```diff
@@ -118,7 +118,7 @@ const command: GluegunCommand = {
 
         switch (from_) {
           case 'mitosis':
-            json = parseJsx(data);
+            json = parseJsx(data, { typescript: !!path.match(/\.tsx?$/) });
```
samijaber, line 121: `generatorOpts` will hold the options for the generator you are currently using. You should be able to use its `generatorOpts.typescript` option and pass that through to `parseJsx`!

```suggestion
            json = parseJsx(data, { typescript: generatorOpts.typescript });
```

samijaber (approved):

samijaber:
Thank you for the contribution @g-cheishvili . Now you can enable typescript by adding:
```js
// mitosis.config.js
module.exports = {
  options: {
    react: { typescript: true }
  }
}
```

PR #787, opened 2022-09-30
fix(core): vue remove ref type in js

raymondmuller:
## Description

Fixes this: (see ref<undefined>)
Screenshot 2022-10-01 at 01 32 42


samijaber (approved):
Thanks!

PR #800, opened 2022-10-07
React Server Components (RSC) support

steve8708:


packages/core/src/generators/rsc.ts
```diff
@@ -0,0 +1,81 @@
+import traverse from 'traverse';
+import { Plugin } from '..';
+import { fastClone } from '../helpers/fast-clone';
+import { isMitosisNode } from '../helpers/is-mitosis-node';
+import { isUpperCase } from '../helpers/is-upper-case';
```
samijaber, line 5: seems like TSC is complaining about this line

samijaber:
```suggestion
```

packages/core/package.json
```diff
@@ -25,6 +25,7 @@
   "engines": {
     "npm": "99999999.9.9"
   },
+  "version": "0.0.72-5",
   "version": "0.0.72",
```
samijaber, line 29: probably wanna remove the bottom line here

PR #803, opened 2022-10-07
refactor(core): nicer vue methods in options API

raymondmuller:
## Description

Changes `onChange: function onChange() {` to `onChange()`, which is a way more natural way of writing Vue components.
It works the same.

### DIFF
--- a/packages/core/src/generators/vue/optionsApi.ts
+++ b/packages/core/src/generators/vue/optionsApi.ts
@@ -128,8 +128,10 @@ export function generateOptionsApiScript(
     data: false,
     getters: false,
     functions: true,
-    valueMapper: (code) => processBinding({ code, options, json: component }),
-  });
+    valueMapper: (code) => {
+      return processBinding({ code, options, json: component });
+    },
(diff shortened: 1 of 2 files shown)


packages/core/src/generators/vue/optionsApi.ts
```diff
@@ -128,8 +128,10 @@ export function generateOptionsApiScript(
     data: false,
     getters: false,
     functions: true,
-    valueMapper: (code) => processBinding({ code, options, json: component }),
-  });
+    valueMapper: (code) => {
+      return processBinding({ code, options, json: component });
+    },
+  }).replace(/: (.*?)\(\)/g, '()');
```
samijaber, line 134: I am very hesitant to add `String.replace` calls like this one. They are highly error-prone and can cause issues. We want to rely on the Babel AST generator/types instead for tasks like this.

In this case, the change might need to be somewhere in `valueMapper`, or in `convertStateMemberToString` which uses `valueMapper`. Logic is a bit hairy in there, which is part of the reason why we haven't been able to make this improvement yet.

PR #810, opened 2022-10-11
fix(core): append .value to refs when using Vue 3 composition API

ajanes93:
# Fix(core): append `.value` to refs when using Vue 3 composition API

## Description

Currently only usages of state keys are being appended with `.value` in the Vue 3 composition API generator.  This PR ensures that `.value` is also appended to usages of refs.

- Include Mitosis component ref keys when checking which nodes should have `.value` appended to them
- Added extra checks to ensure `.value` is not appended in watcher definitions, or in function declarations
- Updated snapshots


samijaber (approved):
Awesome! Thanks for the fix 😄

PR #813, opened 2022-10-11
mitosis-811 added prefer-show-over-ternary-operator rule

### LINKED ISSUE #811: ESLint Rule: no ternaries in JSX render body
We want to encourage folks to use the `<Show>` component instead of ternaries (`foo ? <bar /> : <baz />`). The latter doesn't always work across all generators, especially once they start getting nested inside of each other in complex ways.

sahilmob:
fixes #811

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -17,7 +17,8 @@
     "url": "https://github.com/BuilderIO/mitosis"
   },
   "scripts": {
-    "test": "jest --watch",
+    "test": "jest",
+    "test:watch": "jest --watch",
     "build": "tsc",
     "build:declaration": "tsc --project tsconfig.build.json",
     "start": "tsc -w",
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/BuilderIO/mitosis"
   },
   "scripts": {
-    "test": "jest",
+    "test": "jest --watch",
     "build": "tsc",
     "build:declaration": "tsc --project tsconfig.build.json",
     "start": "tsc -w",
--- a/packages/eslint-plugin/src/rules/__tests__/prefer-show-over-ternary-operator.ts
+++ b/packages/eslint-plugin/src/rules/__tests__/prefer-show-over-ternary-operator.ts
@@ -1,0 +1,61 @@
+import { RuleTester } from 'eslint';
+import rule from '../prefer-show-over-ternary-operator';
+
+const opts = {
+  filename: 'component.lite.tsx',
+  parserOptions: {
+    ecmaVersion: 2018, // Allows for the parsing of modern ECMAScript features
+    sourceType: 'module', // Allows for the use of imports
+    ecmaFeatures: {
+      jsx: true, // Allows for the parsing of JSX
+    },
+  },
+} as const;
+
+var ruleTester = new RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
+});
+
+ruleTester.run('only-default-function-and-imports', rule, {
+  valid: [
+    {
+      ...opts,
+      code: `
+      export default function MyComponent(props) {
+        return (
+          <div>
+            <Show when={foo}>
+              <bar />
+            </Show>
+            <Show when={!foo}>
+              <baz />
+            </Show>
+          </div>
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin/src/rules/prefer-show-over-ternary-operator.ts
```diff
@@ -0,0 +1,46 @@
...
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+    //
+    const listener: Rule.RuleListener = {
+      ConditionalExpression(node) {
+        context.report({
+          node,
+          message:
+            '<Show> is preferred over ternary expression as its guaranteed to work across all generators',
```
samijaber, line 38:
```suggestion
            'Ternary expression support is minimal. Please use the Mitosis `<Show>` component instead.',
```

packages/eslint-plugin/src/rules/__tests__/prefer-show-over-ternary-operator.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+    {
+      ...opts,
+      code: `
+      export default function MyComponent(props) {
+        return <div>{foo ? <bar /> : <baz />}</div>;
+      }
+    `,
+      filename: 'file.jsx',
+    },
```
samijaber, line 46: I'm surprised this block here is not causing the PR to fail, since this should be `invalid` code 🤔

packages/eslint-plugin/src/rules/prefer-show-over-ternary-operator.ts
```diff
@@ -0,0 +1,46 @@
...
+    // ----------------------------------------------------------------------
+
+    // any helper functions should go here or else delete this section
+
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+    //
+    const listener: Rule.RuleListener = {
+      ConditionalExpression(node) {
```
samijaber, line 34: We actually want to allow ternary expression if they're: 
- outside of the render body
- inside of the render body, but not dictating the render logic

Here are some more examples:

```ts
//valid
  const state = useState({ 
    getName() {
      props.a ? 'a' : 'b'
    } 
  })
  return <div />;
}
```


```ts
//valid
  return <div> <input value={props.a ? 'a' : 'b'} /> </div>;
}
```

So I think we'll want to improve the check here such that it makes sure the ternary is:
- within a `JSXElement` node
- not within a `JSXAttribute` node (unless its key is exactly `children`)


this could help explore the differences https://astexplorer.net/#/gist/17d9685f8230c74cee49060e91a30d47/latest

packages/eslint-plugin/src/rules/__tests__/prefer-show-over-ternary-operator.ts
sahilmob: since the filename property is set, all eslint rules don't apply because it is not a mitosis file.

samijaber: Ohhh, right! I see. I missed that, thanks :)

sahilmob:
@samijaber I improved the checks such that it doesn't report an error if the ternary expression's `parent.parent` is `JSXAttribute` or its parent is `ExpressionStatement`. These checks will cover all the cases I've written initially, plus the two you've added in your comment.

packages/eslint-plugin/package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/BuilderIO/mitosis"
   },
   "scripts": {
-    "test": "jest",
+    "test": "jest --watch",
```
samijaber, line 20: `test` is used by our CI. If it has `--watch` I think this means our CI tests will never pass.

```suggestion
    "test": "jest",
    "test:watch": "jest --watch",
```

samijaber:
@sahilmob Sorry for the delay, forgot to follow up on this PR.

It's good to go! Nne useful thing would be to add it to the recommended config here: https://github.com/BuilderIO/mitosis/blob/main/packages/eslint-plugin/src/configs/recommended.ts

Probably as a `warning` instead of a hard `error`. I'll merge this PR though and we can do that in a followup PR 👍🏽

samijaber (approved):

Issue #839, opened 2022-10-25
mitosis build  clean options.dest all matching directories


kingzez:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

no link

### Expected Behaviour

```js
module.exports = {
  files: 'components/**',
  dest: 'packages',
  targets: [ 'vue3', 'react' ],
  options: {},
  getTargetPath: ({ target }) => `${target}/src`,
}
```
output is `${dest}/${target}/src/${files}`, eg: packages/vue3/src/components/*

```
├── components
│   ├── button
│   │   ├── button.lite.tsx
├── packages
│   ├── react
│   │   ├── src
│   │   │   └── components
│   ├── vue3
│   │   ├── src
│   │   │   └── components
│   ├── site
│   │   ├── src
│   │   │   └── components
```

and my project struct is here，when run mitosis build，will delete packages/site/components/*,  because glob match the 
pattern:  `${options.dest}/**/*/${options.files}`

> https://github.com/BuilderIO/mitosis/blob/main/packages/cli/src/build/build.ts#L71
```js
async function clean(options: MitosisConfig) {
  const files = await glob(`${options.dest}/**/*/${options.files}`);
  await Promise.all(
    files.map(async (file) => {
      await remove(file);
    }),
  );
}
```

so, it should clean one or all targets directory,  includes targets, should not clean others directory.

### Actual Behaviour

it clean all match pattern directory

PR #840, opened 2022-10-25
fix(cli): build clean only targets directory files

kingzez:
## Description

Add a short description of:
- what changes you made,  https://github.com/BuilderIO/mitosis/issues/839
- why you made them, and 

### DIFF
--- a/packages/cli/src/build/build.ts
+++ b/packages/cli/src/build/build.ts
@@ -68,7 +68,8 @@ const getOptions = (config?: MitosisConfig): MitosisConfig => ({
 });
 
 async function clean(options: MitosisConfig) {
-  const files = await glob(`${options.dest}/**/*/${options.files}`);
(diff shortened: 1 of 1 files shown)


packages/cli/src/build/build.ts
```diff
@@ -68,7 +68,8 @@ const getOptions = (config?: MitosisConfig): MitosisConfig => ({
 });
 
 async function clean(options: MitosisConfig) {
-  const files = await glob(`${options.dest}/**/*/${options.files}`);
+  const patterns = options.targets.map((target) => `${options.dest}/${target}/*/${options.files}`);
```
samijaber, line 71: We should use `getTargetPath` instead of `${target}`:

```suggestion
  const patterns = options.targets.map((target) => `${options.dest}/${options.getTargetPath(target)}/${options.files}`);
```

samijaber (approved):

PR #844, opened 2022-10-25
fix: camelcase attributes for react output

chadalen:


alujs (approved):

PatrickJS (approved):

packages/core/src/__tests__/react.test.ts
```diff
@@ -22,4 +24,11 @@ describe('React', () => {
     })({ component });
     expect(output).toMatchSnapshot();
   });
+
+  test('verify attributes are transformed to camelCase', () => {
+    const component = parseJsx(attribute);
+    const output = componentToReact()({ component });
+    console.log(output);
+    expect(output).toMatchSnapshot();
+  });
```
samijaber, line 33: Please add this test to this object instead, so its tested by every generator: https://github.com/BuilderIO/mitosis/blob/07fab0c4ed332fde2fa9b56e10eb9f4816c237d1/packages/core/src/__tests__/shared.ts#L96

then run `yarn snapupdate`

samijaber (changes requested):

packages/core/src/__tests__/react.test.ts
chadalen: Done

samijaber (approved):

PR #845, opened 2022-10-25
feat(angular): NgModules Generator

alujs:
Add ngModule generation to Angular generator.


packages/core/src/generators/angular.ts
```diff
@@ -76,6 +80,29 @@ const mappers: {
...
+  content: string,
+  name: string,
+  componentsUsed: string[],
+  component: MitosisComponent,
+  bootstrapMapper: Function | null | undefined,
+): string => {
+  return `import { NgModule } from "@angular/core";
+import { BrowserModule } from "@angular/platform-browser";
+
+${content.replace('export default', 'export')}
```
samijaber, line 93: You removed the `default` already, so I think this is not needed, right?

```suggestion
${content}
```

samijaber (approved):

PR #846, opened 2022-10-25
feat: Alpine.js generator

sbrow:
## Description

This PR adds a new `alpine` generator to `@builder.io/mitosis` (core). It makes no changes to the existing codebase except for adding the `alpine` generator to the fiddle and to the cli.

## Status: Work In Progress

This PR Is by no means production ready. However, as there is currently limited documentation on the subject of wrtining custom generators, by opening this PR I hope to get feedback from the Builder.io team on exactly what needs to happen in order for this to be merged; or, how to integrate the generator as a plugin of some sort if the Builder.io team is not interested in maintaining Alpine.js support.

### To-dos
- [ ] Tests: How should we test?
  - Is there a standard suite of components we should render w/ Alpine.js?
- [x] Fiddle integration
- [ ] CLI Integration
- [ ] directives
  - [x] [x-data](https://alpinejs.dev/directives/data): Currently supports inline (`x-data={open: false}`) and separate: 
 ```html
<script>
  document.addEventListener('alpine:init', () => {
    Alpine.data('dropdown', () => ({open: false}))
  })
</script>
```
  - [ ] [x-init](https://alpinejs.dev/directives/init)
  - [ ] [x-show](https://alpinejs.dev/directives/show): Should be implemented the same as `v-show` in Vue.
  - [ ] [x-bind](https://alpinejs.dev/directives/bind)
  - [x] [x-on](https://alpinejs.dev/directives/on)
    - [ ] Modifiers?
  - [ ] [x-text](https://alpinejs.dev/directives/text)
  - [ ] [x-html](https://alpinejs.dev/directives/html)
  - [ ] [x-model](https://alpinejs.dev/directives/model)
  - [ ] [x-modelable](https://alpinejs.dev/directives/modelable)
  - [x] [x-for](https://alpinejs.dev/directives/for)
  - [ ] [x-transition](https://alpinejs.dev/directives/transition)
  - [ ] [x-effect](https://alpinejs.dev/directives/effect)
  - [ ] [x-ignore](https://alpinejs.dev/directives/ignore)
  - [x] [x-ref](https://alpinejs.dev/directives/ref)
    - [ ] Should refnames be altered? i.e. `const inputRef = useRef();` => `this.$refs.input` ?
  - [ ] [x-cloak](https://alpinejs.dev/directives/cloak)
  - [ ] [x-teleport](https://alpinejs.dev/directives/teleport)
  - [x] [x-if](https://alpinejs.dev/directives/if)
  - [ ] [x-id](https://alpinejs.dev/directives/id)

### DIFF
--- a/packages/core/src/generators/alpine/generate.ts
+++ b/packages/core/src/generators/alpine/generate.ts
@@ -167,7 +167,8 @@ const blockToAlpine = (json: MitosisNode, options: ToAlpineOptions = {}): string
       } else if (key === 'ref') {
         str += ` x-ref="${useValue}"`;
       } else if (isValidAlpineBinding(useValue)) {
-        str += ` :${key}="${useValue}" `;
+        const bind = options.useShorthandSyntax ? ':' : 'x-bind:'
+        str += ` ${bind}${key}="${useValue}" `;
       }
     }
     if (selfClosingTags.has(json.name)) {
--- a/packages/core/src/generators/alpine/generate.ts
+++ b/packages/core/src/generators/alpine/generate.ts
@@ -88,7 +88,7 @@ const bindEventHandlerValue = compose(
   stripStateAndPropsRefs
 );
 
-const bindEventHandler = ({useShorthandSyntax}: ToAlpineOptions) => (eventName: string, code: string) => {
+const bindEventHandler = ({ useShorthandSyntax }: ToAlpineOptions) => (eventName: string, code: string) => {
   const bind = useShorthandSyntax ? '@' : 'x-on:'
   return ` ${bind}${bindEventHandlerKey(eventName)}="${bindEventHandlerValue(code).trim()}"`;
 };
@@ -166,16 +166,9 @@ const blockToAlpine = (json: MitosisNode, options: ToAlpineOptions = {}): string
       str += ` ${bind}${key}="${useValue}" `;
     }
   }
-  if (selfClosingTags.has(json.name)) {
-    return str + ' />';
-  }
-  str += '>';
-  if (json.children) {
-    str += json.children.map((item) => blockToAlpine(item, options)).join('\n');
-  }
-
-  str += `</${json.name}>`;
-  return str;
+  return selfClosingTags.has(json.name)
+    ? `${str} />`
+    : `${str}>${(json.children ?? []).map((item) => blockToAlpine(item, options)).join('\n')}</${json.name}>`;
 };
 
 
@@ -197,10 +190,10 @@ export const componentToAlpine: TranspilerGenerator<ToAlpineOptions> =
         ? stateObjectString
         : `${camelCase(json.name)}()`;
 
-      let str = json.children.map((item) => blockToAlpine(item, options)).join('\n');
-      if (css.trim().length) {
-        str += `<style>${css}</style>`;
-      }
+      let str = css.trim().length
+        ? `<style>${css}</style>`
+        : '';
+      str += json.children.map((item) => blockToAlpine(item, options)).join('\n');
 
       if (!options.inlineState) {
         str += `<script>
--- a/packages/core/src/generators/alpine/generate.ts
+++ b/packages/core/src/generators/alpine/generate.ts
@@ -16,6 +16,8 @@ import { BaseTranspilerOptions, TranspilerGenerator } from '../../types/transpil
 import { dashCase } from '../../helpers/dash-case';
 import { removeSurroundingBlock } from '../../helpers/remove-surrounding-block';
 import { camelCase } from 'lodash';
+import { getRefs } from '../../helpers/get-refs';
+import { MitosisComponent } from 'src/types/mitosis-component';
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by PatrickJS** [comment]
nice

**Comment by sbrow** [comment]
Would love some feedback from the team on how to handle the remaining items :)

**Comment by samijaber** [comment]
Hey @sbrow! I went through a long-distance move across Canada last week, so the past few days have been very busy for me.

I am catching up Mitosis PRs this week and will make sure to get you feedback on this ASAP.

This is absolutely phenomenal work btw! Thank you for taking the initiative of adding an Alpine.js generator 🙏🏽

**Comment by sbrow** [comment]
> Hey @sbrow! I went through a long-distance move across Canada last week, so the past few days have been very busy for me.

Ah, there I was thinking I was getting snubbed 🤣 In that case I'm impressed you've been as responsive as you have been. Hope everything arrives on time and undamaged!

Love Mitosis and I'm eager to help it flourish.

**Comment by samijaber** [comment]
@sbrow Amazing. 

Thanks for adding the snapshot tests too, that's very helpful.

One thing that would be _great_ to have from the get-go is an integration test, as we currently do for the majority of our web frameworks: https://github.com/BuilderIO/mitosis/tree/main/e2e

The architecture might be a bit complex (we have some improvements to make there). The easiest thing to do is copy another e2e webapp (`e2e-svelte` is a good candidate) and modify it to work with alpine. It needs to have:

- a `build` and a `serve` command
- be added here https://github.com/BuilderIO/mitosis/blob/d49b28a58321a11595c0deb312aee89b89cf9e6a/e2e/e2e-app/e2e.ts#L18-L30

If you need any other help with this, please reach out and we can pair on it.

**Comment by sbrow** [comment]
> The architecture might be a bit complex (we have some improvements to make there). The easiest thing to do is copy another e2e webapp (e2e-svelte is a good candidate) and modify it to work with alpine.

Roger. That might take some time, because Alpine components can't just be imported as javascript like every other framework. I'll have to use a templating language of some kind- typically I use Blade (php), but that clearly won't work here. I'm thinking I'll try ejs, since that seems to be the classic, battle-tested solution, but if there's another language already being used in the e2e tests, let me know and I'll use that instead.

This is uncharted territory for me, so we'll see how it goes 🤞

**Comment by samijaber** [comment]
@sbrow Ah, that does sound complicated. I don't mind leaving this for a secondary PR, and let you merge this one in for now so it doesn't get held up for too long!

**Comment by sbrow** [comment]
@samijaber Fortunately, I was able to get it working using `mustache`. When I run the playwright test by itself, it works (most of the time*), but when I run it together with the other generators, I get issues. I think this is an upstream problem and nothing I introduced, but I'll have to investigate later :)

\* I think sometimes the test looks for elements before Alpine has finished loading.

File: packages/core/src/__tests__/alpine.test.ts [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+describe('Alpine.js', () => {
+  const possibleOptions: ToAlpineOptions[] = [
+    {},
+    // { inlineState: true },
+    // { useShorthandSyntax: true },
+    // { inlineState: true, useShorthandSyntax: true },
+  ];
+  possibleOptions.map((options) =>
+    runTestsForTarget({ options, target: 'alpine', generator: componentToAlpine }),
+  );
```
samijaber, line 13: Ah, I missed one piece of info. You'll want to add an entry here for `alpine`: https://github.com/BuilderIO/mitosis/blob/a762948e6014450d5b8b80c961b771df4568f847/packages/core/src/__tests__/shared.ts#L228-L239

and add all of the tests. Only comment out tests if they cause the generator to crash, otherwise the more the merrier

sbrow: Can do!

packages/core/src/__tests__/__snapshots__/alpine.test.ts.snap
```diff
@@ -1,17 +1,4199 @@
 // Vitest Snapshot v1
 
-exports[`Alpine.js > Remove Internal mitosis package 1`] = `
+exports[`Alpine.js > Javascript Test > AdvancedRef 1`] = `
```
samijaber, line 3: I know this file is huge, but generating tons of snapshots is actually the easiest way for us to, at a glance, catch glaring errors. You don't need to fix anything right away, but this is probably the best place for you to get a feel for what you've missed, or edge-cases that are still unaccounted for.

sbrow: I see some issues, but so far only ones that I already had marked as a problem, mostly relating to child components and props. 🙂

samijaber (approved):
This is massive. Thank you for taking the time and effort to put this together @sbrow .

I'm going to merge this so that you don't trail behind `main` for too long. Feel free to send bugfixes and improvements as you encounter them!

samijaber:
Welcome to Mitosis! 😎 🏆

sbrow:
Feel free to mention me on GH or discord with any Alpine.js related issues!

PR #855, opened 2022-10-28
Change useClientEffect for useMount on qwik

CKGrafico:
## Description

Not sure why we are using a method that is not compatible with SSR but I think the correct one is https://qwik.builder.io/tutorial/hooks/use-mount/

### DIFF
--- a/packages/core/src/generators/qwik/component-generator.ts
+++ b/packages/core/src/generators/qwik/component-generator.ts
@@ -151,11 +151,9 @@ function emitTagNameHack(file: File, component: MitosisComponent, metadataValue:
 
 function emitUseMount(file: File, component: MitosisComponent) {
   if (component.hooks.onMount) {
-    // This is called useMount, but in practice it is used as
-    // useClientEffect. Not sure if this is correct, but for now.
     const code = component.hooks.onMount.code;
     file.src.emit(
-      file.import(file.qwikModule, 'useClientEffect$').localName,
+      file.import(file.qwikModule, 'useMount').localName,
       '(()=>{',
       code,
       '});',
--- a/packages/core/src/generators/qwik/component-generator.ts
+++ b/packages/core/src/generators/qwik/component-generator.ts
@@ -153,7 +153,7 @@ function emitUseMount(file: File, component: MitosisComponent) {
   if (component.hooks.onMount) {
     const code = component.hooks.onMount.code;
     file.src.emit(
-      file.import(file.qwikModule, 'useMount').localName,
+      file.import(file.qwikModule, 'useMount$').localName,
       '(()=>{',
       code,
       '});',


packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -151,11 +151,9 @@ function emitTagNameHack(file: File, component: MitosisComponent, metadataValue:
 
 function emitUseMount(file: File, component: MitosisComponent) {
   if (component.hooks.onMount) {
-    // This is called useMount, but in practice it is used as
-    // useClientEffect. Not sure if this is correct, but for now.
```
samijaber, line 155: From looking into this: `useMount$` will run on the client _and_ on the server, whereas all other equivalents in other frameworks:
- `mounted` in vue https://vuejs.org/api/options-lifecycle.html#mounted
- `useEffect(() => code, [])` in React
- `onMount` in svelte https://svelte.dev/tutorial/onmount

All of these do not run during SSR. so using `useClientEffect$` in qwik would be more in line with the rendered code of other frameworks for the `onMount` mitosis hook.


cc @mhevery in case I got this wrong

packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -151,11 +151,9 @@ function emitTagNameHack(file: File, component: MitosisComponent, metadataValue:
...
     const code = component.hooks.onMount.code;
     file.src.emit(
-      file.import(file.qwikModule, 'useClientEffect$').localName,
+      file.import(file.qwikModule, 'useMount').localName,
```
samijaber, line 156:
```suggestion
      file.import(file.qwikModule, 'useMount$').localName,
```

mhevery:
@CKGrafico What is your motivation behind this PR? Could you explain what problem you are trying to fix?

@samijaber This is very confusing. You are correct that current use, the `useMount()` method in mitosis is really `useEffect()` method. Could we change the API in Mitosis to better align the intent. 

To me `useMount()` means run when the component is created, wether on server or client. and `useEffect()` is only run on the client.

samijaber:
OK, sounds to me like we should be able to use `useMount$` for Qwik, as it ultimately will only run once per component instance. The fact that this single time might be on the server side is not necessarily an issue.

@CKGrafico You're going to have to run `yarn snapupdate` in `packages/core` to update the qwik snapshots tests

mhevery:
But there are cases in the SDK where we want to run it on the client, and as of right now we don't have such API in the Mitosis. So I think we need to add `useEffect` like API to the mitosis. Because committing the change as is will break the Qwik SDK.

samijaber:
> So I think we need to add `useEffect` like API to the mitosis

Hmm, that might be problematic. We already represent every possible way of using React's `useEffect` into a Mitosis hook already:

- `useEffect(() => myCode)` (no dep array): `onUpdate` without a dep array. Re-run on every re-render.
- `useEffect(() => myCode, [deps])` (dep array): `onUpdate` + deps array. Runs whenever deps update.
- `useEffect(() => myCode, [])` (empty dep array): `onMount`. Runs once on component instance mount.

The confusion here is whether we should tell Qwik to run Mitosis' `onMount` only on the client (by keeping `useClientEffec$`, or potentially on the server (by switching to `useMount$`). Adding

 an entire `useEffect` hook only to add this flexibility for 1 generator's output is overkill IMO. A better way might be to allow users to provide an options prop to `onMount` that qwik can consume to decide what to do:

```ts
onMount(() => {
  // my code
}, { clientSideOnly: boolean })
```

And then in qwik generator:

``` ts
const hookToUse = component.hooks.onMount.clientSideOnly ? 'useClientEffect$' : 'useMount$';
```

samijaber:
Had a chat with @mhevery . I think the best way forward right now is to:

- keep Mitosis `onMount` mapped to Qwik `useClientEffect$`
- map Mitosis `onInit` to Qwik `useMount$`. Currently, Mitosis' `onInit` is not handled by Qwik generator at all.

CKGrafico:
Understood @samijaber then is better to abandon this and you implement the other approach, right?

samijaber:
Yeah @CKGrafico, I think we'll need a separate PR. Made an issue to track this here: https://github.com/BuilderIO/mitosis/issues/861

Issue #858, opened 2022-10-30
Build: transformImports RegExp can't replace without semicolon


kingzez:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://github.com/BuilderIO/mitosis/blob/main/packages/cli/src/build/helpers/transpile.ts#L10

### Expected Behaviour

![image](https://user-images.githubusercontent.com/10891613/198893053-d800109b-a02b-41b4-bc10-18546b5a5b70.png)

<img width="1244" alt="image" src="https://user-images.githubusercontent.com/10891613/198893080-fd85140b-74f1-4dcc-ab89-7883f280a921.png">

The quote and semicolon is not necessary, it can't replace without semicolon, 
eg: `import Button from './button.lite'; ` can be replace to `import Button from './button.vue`
eg: `import Button from './button.lite' ` can't be replace to `import Button from './button.vue`

`import Button from './button.lite'; `  and  `import Button from './button.lite' ` both should be replace to `import Button from './button.vue`

> https://regex101.com/

### Actual Behaviour

`import Button from './button.lite' ` can't replace to `import Button from './button.vue' `

### Additional Information

https://github.com/BuilderIO/mitosis/blob/084de1af31f1181772fd73dcbfd20382e1508d47/packages/cli/src/build/helpers/transpile.ts#L10-L27

kingzez:
close https://github.com/BuilderIO/mitosis/pull/859

PR #859, opened 2022-10-30
fix(cli): transformImports better RegExp for replace extension

kingzez:
## Description

Add a short description of:
- what changes you made, #858

### DIFF
--- a/packages/cli/src/build/helpers/transpile.ts
+++ b/packages/cli/src/build/helpers/transpile.ts
@@ -18,7 +18,7 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
     )
     .replace(
       // afterwards, we replace all `.lite` imports with the correct file extension
-      /\.lite/g,
+      /\.lite['"]/g,
       `${getFileExtensionForTarget({ type: 'import', target, options })}$1`,
     )
     .replace(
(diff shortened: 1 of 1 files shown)


packages/cli/src/build/helpers/transpile.ts
```diff
@@ -13,12 +13,12 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
       // we start by replacing all `context.lite` imports with `context`
       // This Context replace is only needed for non-mitosis components, i.e. plain `.js`/`.ts` files.
       // Mitosis components have logic that transform context import paths correctly.
-      /\.context\.lite(['"][;\)])/g,
+      /\.context\.lite/g,
```
samijaber, line 16: We still need to make sure it's the end of a string, i.e. ends with `'` or `"`, otherwise we might replace every instance of `context.lite` in the code, which could be non-imports too

```suggestion
      /\.context\.lite['"]/g,
```

packages/cli/src/build/helpers/transpile.ts
```diff
@@ -13,12 +13,12 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
...
       `.context.js$1`,
     )
     .replace(
       // afterwards, we replace all `.lite` imports with the correct file extension
-      /\.lite(['"][;\)])/g,
+      /\.lite/g,
```
samijaber, line 21:
```suggestion
      /\.lite['"]/g,
```

samijaber (changes requested):

samijaber (approved):

Issue #818, opened 2022-10-13
Allow components to output multiple files per target


sbrow:
## What?

Certain frameworks, (such as Angular)  allow you to split your components into multiple files. It would be useful if mitosis could allow components to compile into more than one output file per target.

## Why?

Although this would be a helpful feature for teams that prefer multi-file components over single-file components, there are other instances, such as Alpine.js where it can be down-right necessary. When you write an Alpine component that requires importing code, that code needs to rolled into to the js bundle. However, if the component's DOM is rendered by a static templating engine, (which is a common use for Alpine), then you won't be able to put the js and the html in the same file.

## How?

The `useMetadata` hook seems like the ideal place to implement this feature. I think most components would be split by html/js/css, but not all. For example, Marko also supports [split components](https://markojs.com/docs/class-components/#split-components) where brower-only code resides in one file, and shared code goes in the other. I'm not sure about other frameworks, but I imagine that trying to add settings for each of them could get quite complicated, so I leave the implementation details up to the fine minds at BuilderIO 😄 

## Conclusion

I don't think this is a high priority issue, but as I mentioned in issue #11 I do think it's a necessary feature to make certain frameworks (i.e. Alpine.js) a viable target for Mitosis.

I love mitosis and I'm excited to see it grow!

PatrickJS:
this is interesting because Mitosis preprocesses the component but it's up to the framework and other tools to optimize steps after. For example qwik, angular, even typescript is processing the files after. I would say the goal for mitosis isn't for an opinionated optimized output just a good enough step for devs to decide those. This will allow anyones pipeline to further optimize in other steps since each framework has their own opinion on what that means. If this is the only way alpine works then it's the only way. HTML output for example can use this too since it should be separate files

PatrickJS:
I think this is fine until we find another/better approach

sbrow:
> If this is the only way alpine works then it's the only way. HTML output for example can use this too since it should be separate files

I can't tell whether you're for or against this idea, so let me elaboarate a little bit. Alpine.js components typically get rendered in raw html, without any pre-processing. 

```html
<!-- index.html -->
<html>
<head>
    <script defer src="https://unpkg.com/alpinejs@3.x.x/dist/cdn.min.js"></script>
</head>
<body>
  <style>
    .input {
      color: red;
    }
  </style>
  <div x-data="myComponent()">
    <input
      class="input"
      x-bind:value="name"
      x-on:change="name = $event.target.value"
    />

[...]

Issue #808, opened 2022-10-10
Elm + Webcomponents output


gampleman:
Hi this looks like a super useful project!

I would be interested in working on an output target aimed at integrating in Elm. This is somewhat challenging as Elm has a fairly radically different model than most of the existing targets for mitosis.

The ideal outcome is that most components could be compiled to pure Elm. To do this, there needs to be a reasonably good transpiration story from Typescript to Elm. [I've started prototyping this and it seems like it could work for a decent amount of relatively simple components](https://observablehq.com/@gampleman/typescript-subset-elm). 

The fallback would then be to generate 2 files: a web component using the existing web component output + an Elm "interface" file that would reference and wrap that component. I would foresee this happening in 3 scenarios:

1. Basically whenever `ref` is used, as Elm has no such concept.
2. When the JS/TS code uses concepts that aren't well suited to compiling into Elm:
     - heavy mutation
     - DOM APIs
     - importing random library code
3. When the typescript type information isn't good enough to allow sufficient analysis to transpile into Elm. This could alternatively throw an error. I suspect this wouldn't happen if compiling on the strictest TS settings (like forbidding `any`), but not sure how that's setup in Mitosis projects...

---

So, are there any docs for contributing code emitters? Also, looking at the JSON output, the code seems to get transformed into an imperative form. Is it possible to emit full on typescript AST with type information included?

samijaber:
Would love more folks contributing generators! :)

> So, are there any docs for contributing code emitters?

Yes! https://github.com/BuilderIO/mitosis/blob/main/developer/generators.md

Also, @raymondmuller worked on our Vue Composition API generator, and @sbrow is working on an Alpine generator here https://github.com/BuilderIO/mitosis/pull/846. Feel free to reach out in the discord to me/them/anyone with questions

I'm not familiar enough with Elm and/or Elm+webcomponents to know about the challenges you outlined. What I can note:

1- Mitosis currently only does a 1:1 mapping of Mitosis component file to output file. So it's going to require a non-trivial refactor across all generators to allow generators to output many files for each Mitosis component file.

2- We have been improving our preservation of types in the code, but we do not currently store a full TS AST. The premise of Mitosis is to parse a given syntax into the Mitosis Component JSON schema, acting as an intermediate representation between parsers and generators. So I don't know if we'll ever store a full AST like that.
[...]

PR #865, opened 2022-11-02
feat: Generators can now export more than one output file.

sbrow:
BREAKING CHANGE: Generators no longer return a string.

# Description

This PR makes the changes discussed in #808, and (when completed) will also close issue #818. The basic premise is to allow generators to output to more than one file, allowing Mitosis to output Multi-file components as well as Single-file components.

# To-Dos
- [ ] core
  - [x] Change the output of existing generators from string to array
  - [ ] Enumerate the types of outputs a component should have. e.g.
    - file extension based: `html`, `css`, `js`, `jsx`, `scss`, etc.
    - 'type' based: `component`, `dom`, `styles`, `state`...
- [ ] cli
  - [ ] save more than one file
- [ ] fiddle
  - [ ] Display multiple files

### DIFF
--- a/packages/core/src/types/transpiler.ts
+++ b/packages/core/src/types/transpiler.ts
@@ -6,7 +6,14 @@ export interface TranspilerArgs {
   component: MitosisComponent;
 }
 
-export type Transpiler<R = string> = (args: TranspilerArgs) => R;
+export type GeneratorOutput<R = string> = {
+  // content of output. Currently either a component string or a builder component JSON.
+  content: R;
+  // in the future, we will add more types like 'styles' for CSS Modules, etc.
+  type: 'css' | 'html' | 'js' | 'jsx' | 'component' | 'error';
+};
+
+export type Transpiler<R = string> = (args: TranspilerArgs) => GeneratorOutput<R>[];
 
 /**
  * This type guarantees that all code generators receive the same base options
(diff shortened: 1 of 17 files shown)


packages/core/src/types/transpiler.ts
```diff
@@ -6,7 +6,14 @@ export interface TranspilerArgs {
   component: MitosisComponent;
 }
 
-export type Transpiler<R = string> = (args: TranspilerArgs) => R;
+export type GeneratorOutput<R = string> = {
+  // content of output. Currently either a component string or a builder component JSON.
+  content: R;
+  // in the future, we will add more types like 'styles' for CSS Modules, etc.
+  type: 'css' | 'html' | 'js' | 'jsx' | 'component' | 'error';
```
samijaber, line 13: I would prefer to start off with only the `type`s that we currently generate, and add incrementally as we see fit. Otherwise folks will assume that the other types are already being generated in certain cases, which might cause confusion.

sbrow: Totally makes sense. The qwik generator sometimes returns an error. Should this be counted as `'component'`, or should I keep the `'error'` type?

samijaber: We should just throw these in the generator for now

sbrow:
> Excuse me, is there any progress on this currently?

Not that I know of. Still waiting on input from sami to proceed.

PR #870, opened 2022-11-03
feat: add onInit hook support in Qwik generator

### LINKED ISSUE #861: add `onInit` hook support in Qwik generator
Came out of this PR discussion: https://github.com/BuilderIO/mitosis/pull/855#issuecomment-1297496490

the Qwik generator needs to support the `onInit` mitosis hook, and its contents should be rendered within Qwik's `useMount$` hook.

How:

Essentially copy this logic: https://github.com/BuilderIO/mitosis/blob/084de1af31f1181772fd73dcbfd20382e1508d47/packages/core/src/generators/qwik/component-generator.ts#L151-L165

but replace `component.hooks.onMount.code` with `component.hooks.onInit.code`, and `useClientEffect$` with `useMount$`. That should be all!

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fix #861


packages/core/src/generators/qwik/component-generator.ts
```diff
@@ -162,6 +163,16 @@ function emitUseMount(file: File, component: MitosisComponent) {
     );
   }
 }
+
```
decadef20, line 166: I'm not sure should we rename `emitUseMount` to `emitUseClientEffect`  on line 153?

samijaber: yes, that's a good call. Let's do that 👍🏽

samijaber (approved):

PR #877, opened 2022-11-04
feat: Speed up builds with turborepo.

sbrow:
## Description

This PR attempts to demonstrate the benefits of using [turborepo](https://turbo.build/repo/docs), chiefly:

- Build steps are only run if their dependencies change.
- Running a task (i.e. `build`) for one package will automatically build its dependencies. No more forgetting to run `yarn build:core` before running the fiddle.
- Build tasks can be cached, which can dramatically speed up a stateful CI pipeline.

This works great for `@builder.io/mitosis` and `@builder.io/mitosis-fiddle`, but falls down with the e2e tests. I've tried to set it up to only run the `01-one-component` tests, but it seems like some of them might be broken?

Because the e2e tests rebuild their packages with each `case` folder, a radical change to the setup of the e2e tests would be required to make this work properly.

## How to use

```bash
git clone -b feature/turborepo git@github.com:sbrow/mitosis.git
cd mitosis
yarn && yarn turbo run start --filter='*fiddle'
// run it again to see the effects of caching
yarn && yarn turbo run start --filter='*fiddle'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ e2e/**/output
 packages/e2e-*/**
 **/test-results/
 **/playwright-report/
+
+.turbo
+**/.turbo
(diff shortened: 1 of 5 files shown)


sbrow:
Deprecated in favor of NX.

PR #881, opened 2022-11-05
Feature/fiddle add lint

### LINKED ISSUE #768: Fiddle: add ESLint plugin rules
It would be super cool if the Mitosis Fiddle applied our ESLint rules to the code editor. We already use eslint here:

https://github.com/BuilderIO/mitosis/blob/a7e54bb7d38d671a2c52fb272fab9c933d684173/packages/fiddle/src/components/CodeEditor.tsx#L18

We'd need to import `@builder.io/eslint-plugin-mitosis` and add it to the list of rules here

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fix #768


packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -132,15 +129,11 @@ export function CodeEditor(props: MonacoEditorProps) {
       monaco.editor.setModelMarkers(model, 'eslint', result.markers);
     },
     2000,
-    [props.value],
+    [props.value, editor],
```
decadef20, line 132: When `onMount` is triggered,  we should rerun here.

packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -12,15 +10,14 @@ import MonacoEditor, { EditorProps as MonacoEditorProps, useMonaco } from '@mona
...
-// TODO: add back when build fixed
-// linter.defineRules(rules as any);
 
 function eslint(code: string, version: any) {
   try {
     const markers = linter
       .verify(
         code,
         {
+          extends: ['plugin:@builder.io/eslint-plugin-mitosis/recommended'],
```
decadef20, line 20: I'm not sure whether it's right place to add.

decadef20:
image

Now it works.

packages/eslint-plugin/README.md
```diff
@@ -19,10 +19,10 @@ module.exports = {
       jsx: true,
     },
   },
-  plugins: ['@builder.io/mitosis'],
+  plugins: ['@builder.io/eslint-plugin-mitosis'],
   extends: [
     // Use this approach for our recommended rules configuration
-    'plugin:@builder.io/mitosis/recommended',
+    'plugin:@builder.io/eslint-plugin-mitosis/recommended',
```
samijaber, line 25: The original code is correct: https://eslint.org/docs/latest/user-guide/configuring/plugins#naming-convention

```
"@jquery/jquery", // means @jquery/eslint-plugin-jquery
```

packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -22,7 +20,21 @@ function eslint(code: string, version: any) {
...
+            'no-assign-props-to-state': 'error',
+            'use-state-var-declarator': 'error',
+            'no-async-methods-on-state': 'error',
+            'no-var-declaration-in-jsx': 'error',
+            'no-var-name-same-as-prop-name': 'error',
+            'jsx-callback-arrow-function': 'error',
+            'no-var-name-same-as-state-property': 'error',
+            'only-default-function-and-imports': 'error',
+            'no-conditional-logic-in-component-render': 'error',
+            'no-var-declaration-or-assignment-in-component': 'error',
```
samijaber, line 37: I think it would be better to actually use our `recommended` eslint rules, so that we don't have to remember to add each rule manually:

```
{
  extends: ['plugin:@builder.io/mitosis/recommended']
}
```

packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -1,8 +1,7 @@
 import { Linter as ESLinter } from 'eslint';
 import { useEffect, useState } from 'react';
 import { useDebounce } from 'react-use';
-// TODO: add back when build fixed
-// import { rules } from 'eslint-plugin-mitosis';
+import { rules } from '@builder.io/eslint-plugin-mitosis';
```
samijaber, line 4: The build is failing because you are not importing `@builder.io/eslint-plugin-mitosis` in the fiddle project. Need to add it to `package.json` `dependencies`

samijaber (changes requested):

packages/fiddle/src/components/CodeEditor.tsx
decadef20: I updated as your CR. But lint won't work if I don't specify rules.  Maybe there is something I must missed.

packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -12,18 +10,15 @@ import MonacoEditor, { EditorProps as MonacoEditorProps, useMonaco } from '@mona
...
   try {
     const markers = linter
       .verify(
         code,
         {
-          rules: {
-            'static-control-flow': 'error',
-          },
+          plugins: ['@builder.io/mitosis'],
+          extends: ['plugin:@builder.io/mitosis/recommended'],
```
samijaber, line 21: Unfortunately, this doesn't seem to work when I test it. Does it work for you if you write invalid code in the Fiddle?


I realize now that the linter needs to define the rules using the `linter.defineRules` function we removed: https://eslint.org/docs/latest/developer-guide/nodejs-api#linterdefinerules

So we'll need to bring this back:

```ts
import { rules } from '@builder.io/eslint-plugin-mitosis';
linter.defineRules(rules as any)
```

But that only defines the rules, it doesn't set them. It also does not set them with the `@builder.io/mitosis` prefix. 

So we'll need to add the recommended config separately, _but_ remember to remove the `@builder.io/mitosis` prefix:

```ts
import mitosisPlugin from '@builder.io/eslint-plugin-mitosis';

const recommendedRules: Record<string, string> = {}
Object.entries(mitosisPlugin.configs.recommended.rules).forEach(([key, value]) => {
  // remove `'@builder.io/mitosis/'` prefix from key using regex, and store it in trimmedKey
  const trimmedKey = key.replace(/^@builder.io\/mitosis\//, '');
  recommendedRules[trimmedKey] = value;
}

```

```ts
    const markers = linter
      .verify(
        code,
        {
          rules: recommendedRules
        }
```

I haven't tested this but I think it should work. Let me know 🙏🏽

samijaber (changes requested):

packages/fiddle/next.config.js
```diff
@@ -11,6 +11,7 @@ const nextConfig = {
       // https://webpack.js.org/configuration/resolve/#resolvealias
       fs: false,
     };
+    config.resolve.mainFields = ['browser', 'main', 'module'];
```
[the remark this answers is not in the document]
decadef20: image
Here is a error.

`esquery.parse is not a function Occurred while linting mitosis.lite.tsx`

I fixed it by [here](https://github.com/eslint/eslint/issues/15065#issuecomment-919890660)

packages/fiddle/src/components/CodeEditor.tsx
decadef20: It really confuses me. I have added it to `package.json`, the build is still failing.

samijaber:
@decadef20 By the way, If you want to test locally if the Fiddle build works, you can run `yarn run build` in `packages/fiddle` to see if you get the errors (to avoid having to commit the changes every time to test).

decadef20:
> @decadef20 By the way, If you want to test locally if the Fiddle build works, you can run `yarn run build` in `packages/fiddle` to see if you get the errors (to avoid having to commit the changes every time to test).

Totally works well on my machine

decadef20:
> Oh, I found it @decadef20. We need to build the eslint plugin for the Fiddle CI task:
> 
> https://github.com/BuilderIO/mitosis/blob/a762948e6014450d5b8b80c961b771df4568f847/.github/workflows/checks.yml#L52-L58
> 
> Right now we only build the `core` package as a pre-step. You'll want to add a `ci:build:eslint-plugin` npm script at the root and call it here before we build the fiddle

You saved my day. I'll try it again.

decadef20:
What's wrong with the deployment？ I visit Vercel pages, it occurs 404.  @samijaber

decadef20:
What's wrong with the deployment？ I visit Vercel pages, it occurs 404.  @samijaber

samijaber:
Ah, the new `build:eslint-plugin` command is missing from here too... https://github.com/BuilderIO/mitosis/blob/3ca7e8cb0f9b0b7d1134d2a0de692a0a0d306d03/packages/fiddle/package.json#L42

samijaber:
The rules work! amazing job, thank you for your patience with this @decadef20 
CleanShot 2022-11-10 at 08 29 00@2x

samijaber (approved):

Issue #882, opened 2022-11-07
Fiddle: Eslint cannot parse TS code


samijaber:
<img width="638" alt="CleanShot 2022-11-07 at 09 12 27@2x" src="https://user-images.githubusercontent.com/1393142/200319146-b03170d0-f408-4f0b-9ec5-ef6ebde303ac.png">

ESLint is throwing errors due to TS code. We need to configure the eslint instance in MonacoEditor to use the typescript-eslint parser https://typescript-eslint.io/docs/

https://github.com/BuilderIO/mitosis/blob/323af7949ec12902e3df095342a213d2c619fd49/packages/fiddle/src/components/CodeEditor.tsx#L20-L38

decadef20:
Once I load and use `@typescript-eslint/parser` in component file , it occurs error.

<img width="1086" alt="image" src="https://user-images.githubusercontent.com/7866718/201150782-21904a43-2fe2-46c2-aefa-fa9ee3686576.png">

here is [`constants.js`](https://splibanais.org/node_modules/%40nodelib/fs.scandir/out/constants.js)

<img width="1196" alt="image" src="https://user-images.githubusercontent.com/7866718/201151276-d779f0ed-a428-46d4-8c19-b6c277d95de9.png">

I guess it doesn't support browser environment.

samijaber:
Agh, that seems messy...

decadef20:
> I wonder if there's some way to hack around this, or if there's some other TS parser out there that can do the job in the browser. 😕

I'll try find a new way

decadef20:
[ts eslint playground](https://typescript-eslint.io/play/) is a good case to learn from. I'll figure it out later

PR #884, opened 2022-11-08
Feature/fiddle use ts eslint

### LINKED ISSUE #768: Fiddle: add ESLint plugin rules

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fix #768  and #882

### DIFF
--- a/packages/fiddle/src/components/CodeEditor.tsx
+++ b/packages/fiddle/src/components/CodeEditor.tsx
@@ -1,8 +1,7 @@
 import { Linter as ESLinter } from 'eslint';
 import { useEffect, useState } from 'react';
 import { useDebounce } from 'react-use';
-// TODO: add back when build fixed
-// import { rules } from 'eslint-plugin-mitosis';
+import { rules } from '@builder.io/eslint-plugin-mitosis';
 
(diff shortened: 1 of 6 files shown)


packages/fiddle/src/components/CodeEditor.tsx
```diff
@@ -1,8 +1,7 @@
 import { Linter as ESLinter } from 'eslint';
 import { useEffect, useState } from 'react';
 import { useDebounce } from 'react-use';
-// TODO: add back when build fixed
-// import { rules } from 'eslint-plugin-mitosis';
+import { rules, configs as lintConfigs } from '@builder.io/eslint-plugin-mitosis';
```
samijaber, line 4: I'm very confused why the Fiddle can't find this import, even if you add it to the `dependencies` in `package.json`....

PR #890, opened 2022-11-09
chore: add nx

FrozenPandaz:
## Description

This PR is the initial introduction of [Nx](https://nx.dev) as a build tool.

The first thing is altering the `yarn ci:build` command to an nx command that builds the same projects as before but has caching built in.

If you run `yarn ci:build` locally twice on this branch, you will see the 2nd run is completely cached.

After you run it (and also in CI), a link will be provided where you may view the logs separated out by project:
https://cloud.nx.app/runs/MHIJFR0ZK8

If Nx Cloud is setup (free for OSS repos like this), then those links will be posted to the PR via a Github comment.

The other thing you can do is run `nx graph --target @builder.io/mitosis` and you will see the below image.

image

You can explore this interactive graph further to see how this repo is structured.


FrozenPandaz:
The gitguardian comment is a false positive.

The Access Token in this PR is a READ-ONLY token which allows developers to pull results from the cache in Nx Cloud. But developers (people with access to this codebase) cannot write to the cache from their local machines without the secret READ-WRITE access code. The secret READ-WRITE access code should be setup as a secret in Github Actions.

samijaber (approved):
@FrozenPandaz Thank you for putting this together 🙏🏽

PR #893, opened 2022-11-10
fix: transform imports regex

raymondmuller:
## Description

If a custom extension is provided in the Mitosis config, it currently only replaces the first match, as the global flag is missing from the regex.

### DIFF
--- a/packages/cli/src/build/helpers/transpile.ts
+++ b/packages/cli/src/build/helpers/transpile.ts
@@ -22,7 +22,7 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
       `${getFileExtensionForTarget({ type: 'import', target, options })}$1`,
     )
     .replace(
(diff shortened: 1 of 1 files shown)


packages/cli/src/build/helpers/transpile.ts
```diff
@@ -22,7 +22,7 @@ export const transformImports = (target: Target, options: MitosisConfig) => (cod
       `${getFileExtensionForTarget({ type: 'import', target, options })}$1`,
     )
     .replace(
-      `.${options.extension}`,
+      new RegExp(`.${options.extension}`, 'g'),
```
samijaber, line 25: I realize now we probably wanna make sure it's at least followed by quotes, to avoid replacing `.svelte` strings in random parts of the code:

```suggestion
      new RegExp(`.${options.extension}['"]`, 'g'),
```

samijaber (approved):

Issue #847, opened 2022-10-26
defaultProps: default function type is lost


kingzez:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKNroDQoYAi6AZgIbIA2MAClBGKnAL5ykMhwDkAAgEbJgVACbooAOmAQA9CGA5UwVNwDcAKHXAAdjHEUAxujgAhZDBzb6jZgnVw4SgF7oAXI5hQdAcw0ORyuT8VOgi7vwQECHk2n5wENoAwlTABgDW7gAUbnAxAJ4AlHAAvAB8cABuEMAiGiya6AAekLBwYhTU8KTI2gYw0tqm5paZYAxM7mYWCdZMRXYOaOgkHTSzqAA8U5brpZkLDo7ALu7cIKHAyCDcePaHAahBIWHs5FQYt4fxSSnpWUVlRAsAgyGRwHAiCDsYCNODyRTKOAGCDgYTGbSRMDgqEIeoOFgFDR3KDoGDIKCDTJ3BwbQTTbTUw4JZKpNLFBDZCroXQA8qZACEYxsEgeT1CcAA%2FHAhUwJMzfmlOdyYEV3D12jpQgU8YdSoyHNsEoyNjI6ZY9Q5Cep6kA%3D

### Expected Behaviour

`onClick` default value should be  `() => {}`, but vue got `{}`, react got lost.

```js
import { useStore, useDefaultProps } from '@builder.io/mitosis';

interface ButtonProps {
  size: string;
  disabled: boolean;
  onClick: (e: any) => void;
}

export default function Button(props: ButtonProps) {
  useDefaultProps<ButtonProps>({
    size: 'medium',
    disabled: false,
    onClick: () => {}, // todo fix mitosis compile noop to {}
  });

  return (
    <button
      onClick={(event) => (!props.disabled ? props.onClick(event) : undefined)}
    >
      Button
    </button>
  );
}

```

vue
```js
<template>
  <button @click="!disabled ? onClick($event) : undefined">Button</button>
</template>

<script>
export default {
[...]

PR #875, opened 2022-11-04
fix(core): #832 handle arrow functions within objects

raymondmuller:
## Description
Fixes #832 

This is currently an issue in other frameworks as well (such as React and Solid).
Converting an arrow expression inside an object property to an object method seems to solve it.

Also fixed a typo in this PR (tarverse > traverse)

Before:

Screenshot 2022-11-04 at 15 52 54
Screenshot 2022-11-04 at 15 53 01

After:
832-solid
832-react
832-svelte

PR #896, opened 2022-11-11
Fix/default props function type lost

### LINKED ISSUE #913: Default props can't be `undefined` or `null`

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [X] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=M4NwpgNgLmQ%3D&code=JYWwDg9gTgLgBAbzgVwM4FMAi6BmBDZAGxgAUoIxU4BfOHckOAcgAEAjZYQgE3SgDpgEAPQhgMCKmComAbgBQ89AA9IsOL3xF4OZADsAxjCF64AWQCeAYQjgIe9HpgAKMOUoBKRPLgoM2LWIyClRnBB9fOD08EHQALhQ9TWAHbgAaCOoPBQiodBhkKFNnCN8AHm5gADcAPgQ3EP5o2Ooy4UraiOz5aiA

### Expected Behaviour

The output should be:
```svelte
<script>
  export let name = undefined;
</script>

<div>{name}</div>
```

### Actual Behaviour

```svelte
<script>
  export let name;
</script>

<div>{name}</div>
```

### Additional Information

In Svelte, props without an assigned value are considered as a _required_ prop. The example above might work as well since it's in JavaScript. However, consider the next example in Typescript.

```tsx
import { useDefaultProps } from '@builder.io/mitosis';

interface MyComponentProps {
  name?: string;
}

export default function MyComponent(props: MyComponentProps) {
  useDefaultProps<MyComponentProps>({
    name: undefined,
  });

  return (
    <div>{props.name}</div>
  );
}
```

which yields:

```svelte
<script context="module" lang="ts">
  interface MyComponentProps {
    name?: string;
  }
</script>

<script lang="ts">
  export let name: MyComponentProps["name"];
</script>

<div>{name}</div>
```

Which apparently is not enough for TypeScript (and/or Svelte?). It needs explicit `undefined` (as follow) so that it will stop saying that "Property 'name' is missing".

```svelte
<script context="module" lang="ts">
  interface MyComponentProps {
    name?: string;
  }
</script>

<script lang="ts">
  export let name: MyComponentProps["name"] = undefined;
</script>

<div>{name}</div>
```

This issue might need additional check for other frameworks since I'm not familiar with them.

kingzez:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

Fix https://github.com/BuilderIO/mitosis/issues/847
Fix #913


packages/core/src/__tests__/__snapshots__/parse-jsx.test.ts.snap
```diff
@@ -7674,9 +7702,22 @@ exports[`Parse JSX > defaultPropsOutsideComponent 1`] = `
...
   "defaultProps": {
-    "link": "https://builder.io/",
-    "openLinkInNewTab": false,
-    "text": "default text",
+    "link": {
+      "code": "https://builder.io/",
+      "type": "property",
+    },
+    "onClick": {
+      "code": "() => {",
```
samijaber, line 7710: This doesn't look right...it's missing `}` from the end 😮

samijaber (changes requested):
Amazing, thank you for this PR!

Just need to fix the issue with the trailing `}` being removed in the JSX parser.

packages/core/src/parsers/jsx/function-parser.ts
```diff
@@ -20,6 +20,21 @@ export function generateUseStyleCode(expression: babel.types.CallExpression) {
...
 
+function processDefaultPropsValue(value: any) {
+  if (types.isFunctionExpression(value) || types.isArrowFunctionExpression(value)) {
+    const code = generate(value)
+      .code.trim()
+      // Remove arbitrary block wrapping if any
+      // AKA
+      //  { console.log('hi') } -> console.log('hi')
+      .replace(/^{/, '')
+      .replace(/}$/, '');
```
kingzez, line 31: @samijaber  I  trailing } here, I follow your [issues tips](https://github.com/BuilderIO/mitosis/issues/847#issuecomment-1307439504), I thought it should be done that way,  I change it later ^-^

samijaber: Yup, sorry for that! Mistake on my part

kingzez: Hahaha，it's my fault

packages/core/src/parsers/jsx/function-parser.ts
```diff
@@ -32,7 +41,7 @@ export function parseDefaultPropsHook(
       if (i.key?.name) {
         component.defaultProps = {
           ...(component.defaultProps ?? {}),
-          [i.key?.name]: i.value.value,
+          [i.key?.name]: processDefaultPropsValue(i.value),
```
samijaber, line 44: Actually, I recently merged logic that handles all of this logic cleanly: `parseStateObjectToMitosisState`

You can simplify this function and write:

```ts
  if (types.isObjectExpression(firstArg)) {
    component.defaultProps = parseStateObjectToMitosisState(firstArg)
  }
```

If you don't mind doing that, would be amazing 🙏🏽

kingzez: I tried, but change to `parseStateObjectToMitosisState `, the snapshot mismatched,  property `code` has extra double quote and function with name

```js
 FAIL  src/__tests__/parse-jsx.test.ts > Parse JSX > defaultProps
AssertionError: Snapshot `Parse JSX > defaultProps 1` mismatched
  - Expected  - 4
  + Received  + 4

        set": {},
      },
      "defaultProps": {
        "link": {
  -       "code": "https://builder.io/",
  +       "code": "'https://builder.io/'",
          "type": "property",
        },
        "onClick": {
  -       "code": "() => {
  +       "code": "onClick() {
      console.log('hi');
    }",
          "type": "method",
        },
        "openLinkInNewTab": {
  -       "code": false,
  +       "code": "false",
          "type": "property",
        },
        "text": {
  -       "code": "default text",
  +       "code": "'default text'",
          "type": "property",
        },
      },
      "exports": {},


```

samijaber: Oh, right! Sorry for the confusion. We recently changed this part of our JSON schema, during the time you made your PR. 

We actually no longer need to `JSON.stringify` the output in the generators! So everything that you do this:

```ts
return `${prop}: ${isMethod ? `${value}` : json5.stringify(value)}`;
```

You can now replace with just:

```ts
return `${prop}: ${value}`;
```

the `parse-jsx` snapshot are going to need to be updated, but the final code snapshots in React/Vue/etc. will remain the same.

Sorry for sending you on a long journey with this PR: I forgot to tell you all this information at the very start. Thank you for your patience 🙏🏽

kingzez: Double quotes can solve it, but function code get `onClick() {
      console.log('hi');
    }`,  the defaultProps's function should not has function name, it should be `() => {
      console.log('hi');
    }`, should not converted

Is defaultProps that should handled as state? 🤔

kingzez: https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/parsers/jsx/state.ts#L108-L124

```ts
if (types.isFunctionExpression(item.value)) {
  return {
    code: parseCode(item.value),
    type: 'function',
  };
} else if (types.isArrowFunctionExpression(item.value)) {
  const n = babel.types.objectMethod(
    'method',
    item.key as babel.types.Expression,
    item.value.params,
    item.value.body as babel.types.BlockStatement,
  );
  const code = parseCode(n);
  return {
    code: code,
    type: 'method',
  };
}
```

I see this part is dealing with that PR https://github.com/BuilderIO/mitosis/pull/875

samijaber (approved):
Amazing @kingzez , thank you for this PR 😄

PR #902, opened 2022-11-12
fix(fiddle): hide svelte output for sveltosis input

raymondmuller:
## Description

When Sveltosis is selected as the input tab in the fiddle, we should hide the Svelte output tab.
If it is currently selected, we switch to another tab.


samijaber (approved):

milahu:
please revert

its interesting to see the `svelte -> mitosis -> svelte` transform

PR #914, opened 2022-11-15
fix(core): sveltosis reactive and event handlers

raymondmuller:
## Description
Also added a test for this bug flagged by someone from the Svelte discord
```
<script>
  let name = 'Steve';
  let a = 2
  let b = 5
  let result;

  $: calculateResult(a, b)

  function calculateResult(a, b) {
    result = a * b
  }
</script>

<div>
  <input type="number" bind:value={a} />
  <input type="number" bind:value={b} />
  Result: {result}
</div>
```


packages/core/src/__tests__/__snapshots__/svelte.test.ts.snap
```diff
@@ -4833,6 +4892,43 @@ exports[`Svelte > svelte > Typescript Test > each 1`] = `
...
+
+    
+    
+    
+    
+    
+
+    function log(msg = 'hello') {
+console.log(msg);
+}
```
samijaber, line 4914: non-formatted snapshots usually mean there's invalid JS syntax that causes prettier to fail. Worth looking into this one here.

packages/core/src/__tests__/__snapshots__/svelte.test.ts.snap
```diff
@@ -4833,6 +4892,43 @@ exports[`Svelte > svelte > Typescript Test > each 1`] = `
...
+    
+    
+    
+
+    
+
+    
+  </script>
+
+  <div ><button  on:click=\\"{a => {log('hi')}}\\" >Log</button><button  on:click=\\"{event => {log(event)}}\\" >Log</button><button  on:click=\\"{event => {log(event)}}\\" >Log</button><button  on:click=\\"{ => {log('bye')}}\\" >Log</button></div>"
```
samijaber, line 4929: I think this is the issue:

```ts
on:click=\\"{ => {log('bye')}}\\"
```

no fn args

samijaber (approved):

PR #923, opened 2022-11-21
feat(core): fallback slot content and named slot improvements

raymondmuller:
## Description

**Allows you to set default slot content** 
E.g.
```
import { Slot } from '@builder.io/mitosis';

type Props = { [key: string]: string };

export default function SlotCode(props: Props) {
  return (
    <div>
      <Slot>
        <div class="default-slot">Default content</div>
      </Slot>
      <Slot name="footer">Copyright 2022</Slot>
    </div>
  );
}
```

**Define named slots like this**
```
<Slot name="top"/>
```

Added the generator implementation for Svelte, Vue and React


samijaber (approved):
👌🏽

Issue #927, opened 2022-11-28
Angular Output Attribute name is all lowercase


rub-sky:
### I am interested in helping provide a fix!

No

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Qwik
- [X] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

[...]

PR #941, opened 2022-12-01
#927 angular event bindings

rub-sky:
## Description
Fix for issue #927
https://github.com/BuilderIO/mitosis/issues/927

### Changes
* Angular output event binding requires that the name be camel-cased and rendered with parens when using the prefix `on`. So I removed the `toLowercase()` to keep the camel-case format.
* Added tests with updated snapshots


rub-sky:
Having some merge issues. Will close this one to open a new one.

Issue #916, opened 2022-11-16
`htmlFor` does not transpile properly in none-JSX frameworks


rootEnginear:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [X] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [X] Svelte
- [X] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAwhALaR7B4wAUYUEYAzgJRwA3gCg4cKMBjoo5NqLFwAPAmIA3AHzyFS4njDp4xBAF4ARKlNwA9Ju1jFmVACNgmOAAsY9TADFoZi3UAQUUrRxdMWwVQlQ15PgBuYQBfIA%3D

### Expected Behaviour

```vue
<template>
  <div>
    <input id="a" />
    <label for="a">A</label>
  </div>
</template>
```

### Actual Behaviour

```vue
<template>
  <div>
    <input id="a" />
    <label htmlFor="a">A</label>
  </div>
</template>
```

### Additional Information

(might be related to #526) In none-JSX frameworks (please tell me if I'm incorrect), `htmlFor` attribute should be transpiled as `for`.

samijaber:
Thanks for filing this bug, and being interested in contributing a fix! This one is a good first issue :)

Easiest way to fix this for all generators is to:

## Add logic in the JSX parser to map the `htmlFor` key to `for`
- We need to do this for `bindings`: https://github.com/BuilderIO/mitosis/blob/24688615f284a58c256211147cdf85591be182cf/packages/core/src/parsers/jsx/element-parser.ts#L206

- and also add it for `properties`: 
https://github.com/BuilderIO/mitosis/blob/24688615f284a58c256211147cdf85591be182cf/packages/core/src/parsers/jsx/element-parser.ts#L191


Now, the Mitosis JSON will store it as `for` everywhere. This means it works for all non-JSX frameworks. We will need to update the JSX generators (SolidJS, React, Qwik) to map `for` back to `htmlFor` for it to work correctly there.


## React
React has a binding mapper object to do this: 

[...]

PR #943, opened 2022-12-03
feat: properly transpile `htmlFor` into `for`

rootEnginear:
## Description

As guided in #916

- Transform all `htmlFor` into `for` in parsing stage. I created a `HTML_ATTR_FROM_JSX` lookup object, and `transformAttributeName(name)` function. Generally, you could do `HTML_ATTR_FROM_JSX[name] ?? name` or type `HTML_ATTR_FROM_JSX` to be broader. In this case, I went with the proper way of TS by create `objectHasKey` function to handle proper type checking.
- Transform `for` attribute back into `htmlFor` in React, Solid, and Qwik. In React and Qwik, it was just adding an additional condition. In Solid, I created similar function to the previous item to replace the key.

I apologize for a late reply. If you need any modification please feel free to tell me.


samijaber (commented):
Amazing!

2 things you need to do:

- merge `main` into your branch
- after that, run `yarn snapupdate` to commit the new snapshot tests

samijaber:
@rootEnginear sorry for that trouble. I think we accidentally kept a flag `true` that we should revert: https://github.com/BuilderIO/mitosis/blob/4aa8eddf6a049998b481a11ac1c1e9214dd89c50/packages/core/src/__tests__/qwik.test.ts#L20

I don't use Windows, so I wouldn't know about this OS specific issue.

samijaber (approved):
Phenomenal. Thank you for the contribution to Mitosis! ⚡

PR #949, opened 2022-12-12
feat(eslint): no map function in jsx return body

### LINKED ISSUE #812: ESLint Rule: no `.map` in JSX render body
We want to encourage folks to use `<For>` to loop through arrays in the render body, instead of using `Array.map()` which might not reliably work (depending on the complexity of the JSX render body).

decadef20:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

fix #812

### DIFF
--- a/packages/eslint-plugin/src/configs/recommended.ts
+++ b/packages/eslint-plugin/src/configs/recommended.ts
@@ -20,6 +20,7 @@ const recommendedRules: Record<RulesKeys, string> = {
   '@builder.io/mitosis/use-state-var-declarator': 'error',
   '@builder.io/mitosis/static-control-flow': 'error',
   '@builder.io/mitosis/no-var-name-same-as-prop-name': 'error',
+  '@builder.io/mitosis/no-map-function-in-jsx-return-body': 'warning',
 };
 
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/configs/recommended.ts
```diff
@@ -20,6 +20,7 @@ const recommendedRules: Record<RulesKeys, string> = {
   '@builder.io/mitosis/use-state-var-declarator': 'error',
   '@builder.io/mitosis/static-control-flow': 'error',
   '@builder.io/mitosis/no-var-name-same-as-prop-name': 'error',
+  '@builder.io/mitosis/no-map-function-in-jsx-return-body': 'error',
```
samijaber, line 23: I think for this one, we want a `warning`, not an `error`, because technically it will work. So it's more of a suggestion

```suggestion
  '@builder.io/mitosis/no-map-function-in-jsx-return-body': 'warning',
```

decadef20: updated

samijaber (approved):

PR #954, opened 2022-12-16
fix: allow vue to output standard html tags like feFlood instead of converting them to kebab case

### LINKED ISSUE #952: Some SVG output breaks when exporting to Vue2 and Vue3

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [X] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnAIIBuqMqUAsvQEabAAUAlHAN4BQfOHCjAY6KOXaChcADwBnKgHNpMuFWLAA7gCEIIALw8ARAAY4543ADUcACwAmG3CtXbjgL6qZyYpkwGxnhkwMbeQlAQbIHEALZKYWpCWsQIMAAWRp7hcOnAxErpMFkOXmoAfDmySMjylUkysr7YwFA5aqmBYCK%2BIAD6fc0wraaDiQ0%2BfsNQAKp4xDDygejyrQDKYKgEwADyeDOr4w31E0JNwABimBAQCMLA8ljFxjpbANZKkfgIAJKxqEpLsQQFYAPQnU7nHRsPB3Yh4QJrCDibYAcSgqDA6WIBCs8IcgVeBA%2BX1hfwBQJB90e2EC8nSmNCcHB7UayGAqIw8nkxFQeGh4jg8hgCAAIsANLRSAieAB2DzUp6BFDsogARkG0GAn2RsIFUDBEKSslBQ1aRrOoJqdW8JsUShOnAA3HwvEA%3D%3D

### Expected Behaviour

feFlood, feBlend, etc should stay in their existing case.

### Actual Behaviour

fsFlood gets converted to fe-flood which is not a valid HTML/SVG tag.

### Additional Information

I believe we just need to use the HTML_TAGS list to make sure we are not converting case of legitimate tags inside the Vue generator.

cmgriffing:
## Description

SVG has some camel-cased tags that were being converted to kebab-case which broke SVG output.

This PR checks if the tag is a valid tag first before converting to kebab case.

I'm not sure why the `snapupdate` added the extra whitespace.

fixes #952


cmgriffing:
Let's hold on the merge of this, I don't think I have a good test of it yet.

cmgriffing:
Ok, I added an SVG test to the base set of tests because I think SVG is worth preventing regressions on.

packages/core/src/__tests__/shared.ts
```diff
@@ -521,7 +523,7 @@ export const runTestsForTarget = <X extends BaseTranspilerOptions>({
                   try {
                     expect(getOutput()).toMatchSnapshot();
                   } catch (error) {
-                    expect(getOutput).toThrowErrorMatchingSnapshot();
+                    expect(getOutput()).toThrowErrorMatchingSnapshot();
```
samijaber, line 526: I think we want to stick with providing `getOutput` here: https://jestjs.io/docs/expect#tothrowerrormatchingsnapshothint

We need to provide the function that throws an error lazily without calling it, so that jest can collect the error and snapshot it. Otherwise, `getOutput()` will throw and we'll never be able to check/store the snapshot

cmgriffing: Yeah, I reverted that already in my latest changes. It seemed like it could have been the issue with my failing tests.

cmgriffing: This is the test that fails now and why I thought to check the getOutput function:
➤ YN0000:    ❯ src/__tests__/angular.import.test.ts > Angular with Preserve Imports and File Extensions > jsx > Typescript Test > svgComponent
➤ YN0000:      → snapshot function didn't throw

Fun fact: this error shows in my local but not my VM. (the opposite of the style/prettier issue I had). My env is cursed.

cmgriffing: I finally understand what's going on there.

The snapshot was a mismatch, but the error was getting caught and then expecting the thrown snapshot.

I wonder if there is something more that could be logged to make that more obvious. Not sure though.

That said, the PR should be good now.

samijaber (approved):
Thanks @cmgriffing 🙏🏽

PR #957, opened 2022-12-18
fix: update failing snapshots on main when no changes have been made

cmgriffing:
## Description

I have been trying for a day or two to understand why my other PR has failing snapshots. It turns out that the main has failing snapshots when no changes have been made.

This updates the snapshots on main. I am not 100% certain this is the right fix.

I might have an inconsistent prettier config somehow, or maybe another submitter had one. Please verify the issue independently.

I wonder if a Dockerfile would be useful to isolate developer environments from local testing.


cmgriffing:
Seeing this fail, it must be an issue on my machine, but I'm not sure what.

PR #958, opened 2022-12-19
feat: dev script for mitosis cli

rgodha24:
The main changes made were in the packages/cli/src/build/dev.ts file and packages/cli/src/commands/dev.ts. 

Basically, it runs something like the "build" command on each file change, only on the files changed.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -25,6 +25,19 @@ You can get more information by running `mitosis --help`
   - transpiles them as-is to JS
 - Performs necessary transformations to both Mitosis & non-Mitosis files so that the output folder is coherent and valid (like renaming all component imports in a Svelte target such that they match the output name, ending in `.svelte`)
 
+## `mitosis dev`
+
+**Warning: mitosis dev is still experimental.**
+
+`mitosis dev` is meant to reflect changes in your mitosis components more quickly. It: 
+
+- Reads the config in `mitosis.config.js` (also could specify config file by option: `--config=<file>`)
+- Identifies a source folder
+- Starts with running mitosis build (see above)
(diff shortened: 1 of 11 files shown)


rgodha24:
I have not added tests because I'm not very good at writing them. If you could provide me with guidance, that would be great!

rgodha24:
@samijaber anything else I can do to get this pr merged?

ohkimur:
@samijaber Isn't this feature very important? Does it need some more love? I would very much use this feature. Looking more at what it does, I don't think it should be called `dev` but rather it should be a `--watch` flag.

@rgodha24 I guess you closed this issue because it passed a long time with no answer, am I right? Would you care to reopen this issue and refactor it from dev to `--watch`? I'm willing to help if necessary.

rgodha24:
mitosis already has a "--watch" flag, but that's much slower, especially when there's large amounts of files. this only rebuilds the files that are actually changed which is why I called it "dev"

ohkimur:
@rgodha24 I see. Maybe we should try to improve that one instead. What do you think about that?

kingzez:
It would be perfect if we supported incremental build.

rgodha24:
I'd be down to reimplement a pr starting on may 11 (my last ap test 💀).

samijaber:
Apologies for the delay in getting to this PR @rgodha24. This is definitely a very useful feature! And there is currently no `--watch` flag for `mitosis build`, so re-implementing this as `mitosis build --dev` would be very useful indeed. 🙏🏽

If you have any doubts or questions, I would suggest opening a github issue outlining how you plan on doing it so we can go over it there before a PR is put up.

PR #982, opened 2023-01-15
Remove padding from output menu

elfpie:
## Description

Remove unnecessary padding from the fiddle output menu

Before:
image

After:
image


samijaber (approved):
Thanks for contributing! Welcome to the project :)

PR #984, opened 2023-01-15
Preserve types for SolidJS

### LINKED ISSUE #583: Solid: Generate Typescript code

We would have to repeat step 3, but for Vue: https://github.com/BuilderIO/mitosis/issues/372#issuecomment-1133471614

elfpie:
## Description

Resolve #583

This code is really similar to the React generator, the only difference is how we handle refs, the [recommended](https://www.solidjs.com/guides/typescript#the-ref-attribute) option for SolidJS is to use an undefined union type and check if the ref is valid before using it, I thought that this would be too complex to implement and choose to use the second recommended option which is using '!' in the ref assignment.

### DIFF
--- a/packages/cli/src/build/helpers/extensions.ts
+++ b/packages/cli/src/build/helpers/extensions.ts
@@ -16,8 +16,6 @@ export const getFileExtensionForTarget = ({
     case 'alpine':
     case 'html':
       return '.html';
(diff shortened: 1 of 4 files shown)


samijaber (approved):
Wonderful! Thank you for contributing this 🙏🏽

PR #992, opened 2023-01-19
Fix extensions

### LINKED ISSUE #991: Transformed imports contain incorrect file extension

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [X] React
- [X] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://stackblitz.com/edit/node-9aesvr?file=src%2Fcomponents%2Findex.ts,dist%2Freact%2Fsrc%2Fcomponents%2Findex.js

### Expected Behaviour

Typically, modern build tools will auto resolve modules when the file extension is omitted (with a proper loader/config). When running `mitosis build`, the module extension should be omitted so it can be resolved correctly.

_Expected output from `mitosis build`:_
```js
// src/components/index.js
import { default as default2 } from "./Button";
export {
  default2 as Button
};
```

### Actual Behaviour

When using a **named or default** exports for a component in an `index.{ts,js}` file, mitosis transforms on the file type extension (`Button.js`). This causes a problem when consuming the import through bundlers/transpilers causing the import to error since the exported component extension is `.jsx`.

_Actual output from `mitosis build`:_
```js
// src/components/index.js
import { default as default2 } from "./Button.js";
export {
  default2 as Button
};

### Additional Information

I've been running my built components through `esbuild` with the proper js -> jsx loader setup. Here's the build error (for others running into this):

```bash
✘ [ERROR] Could not resolve "./Button.js"

    ../dist/react/src/components/index.js:1:33:
      1 │ import { default as default2 } from "./Button.js";
```

I've traced this to these lines:
- https://github.com/BuilderIO/mitosis/blob/main/packages/cli/src/build/helpers/transpile.ts#L27-L30
- https://github.com/BuilderIO/mitosis/blob/main/packages/cli/src/build/helpers/extensions.ts#L39-L40

I tried to do due-diligence to find a similar issue, but let me know if this is a duplicate 🙏 

Regarding a path forward, I was experimenting locally with config options (e.g. `omitFileExtensions`), but not sure that is the best path forward. You could use `options[target].transpiler.format === 'esm'` to omit file extensions from imports, but very hesitant with that solution. Let me know if y'all have something better in mind.

samijaber:
## Description

- closes #991

### DIFF
--- a/packages/cli/src/build/helpers/extensions.ts
+++ b/packages/cli/src/build/helpers/extensions.ts
@@ -28,16 +28,17 @@ export const getFileExtensionForTarget = ({
       return '.ts';
     case 'lit':
       return '.ts';
+
+    // all JSX frameworks
     case 'solid':
     case 'qwik':
-      return isTs && type === 'filename' ? '.tsx' : '.jsx';
     case 'react':
     case 'reactNative':
     case 'rsc':
       switch (type) {
-        // we can't have `.jsx`/`.tsx` extensions in the import paths.
         case 'import':
-          return isTs ? '.ts' : '.js';
+          // we can't have `.jsx`/`.tsx` extensions in the import paths, so we stick with implicit file extensions.
+          return '';
         case 'filename':
           return isTs ? '.tsx' : '.jsx';
       }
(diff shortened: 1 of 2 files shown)


chanceaclark (approved):
🙇

Issue #1000, opened 2023-01-29
Overrides strip 'tsx', 'ts', 'jsx', 'js' from filename RegExp error


kingzez:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [X] Qwik
- [X] React
- [X] React-Native
- [X] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://github.com/BuilderIO/mitosis/pull/976/files#diff-e026442b3941aadaf063789fef3a8a592cda2de381f078b2829ee00db908e439R38

### Expected Behaviour

> All JSX targets, we want to be flexible and allow any possible extension

`filenameStrippedFromExtensions` regexp replace not success，cause filePath like this `'src/compone/mask/mask.tsx.tsx'`

`(.jsx?|.tsx?)`  dot should be escape, should fix to `(\.jsx?|\.tsx?)` 

case:  "src/components/mask/mask.tsx"

test 1:  "src/compone`nts`/mask/mask.tsx"  
test 2:  "src/compone`nts`/mask/mask.ts"  
test 3:  "src/compone`nts`/mask/mask.jsx"  
test 4:  "src/compone`nts`/mask/mask.js"  

Match `.tsx` `.ts`  `.jsx`  `js`

### Actual Behaviour

case:  "src/components/mask/mask.tsx"

test 1:  "src/components/mask/mask.tsx"  
test 2:  "src/components/mask/mask.ts"  
test 3:  "src/components/mask/mask.jsx"  
test 4:  "src/components/mask/mask.js"  

Match `nts`  not `.tsx` `.ts`  `.jsx`  `js`

PR #1001, opened 2023-01-29
fix: filename strip extension regexp

kingzez:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

- Fixed https://github.com/BuilderIO/mitosis/issues/1000

### DIFF
--- a/packages/cli/src/build/helpers/overrides.ts
+++ b/packages/cli/src/build/helpers/overrides.ts
@@ -35,7 +35,7 @@ const getOverrideFilenames = ({
     case 'solid':
     case 'qwik': {
       // strip 'tsx', 'ts', 'jsx', 'js' from filename
-      const filenameStrippedFromExtensions = filename.replace(/(.jsx?|.tsx?)/, '');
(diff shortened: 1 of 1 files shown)


packages/cli/src/build/helpers/overrides.ts
```diff
@@ -35,7 +35,7 @@ const getOverrideFilenames = ({
     case 'solid':
     case 'qwik': {
       // strip 'tsx', 'ts', 'jsx', 'js' from filename
-      const filenameStrippedFromExtensions = filename.replace(/(.jsx?|.tsx?)/, '');
+      const filenameStrippedFromExtensions = filename.replace(/(\.jsx?|\.tsx?)/, '');
```
samijaber, line 38: Should also add a `$` to make sure we're matching the end of the string:

```suggestion
      const filenameStrippedFromExtensions = filename.replace(/(\.jsx?|\.tsx?)$/, '');
```

samijaber (approved):
Thanks for the fix 🙏🏽

PR #1002, opened 2023-01-29
chore: fix example config type notation

pd4d10:
## Description

Fix example config type notation


samijaber (approved):

samijaber:
Thanks for the contribution @pd4d10 :)

PR #1009, opened 2023-01-31
CLI: `exclude` glob

samijaber:
## Description

- add `exclude` config to exclude files that match a glob

### DIFF
--- a/packages/cli/src/build/helpers/files.ts
+++ b/packages/cli/src/build/helpers/files.ts
@@ -1,7 +1,7 @@
 import glob from 'fast-glob';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/cli/src/build/helpers/files.ts
```diff
@@ -0,0 +1,15 @@
+import glob from 'fast-glob';
+
+/**
+ * get all files that match `files`glob, and filter out those that match `exclude` glob.
```
samijaber, line 4:
```suggestion
 * get all files that match `files` glob, and filter out those that match `exclude` glob.
```

PR #1010, opened 2023-02-01
add generator for taro

PengBoUESTC:
## Description

Add a short description of:
- what changes you made, 

add a generator for Taro framework,

- why you made them, and 
 
https://github.com/NervJS/taro

### DIFF
--- a/packages/core/src/generators/react/generator.ts
+++ b/packages/core/src/generators/react/generator.ts
@@ -27,6 +27,7 @@ import {
   runPostJsonPlugins,
   runPreCodePlugins,
   runPreJsonPlugins,
+  runPreComponentsPlugin,
 } from '../../modules/plugins';
 import { selfClosingTags } from '../../parsers/jsx';
 import { MitosisComponent } from '../../types/mitosis-component';
@@ -471,8 +472,8 @@ export const componentToReact: TranspilerGenerator<ToReactOptions> =
   };
 
 // TODO: import target components when they are required
-const getDefaultImport = (options: ToReactOptions): string => {
-  const { preact, type } = options;
+const getDefaultImport = (json: MitosisComponent, options: ToReactOptions): string => {
+  const { preact, type, plugins } = options;
   if (preact) {
     return `
     /** @jsx h */
@@ -488,7 +489,7 @@ const getDefaultImport = (options: ToReactOptions): string => {
   if (type === 'taro') {
     return `
     import * as React from 'react';
-    import { View, Image, Text, Button, Input, Textarea, Form } from '@tarojs/components';
+    ${runPreComponentsPlugin(json, plugins ||[])}
     `;
   }
 
@@ -603,7 +604,7 @@ const _componentToReact = (
   };
 
   let str = dedent`
-  ${getDefaultImport(options)}
+  ${getDefaultImport(json, options)}
   ${styledComponentsCode ? `import styled from 'styled-components';\n` : ''}
   ${
     reactLibImports.size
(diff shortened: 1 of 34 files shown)


samijaber:
Wow, very interesting @PengBoUESTC . I've never heard of Taro before. I'm open to having this in our core repo.

- Make sure to add a `taro` entry in this object https://github.com/BuilderIO/mitosis/blob/66fb90858e6d4d02cb9d1968b62426b523de4029/packages/core/src/__tests__/shared.ts#L271-L272

- Add all of the tests in the array,
- add a file like this one for Taro https://github.com/BuilderIO/mitosis/blob/66fb90858e6d4d02cb9d1968b62426b523de4029/packages/core/src/__tests__/svelte.test.ts#L1-L7
-  run `yarn snapupdate` to generate snapshots.

PengBoUESTC:
> 

add test file 
and make some change for `Plugin` to get the `components` required

packages/core/src/modules/plugins.ts
```diff
@@ -46,3 +46,18 @@ export const runPostCodePlugins = (code: string, plugins: Plugin[], options?: an
...
+) => {
+  const imports = [];
+  for (const plugin of plugins) {
+    const preFunction = plugin(options).components?.pre;
+    if (preFunction) {
+      imports.push(preFunction(json) || '');
+    }
+  }
+  return imports.join('\n');
+};
```
samijaber, line 63: If you want to write a plugin that alters the `imports`, you don't need a new type of plugin. You can reuse the preJSON plugin:

```ts
json: {
  pre: (json) => {
    const myExtraImports = [/* */]
    json.imports.push(...myExtraImports)
  }
}
```

I think we should remove this `runPreComponentsPlugin` and stick with use what's already here, since it does everything that you need.

PengBoUESTC: wow that‘s great , thanks a lot.

PengBoUESTC: wow that‘s great , thanks a lot.

packages/core/src/generators/react/generator.ts
```diff
@@ -470,6 +470,30 @@ export const componentToReact: TranspilerGenerator<ToReactOptions> =
...
+  if (preact) {
+    return `
+    /** @jsx h */
+    import { h, Fragment } from 'preact';
+    `;
+  }
+  if (type === 'native') {
+    return `
+    import * as React from 'react';
+    import { View, StyleSheet, Image, Text } from 'react-native';
```
samijaber, line 485: We made changes to that recently, you'll want to grab the latest version

packages/core/src/generators/taro.ts
```diff
@@ -0,0 +1,183 @@
...
+  form: 'Form',
+  textarea: 'Textarea',
+};
+
+/**
+ * Plugin that handles necessary transformations from React to React Native:
+ * - Converts DOM tags to @tarojs/components
+ * - Removes redundant `class`/`className` attributes
+ */
+const PROCESS_REACT_NATIVE_PLUGIN: Plugin = () => ({
```
samijaber, line 126:
```suggestion
const PROCESS_TARO_PLUGIN: Plugin = () => ({
```

samijaber (commented):
Amazing work, thanks! Left a few comments

packages/core/src/generators/taro.ts
PengBoUESTC: done

samijaber (approved):
Thank you for your contribution! Welcome to Mitosis 🙏🏽

PR #1016, opened 2023-02-03
fix: allow stencil to output ts files when typescript flag is on

cmgriffing:
## Description

I was adding stencil to the list of support for my library and noticed that it only outputs js files even if the typescript flag is explicitly true.


cmgriffing:
Let me do some more testing for this real quick.

PR #1018, opened 2023-02-04
refactor: react-native components imports

PengBoUESTC:
## Description

Add a short description of:
- what changes you made, 

delete the components imports in `react` generator
create components imports by `json.imports`

- why you made them, and 

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/context.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/context.test.ts.snap
@@ -750,10 +750,10 @@ export default function ComponentWithContext(props) {
 
 exports[`Context > Use and set context in components 3`] = `
 "import * as React from \\"react\\";
-import { View, StyleSheet, Image, Text } from \\"react-native\\";
 import { useContext } from \\"react\\";
 import Context1 from \\"@dummy/1\\";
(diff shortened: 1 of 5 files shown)


samijaber:
The problem with this is that we want the flexibility of having the imports available, like when the user wants to use a React Native import that wasn't initially used by the generator.

I think for now, I'll want to keep this as-is. There is no downside of having more things imported, tree-shaking in all modern tools takes care of not importing code that is unused.

PengBoUESTC:
> The problem with this is that we want the flexibility of having the imports available, like when the user wants to use a React Native import that wasn't initially used by the generator.
> 
> I think for now, I'll want to keep this as-is. There is no downside of having more things imported, tree-shaking in all modern tools takes care of not importing code that is unused.

Okk

jonsherrard:
Could we revisit this, unused imports trigger lots of eslint errors, and it's a pain to have to post-process them to remove them.

> The problem with this is that we want the flexibility of having the imports available, like when the user wants to use a React Native import that wasn't initially used by the generator.

Most editor's will auto import the primatives as you type them or with a quick `.` on the error. Having unused code imports all over source code doesn't feel good.

In general I'd like to not modify mitosis output after build either.

PR #1019, opened 2023-02-05
fix: calculate hasStaticClasses after they have been aggregated

cmgriffing:
## Description

Static classes were not being passed to the resulting code. After inspection, it seems that hasStaticClasses was being calculated before adding to the array.


samijaber (approved):
Thanks @cmgriffing !

Issue #1020, opened 2023-02-06
Props.slotXXX should be convert to $slots.XXX in vue transfrom.Props.slotXXX should be convert to $$slots.XXX in svelte transfrom.


findlay-best-wishes:
# Reproduction case

https://mitosis.builder.io/?outputTab=M4NwpgNgLmQ%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKNroDQoYAi6AZgIbIA2MAClBGKgZgBYQDuLVE8AvnFIMQcAOQABAEbJgVACbooAOmAQA9CGA5UwVKIDcAKFCRYcAMZVyqVADlyIdKkHCxl66gB2DpwcOH0AA9TeAUKanhSZE9zGFVPOFpGCAA3RQAKMAYmAEpEQzhCdBJwmnpGVHSEAsKLCE8YdAaALjFRPBrC1B4YACUyRSbzdFbosOBPdDkO2rhu3gBhesaWlE9xyemavhyjGvN61Hgj8ka4AF4i7FwqzrhvR1aAIkhINKgnmdrUdg4kt8UrQoVAwHTuAHN0PBLJU8tVZoUoFDkFAEu4bPZHJUTo0lA98PkEQiANoAAwAJAgceg8T4%2BABaA4NJowUkAXVaWQqSnmMCWzIaXyJOzufC%2BOz2iORqLg6TuAB42Jw4BxWE1zgguUweT1%2BqRBjF0HwAHx3QryuTAFJm2rouw%2BDV2zFOdLU2mOAjwom1MmUt34hlMlasjlwLWoHWLZYsoUInZ8G2m71wRW%2FFVqzwat0%2FTj%2FVKKOAAMkLYeyEd5%2FODJpts0VPXuDqeQZZT2NmrLSmbDT48rUmB6Se9vaVHEHtbUlpSY5Tfd%2Bg92hgThjUajgADVkCNS9yK9GGir6qJofV3vAcHBybyI134OkJvd6nAR3AYORwTl%2FCun2kaFvw5G%2BT3eAOEPY9PFPF8IAvS8emvICgA%3D%3D%3D

# Expected Behaviour

- Vue: props.slotContent won't convert to $slots.content (in non Show tag)
- Svelte: props.slotContent won't convert to $$slots.content

```ts
import { useStore, useDefaultProps, Show, Slot } from '@builder.io/mitosis';
import classNames from 'classnames';

export default function Popover(props) {
  useDefaultProps({
[...]

PR #1034, opened 2023-02-13
Feat/(vue|svelte) slots convert in template and state

kingzez:
## Description

Fixed  https://github.com/BuilderIO/mitosis/issues/1020

Add a short description of:
- Vue
    - [x] [ convert slots to vue template slot syntax](https://github.com/BuilderIO/mitosis/commit/baa280da2231b704ee49ee2b5ebe2054edeb2c2e)
    - [x] optionsApi [convert slots to vue getter and methods](https://github.com/BuilderIO/mitosis/commit/ecce09d84ad4c864d43714a64504908e09e03646)
    - [x] compositionApi convert slots to vue getter, methods, useSlots, push vueImport
      - [x] [push `useSlots` to vueImports if hasSlotProps](https://github.com/BuilderIO/mitosis/commit/9456e37de13f9fb8f2293d96e3033b7fcef6b641)  https://vuejs.org/api/sfc-script-setup.html#useslots-useattrs 
      - [x] [dynamic `useSlots` if hasSlotProps](https://github.com/BuilderIO/mitosis/commit/a2cd68ee2a9cdca9df7bbec5aec6c54d69b13d47)
      - [x] [replaced slots getters and methods](https://github.com/BuilderIO/mitosis/commit/9c38c8cd14184e81a0e3b2955fde84f46fbcc2f6)
- Svelte
    - [x]  [convert slots to svelte template slot syntax](https://github.com/BuilderIO/mitosis/commit/a5704c302860e93648da7e1afe3f4de61aca2bd1)
    - [x] [convert slots to svelte getter and methods](https://github.com/BuilderIO/mitosis/commit/50d46efad9f5db161110b209041db854faaf0b8f)


samijaber:
>  [stumped!!!] replaced slots getters and methods

Can you elaborate on why you're stuck for that? Maybe I can help.

packages/core/src/generators/svelte/svelte.ts
```diff
@@ -176,6 +176,7 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
         },
       }),
       babelTransformCode,
+      (getterString) => replaceSlotsInString(getterString, (slotName) => `$$slots.${slotName}`),
```
samijaber, line 179: Instead of putting `replaceSlotsInString` in every part where the code is used, you can instead put it just once here:

https://github.com/BuilderIO/mitosis/blob/84645d4ee333996716c1a9ac8e8625b283ade624/packages/core/src/generators/svelte/helpers.ts#L10

We already handle `children` -> `$$slots.default`, so I think we can also include some logic there, like:

```ts
if (isSlotProperty(name)) {
  return replaceSlotsInString(name, x => `$$slots.${x}`);
} else if /**/
```

packages/core/src/generators/vue/optionsApi.ts
```diff
@@ -103,17 +105,23 @@ export function generateOptionsApiScript(
...
-    getters: false,
-    functions: true,
-  });
+  let functionsString = pipe(
+    getStateObjectStringFromComponent(component, {
+      data: false,
+      getters: false,
+      functions: true,
+    }),
+    (functionsString) => replaceSlotsInString(functionsString, (slotName) => `$slots.${slotName}`),
```
samijaber, line 123: Same thing for Vue as for Svelte, here is the function that transforms all code:

https://github.com/BuilderIO/mitosis/blob/84645d4ee333996716c1a9ac8e8625b283ade624/packages/core/src/generators/vue/vue.ts#L108-L135

So I think you can make this change once inside of the `processBinding` function that's used in this plugin:

https://github.com/BuilderIO/mitosis/blob/84645d4ee333996716c1a9ac8e8625b283ade624/packages/core/src/generators/vue/helpers.ts#L157-L165

packages/core/src/generators/svelte/svelte.ts
kingzez: Good idea 👍

kingzez:
> > [stumped!!!] replaced slots getters and methods
> 
> Can you elaborate on why you're stuck for that? Maybe I can help.
How to resolved compositionApi about here, 
https://github.com/BuilderIO/mitosis/blob/e6b77e432fa83cccf7d7692f3a8c7b378b38a2c7/packages/core/src/generators/vue/helpers.ts#L170-L172

packages/core/src/generators/vue/helpers.ts
```diff
@@ -182,8 +181,11 @@ export const processBinding = ({
           switch (options.api) {
             // keep pointing to `props.${value}`
             case 'composition':
+              if (name === 'children' || name.startsWith('children.')) {
+                return `useSlots().default`;
+              }
               return isSlotProperty(name)
-                ? replaceSlotsInString(name, (x) => `$slots.${x}`) // here must use `$slots` instead of `slots`
+                ? replaceSlotsInString(name, (x) => `useSlots().${x}`)
```
kingzez, line 188: 1. use `useSlots().xxx`  
2. use a variable `const $slots = useSlots()`  and then `$slots.xxx`

Both ways work,  which is better? I tend to use option 1 `useSlots.xxx`

samijaber: I don't mind using option 1, it's easier to generate I think.

packages/core/src/generators/vue/blocks.ts
```diff
@@ -288,27 +288,26 @@ const NODE_MAPPERS: {
 const stringifyBinding =
   (node: MitosisNode) =>
   ([key, value]: [string, Binding]) => {
+    const bindingValue = replaceSlotsInString(value.code || '', (slotName) => `$slots.${slotName}`);
```
samijaber, line 291: I want to keep all code transforms out of the `blocks.ts` file, and keep them in the `processBinding` function.

To modify `bindings`, you can add `replaceSlotsInString` in these 2 places:

https://github.com/BuilderIO/mitosis/blob/268c8a45d899abe0a89d3dca451bd7ad3725d0d8/packages/core/src/generators/vue/vue.ts#L115-L117

https://github.com/BuilderIO/mitosis/blob/268c8a45d899abe0a89d3dca451bd7ad3725d0d8/packages/core/src/generators/vue/vue.ts#L126-L127

packages/core/src/generators/svelte/blocks.ts
```diff
@@ -151,25 +156,26 @@ const stringifyBinding =
     }
 
     const { code, arguments: cusArgs = ['event'], type } = binding;
+    const bindingCode = replaceSlotsInString(code || '', (slotName) => `$$slots.${slotName}`);
```
samijaber, line 159: I think we should be able to remove all of these changes from `svelte/blocks.ts`, because you already added the `replaceSlotsInString` logic in `stripStateAndProps` in `svelte/helpers.ts`.

packages/core/src/generators/vue/blocks.ts
kingzez: done

packages/core/src/generators/svelte/blocks.ts
kingzez: done

packages/core/src/__tests__/shared.ts
```diff
@@ -209,6 +209,17 @@ const SLOTS_TESTS: Tests = {
...
+// keep for now, remove in future
+const SLOTS_SVELTE_TESTS: Tests = {
+  ContentSlotJSX: contentSlotJsx,
+  ContentSlotHtml: contentSlotHtml,
+  SlotDefault: slotDefault,
+  // SlotJsx: slotJsx,
+  SlotHtml: slotHtml,
+  SlotNamed: slotNamed,
+  classState,
+};
```
samijaber, line 221: Can we use `SLOTS_TESTS` like the other generators?

kingzez: ```SlotJsx: slotJsx,``` snapshot format code error... so I use new constant.

samijaber: The snapshot system should allow you to store the error as a snapshot, which I prefer. that way when we fix the issue later, we see the snapshot change from an error to a success.

Let me know if that's not the case and the `test` command completely fails due to this test.

kingzez: OK, I updated

samijaber (approved):
Amazing. Thank you for this very big effort to get this PR to the finish line @kingzez 🏆 !

PR #1037, opened 2023-02-14
Refactor: React generator

samijaber:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/generators/react/generator.ts
+++ b/packages/core/src/generators/react/generator.ts
@@ -344,7 +344,9 @@ const _componentToReact = (
       : undefined;
 
   const propType = json.propsTypeRef || 'any';
-  const propsArgs = `props${options.typescript ? `:${propType}` : ''}`;
+  const componentArgs = [`props${options.typescript ? `:${propType}` : ''}`, options.forwardRef]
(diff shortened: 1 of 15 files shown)


packages/core/src/generators/react/generator.ts
```diff
@@ -358,127 +468,33 @@ const _componentToReact = (
...
+        ? ''
+        : options.stateType === 'mobx'
         ? `
-      const observed${json.name || 'MyComponent'} = observer(${json.name || 'MyComponent'});
-      export default observed${json.name || 'MyComponent'};
+      const observed${json.name} = observer(${json.name});
+      export default observed${json.name};
     `
-        : ''
+        : `export default ${json.name};`
```
kingzez, line 494: Why change to this style

```js
// origin
export default function MyComponet() {}

// now
function MyComponent() {}

export default MyComponent
```

samijaber: Because with the new form, we can more easily wrap the export with other functionality, which is something we frequently need to do. Like wrapping it in a HOC or a library.

It's better for long term flexibility

PR #1038, opened 2023-02-20
Fix HMR when using the CLI

AsyncBanana:
## Description

This changes the `clean()` function in the build script for the CLI to only remove files that were actually removed in the source rather than removing every file and replacing them all. This should fix problems with using HMR, as edited files will just be changed rather than temporarily removed.


ohkimur:
@AsyncBanana Wouldn't be much easier to make the entire `clean` step optional via the config? This way, when in dev mode, or under `--dev` flag the output will not be cleaned at all.

@samijaber What do you think about this? I'd love to make something like this work properly.

AsyncBanana:
@ohkimur Personally, I still like the automatic cleaning and would rather not have to toggle it on and off. However, if there is an issue with this change or another use case that requires cleaning to be disabled entirely, then that would likely be the best solution.

ohkimur:
@AsyncBanana It would be great to be able to configure the `build` command to only output the modified files. This will allow proper HMR support, which at the moment doesn't work correctly. For example, I have set up an environment with `StoryBook` and it works fine for a while, but then it simply doesn't work and that's because all the files have been deleted and recreated.

AsyncBanana:
@ohkimur This pull request should fix HMR. I had the same problem with Astro (Vite), but with the changes in this it now works. I would recommend trying it and letting me know if it doesn't work.

ohkimur:
@samijaber Can we merge this in? I'd like to test it via the "official" path.
@AsyncBanana Thanks a lot for fixing this.

ohkimur:
@samijaber Can we take a look at this?

samijaber (approved):
Really appreciate the PR, and thank you so much for your patience. I know it took me a while to get to it. 🙏🏽

ohkimur:
@samijaber When will you make a release with this?

PR #1049, opened 2023-03-01
fix: Sveltosis behaviour on inputs

simongeau:
The goal is to make sure inputs and textarea are binded onInput instead of onChange (which is triggered on blur) so that it is more inline with the Svelte rendering

### DIFF
--- a/packages/core/src/parsers/svelte/html/element.ts
+++ b/packages/core/src/parsers/svelte/html/element.ts
@@ -207,7 +207,8 @@ export function parseElement(json: SveltosisComponent, node: TemplateNode) {
 
           if (name !== 'ref' && name !== 'group' && name !== 'this') {
             const onChangeCode = `${binding} = event.target.value`;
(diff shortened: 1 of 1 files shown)


packages/core/src/parsers/svelte/html/element.ts
```diff
@@ -207,7 +207,8 @@ export function parseElement(json: SveltosisComponent, node: TemplateNode) {
 
           if (name !== 'ref' && name !== 'group' && name !== 'this') {
             const onChangeCode = `${binding} = event.target.value`;
-            mitosisNode.bindings['onChange'] = createSingleBinding({
+            const binding = ['input', 'textarea'].indexOf(mitosisNode.name) > -1 ? 'onInput' : 'onChange';
```
samijaber, line 210: This code is broken: we already have a `binding` variable in this function. You should change it to something else

samijaber (commented):
Make sure to run `yarn snapupdate` in `packages/core` to update the snapshot tets

PR #1050, opened 2023-03-04
fix: Sveltosis behaviour on inputs

simongeau:
The goal is to make sure inputs and textarea are binded onInput instead of onChange (which is triggered on blur) so that it is more inline with the Svelte rendering


samijaber:
@simongeau Thanks for adding this fix!

If you can follow the steps outlined here: https://github.com/BuilderIO/mitosis/blob/main/developer/README.md#developing-for-core--testing

Essentially: add tests that highlight this change, and format your PR to pass the CI checks. Then we can get this merged in 🙏🏽

samijaber:
Closing this PR due to inactivity. If you're ever able to pick this up again, feel free to create a new PR with these changes.

(Make sure to follow the PR contribution template, which explains what steps you need to take to make your PR ready for review)

PR #1061, opened 2023-03-17
Fix code output for Builder.io Button component

steve8708:


packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -79,7 +101,7 @@ export const components: CompileAwayComponentsMap = {
     return createMitosisNode({
       ...node,
       // TODO: use 'button' tag for no link, and add `all: unset` to CSS string only then
-      name: hasLink ? 'a' : node.properties.$tagName || 'span',
+      name: hasLink ? 'a' : 'button',
```
samijaber, line 104: Worth noting we have some visual bugs with using `button` along with the way Builder currently styles the text inside of it: https://builder-io.atlassian.net/browse/ENG-3275

samijaber (approved):

packages/core/src/plugins/compile-away-builder-components.ts
samijaber: The issue is not with applying `all: unset`. I'm referring to you switching from using a `span` to now using a `button` element. We have inconsistencies with styling those buttons, because text styles don't apply to them as you'd expect.

The Builder Visual editor typically uses `text-align` to style a button's contents, and that won't work here.

We'll see if the issue surfaces for people. We might be able to solve it by having a `span` nested inside of the `button`

samijaber: Thanks for the jsfiddle. Maybe what I saw was under more unique circumstances. It was a very long time ago, will need to look back and see at what breaks in the SDKs. I might be wrong.

PR #1062, opened 2023-03-17
Improve Qwik output

steve8708:
Make it so the code is more human friendly for the Builder.io code output, Mitosis fidlde, and any other codegen use case


packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -321,7 +329,13 @@ export class SrcBuilder {
...
         if (isEvent(key)) {
           key = key + '$';
-          binding = `${this.file.import(this.file.qwikModule, '$').localName}((event)=>${binding})`;
+          if (this.file.options.isJSX) {
+            binding = `(event)=>${binding}`;
+          } else {
+            binding = `${
+              this.file.import(this.file.qwikModule, '$').localName
+            }((event)=>${binding})`;
+          }
```
samijaber, line 338: Can you expand on why you don't always want to wrap event handlers with `$`?

steve8708: its not needed (redundant) and not conventional. the way in the docs that you write qwik components is

`onClick$={() => ...}`

the fact that onClick ends in a dollar tells the compiler that the function is extracted, the $ is only needed in cases where the compiler can't already know

some goes for things like `useTask$(() => ...)`. as long as the identifier ends in a $ then you don't wrap the argument with a $

steve8708: if you mean why we don't do it when using non-jsx, that's because in the non jsx version we don't have the optimizer running that does this for us (that mode is used in our qwik API right now and has no "compiler magic" and is not mean to be a conventional / human readable output)

steve8708:
@samijaber I'm slightly confused about the failing lint step. it looks like it's a component not actually used (just for your conf talk?) and I can't reproduce the issue locally (lint working fine locally 🤔)

steve8708:
nvm I fixed it, some weird difference between local and production but running prettier manually from vscode fixed

samijaber (approved):

PR #1065, opened 2023-03-19
Add api option to compile command

CKGrafico:
Vue needs the option called "api" to be able to choose between composition and options apis


samijaber (approved):
🙏🏽

CKGrafico:
@samijaber  any idea about when this will be merged? its a stopper right now to me :(

PR #1067, opened 2023-03-21
Change to make note of {} control

SupaSibs:
In Mitosis you can use {} inside of a return to use a loop or a conditional statement. But the docs only said that you can use the <Show> and <For> elements.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -164,7 +164,7 @@ export default function MyComponent() {
 
 ## Control flow
 
-Control flow in Builder is static like [Solid](https://github.com/ryansolid/solid). Instead of using freeform javascript like in React, you must use control flow components like `<Show>` and `<For>`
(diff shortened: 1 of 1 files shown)


tomByrer:
Might want to add tests to prove this works for all targets.

SupaSibs:
Conditional statements are working on all except for HTML and Liquid. However I found a bug on Webcomponents which i think can be fixed: the conditional statement only runs after the values are changed more than once. This is the same for the loops but I haven't finished testing that's why I haven't sent results on it.

SupaSibs:
If I find any more issues I can fix I might close this and open up an issue instead. Unless the {} functionallity is unintended, then I'll just close the pr.

SupaSibs:
As for for loops using {}, it doesn't work on WebComponents and HTMK

SupaSibs:
I cant read swift tho and I dont have an environment so I dont know if it works there.

SupaSibs:
And what's worth noting is that show and for dont work on Webcomps and HTML either, not just {}. Perhaps i'll make an issue on that later?

PR #1077, opened 2023-03-24
Fix #1072 Preserve whitespace in backtick string literals

### LINKED ISSUE #1072: Dedent breaks multiline strings

Yes

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAwhALaR7B4wAUYUEYAzgJRwA3gCg4cKMBjoo5NqLFwAPJ2AA%2BQQANchEmTjIIENqgA0cAEYCRYiVJlxUcANQWA3MIC%2BGj4oD0K1Xk%2Bdw8gA%3D%3D

### Expected Behaviour

All content inside a backtick multiline string should be left alone

Input:

```jsx
  return (
    <pre>{`function foo(a, b) {
  return a + b;
}`}</pre>
  );
}
```

Expected output:

```jsx

function MyComponent(props) {
  return (
    <pre>{`function foo(a, b) {
  return a + b;
}`}</pre>
  );
}

export default MyComponent;
```

### Actual Behaviour

The `dedent` library de-indents content inside the multiline strings.  In this case, it breaks the formatting of the `<pre>` content.

```jsx

function MyComponent(props) {
  return (
    <pre>{`function foo(a, b) {
return a + b;
}`}</pre>
  );
}

export default MyComponent;
```

### Additional Information

Removing the call to `dedent` easily fixes the issue.  When combined with the `prettier` option, removing `dedent` seems like an easy win.  I don't know if `dedent` is required for other reasons.

codyebberson:
## Description

As discussed in https://github.com/BuilderIO/mitosis/issues/1072 using `dedent` breaks whitespace in multiline strings.

For example, this Mitosis input:

```jsx
  return (
    <pre>{`function foo(a, b) {
  return a + b;
}`}</pre>
  );
}
```

Leads to this React/Solid output, where the whitespace is incorrect (note the missing indentation before `return`):

```jsx

function MyComponent(props) {
  return (
    <pre>{`function foo(a, b) {
return a + b;
}`}</pre>
  );
}

export default MyComponent;
```

After the fix, the whitespace is correct:

```jsx

function MyComponent(props) {
  return (
    <pre>{`function foo(a, b) {
  return a + b;
}`}</pre>
  );
}

export default MyComponent;
```

I considered removing `dedent` from generators entirely, but it is unclear if that would have unintended consequences.

Instead, I included a modified version of `dedent` that ignores newlines inside backtick string literals.

The meat of the PR is here in "mitosis/packages/core/src/helpers/dedent.ts": https://github.com/codyebberson/mitosis/blob/3dc3960ffadb384122424b6a1cd3033890d400c7/packages/core/src/helpers/dedent.ts

Let me know if you have different ideas about implementation, or if there is a better way to handle this entirely.  Ultimately, my goal is just to preserve whitespace inside a `<pre></pre>` block.


samijaber (approved):
Works for me!

samijaber:
You'll need to upgrade the snapshots first before this PR can pass :)

codyebberson:
Thanks @samijaber - snapshots updated 👍

samijaber (approved):
Looks great! just need to update branch to merge

samijaber:
cc @steve8708 are you ok completely removing `dedent`?

PR #1078, opened 2023-03-25
Use Nullish check for default props

patrick-kw-chiu:
## Description

It seems to be more make sense to use the Nullish check here, to avoid rendering the right-hand value when the `prop` is a falsy value e.g. `0`, `false` or `''`

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -429,7 +429,7 @@ const DEFAULT_VALUES = {
   color: 'blue',
 };
 export default function MyComponent(props) {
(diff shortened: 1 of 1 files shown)


samijaber:
Not every environment supports the new `??` operator, and I would rather not confuse folks who are not yet familiar with it. Keeping the docs as simple as possible is my main objective, so I'm going to close this. Thanks for making the suggestion though!

patrick-kw-chiu:
Sure, that makes sense!

PR #1095, opened 2023-04-03
fix(qwik): process loops before show; correct state merging

mhevery:
## Description

- `For` loops need to be processed before `Show`/`Hide` as `Show`/`Hide` can refer to `$index$` from the `For` loop.
- The state merging of component `state`, `props` and `For`-loop `locals` have be improved to be consistent with React implementation.

### DIFF
--- a/packages/core/src/helpers/nodes/for.ts
+++ b/packages/core/src/helpers/nodes/for.ts
@@ -7,7 +7,7 @@ export const getForArguments = (
 ): string[] => {
   return [
     node.scope.forName || 'item',
-    node.scope.indexName || '$index',
+    node.scope.indexName,
     excludeCollectionName ? undefined : node.scope.collectionName,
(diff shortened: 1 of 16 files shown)


packages/core/src/helpers/nodes/for.ts
```diff
@@ -7,7 +7,7 @@ export const getForArguments = (
 ): string[] => {
   return [
     node.scope.forName || 'item',
-    node.scope.indexName,
+    node.scope.indexName || '$index',
```
samijaber, line 10: Why is this change needed for the fixes? Is `$index` a special var name that needs to stay as-is?

mhevery: Yes, that is my understanding. 

Ohh, thinking more; maybe that is for BuilderApp only?

samijaber (approved):

PR #1101, opened 2023-04-07
fix(qwik): ensure that src-set has size

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap
@@ -589,7 +589,7 @@ export const Image = function Image(props) {
         .map((size) => {
           return updateQueryParam(webpImage, \\"width\\", size) + \\" \\" + size + \\"w\\";
         })
-        .concat([image])
+        .concat([tryAppendWidth(image)])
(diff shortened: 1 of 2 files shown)


packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap
```diff
@@ -631,6 +631,14 @@ export const Image = function Image(props) {
...
     return uri + separator + key + \\"=\\" + encodeURIComponent(value);
   }
+  function tryAppendWidth(url) {
+    const match = url.match(/[?&]width=(\\\\d+)/);
+    const width = match && match[1];
+    if (width) {
+      return url + \\" \\" + width + \\"w\\";
+    }
+    return url;
+  }
```
steve8708, line 641: one issue compared to the prior logic, int he prior world if we can't find a width we don't concat at all, so we should match that

steve8708: oh wait nvm I am wrong, you are right. approved!

steve8708 (approved):

steve8708 (approved):

PR #1102, opened 2023-04-07
Added list of all compilable languages in --help

SupaSibs:
I added some entries to the old --to / -t, as it was uncomplete. I also changed the list of --to and --from into alphabetical order because it's super annoying otherwise.

### DIFF
--- a/packages/core/src/parsers/svelte/index.ts
+++ b/packages/core/src/parsers/svelte/index.ts
@@ -71,3 +71,4 @@ export const parseSvelte = async function (
   const componentName = path.split('/').pop()?.split('.')[0] ?? 'MyComponent';
   return mapAstToMitosisJson(ast, componentName, string_, usesTypescript);
 };
+export {parseSvelte, mapAstToMitosisJson}
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -28,7 +28,6 @@ OPTIONS
                 - builder
                 - customElement
                 - html
-                - json
                 - liquid
                 - lit
                 - marko
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -23,26 +23,32 @@ EXAMPLES
 OPTIONS
 	--to=<format>, -t=<format>
 		Specify output format. <format> can be one of:
-		
-		- reactNative
-		- solid
-		- vue
-		- react
-		- template
-		- html
-		- customElement
-		- mitosis
-		- builder
-		- swift
-		- svelte
-		- liquid
-		- angular
+		- alpine
+                - angular
+                - builder
+                - customElement
+                - html
+                - json
+                - liquid
+                - lit
+                - marko
+                - mitosis
+                - preact
+                - qwik
+                - react
+                - reactNative
+                - rsc
+                - solid
+                - stencil
+                - svelte
+                - swift
+                - template
+                 - vue
 	--from=<format>, -f=<format>
-		Specify input format. <format> can be one of:
-		
-		- mitosis
+		Specify input format. <format> can be one of:	
+                - liquid
 		- builder
-		- liquid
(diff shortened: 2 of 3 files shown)


tomByrer:
Should `Sveltosis` be listed in --from, or is that too alpha?

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&inputTab=M4NwpgNgLmQ%3D&code=DwZwxgTglgDgLgPgFAAIUBsCmcUDsCGAtpigLwoDkAynJgG6YUDcSwA9ONPMqwCZR1kaYFFwwArjgBGo3gC46%2BdOMykA3gWIBfFGyEoAEpnToA9gEIUASRRh8uFBHEPRKAEqZ8YOABoUANRU%2FKlN0KF4%2FUwgUABkoAEdxcPNWNn5BIA%3D

SupaSibs:
It definitely should, but when I tried to type in -f sveltosis it didnt work idk why

SupaSibs:
It definitely should, but when I tried to type in -f sveltosis it didnt work idk why

SupaSibs:
Yeah, still not working. It's a shame though as I like svelte code more than the lite jsx code

SupaSibs:
It seems that Sveltosis isn't even an option in -f (i checked the commands folder for compile and build)

tomByrer:
> It seems that Sveltosis isn't even an option in -f (i checked the commands folder for compile and build)

Yes, I was suggesting for you to add it in ;)

I'm not sure how much help this will be, but here is the original sveltosis repo:
https://github.com/sveltosis/sveltosis/tree/main/examples/my-component-library

SupaSibs:
Oh, ok. I'll try and check the sveltosis parser in parsers folder and see if there's naything i can use.

SupaSibs:
I found out how to get the parser into the CLI, but what **is* the string_ argument in parseSvelte for? It only seems to be used in the mapAst function, but it's only used for one function. Should it be say the JSON, or the JSON parsed to the string, or a value?

SupaSibs:
Thanks, now I understood it. But I have a question, how do I build/test mitosis without NPM? I'm trying to test the new cli and core to see if they work but I don't know how...

samijaber (approved):

samijaber:
> Thanks, now I understood it. But I have a question, how do I build/test mitosis without NPM? I'm trying to test the new cli and core to see if they work but I don't know how...

You can test by running the `mitosis` command inside any of the example folders within the monorepo. For example: https://github.com/BuilderIO/mitosis/blob/26a4f4ffdaef0f0001038ed5adbb7f0669a832e8/examples/todo/package.json#L4

You can run `npm exec mitosis compile -f` in this sub-folder, and it should work correctly.

packages/core/src/parsers/svelte/index.ts
```diff
@@ -71,3 +71,4 @@ export const parseSvelte = async function (
   const componentName = path.split('/').pop()?.split('.')[0] ?? 'MyComponent';
   return mapAstToMitosisJson(ast, componentName, string_, usesTypescript);
 };
+export {parseSvelte, mapAstToMitosisJson}
```
samijaber, line 74: I think this line is incorrect and causing build failures. You are already exporting `parseSvelte`. You just need to add `export` behind the `mapAstToMitosisJson` declaration

samijaber (changes requested):

packages/core/src/parsers/svelte/index.ts
SupaSibs: Damn that totally went over my head. I'm going to make a new PR for this as the topic has gone from "fixing -t in CLI" to "adding Sveltosis to cli", so i'll have to close this one.

SupaSibs:
Closed this PR as the topic has changed from fixing -t to adding sveltosis to cli.

PR #1108, opened 2023-04-10
Feature/hook config

PengBoUESTC:
## Description

add hook config for MitosisConfig, 

`beforeBuild`: run to modify config

`afterbuild`: run after each target build end 

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/cli/src/build/build.ts
+++ b/packages/cli/src/build/build.ts
@@ -24,13 +24,14 @@ import {
   MitosisConfig,
   parseJsx,
   Target,
-  TranspilerGenerator,
   parseSvelte,
+  TargetContext,
+  OutputFiles
 } from '@builder.io/mitosis';
 import debug from 'debug';
 import { flow, pipe } from 'fp-ts/lib/function';
 import { outputFile, pathExists, readFile, remove } from 'fs-extra';
-import { kebabCase } from 'lodash';
+import { kebabCase, get } from 'lodash';
 import { fastClone } from '../helpers/fast-clone';
 import { generateContextFile } from './helpers/context';
 import { getFileExtensionForTarget } from './helpers/extensions';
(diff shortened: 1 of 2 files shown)


packages/cli/src/build/build.ts
```diff
@@ -216,6 +211,22 @@ const buildAndOutputNonComponentFiles = async (targetContext: TargetContextWithC
   return await outputNonComponentFiles({ ...targetContext, files });
 };
 
+function runHook(hook: string, config?: MitosisConfig) {
+  const noop = () => {};
+  const hookFn = get(config, `hooks.${hook}`);
```
samijaber, line 216: I would type this better, and remove the `lodash` dependency (I find them too bulky and the string can have any number of typos). Something like:

```suggestion
function runHook(hook: keyof MitosisConfig['hooks'], config?: MitosisConfig) {
  const noop = () => {};
  const hookFn = config?.hooks?.[hook]
```

packages/cli/src/build/build.ts
```diff
@@ -216,6 +211,22 @@ const buildAndOutputNonComponentFiles = async (targetContext: TargetContextWithC
   return await outputNonComponentFiles({ ...targetContext, files });
 };
 
+function runHook(hook: string, config?: MitosisConfig) {
```
samijaber, line 214: config should be required:

```suggestion
function runHook(hook: string, config: MitosisConfig) {
```

Then you'll realize you made a mistake and forgot to provide `config` when you called `runHook` in the code below

samijaber (changes requested):

packages/cli/src/build/build.ts
PengBoUESTC: I have modified the plugin scheme

packages/cli/src/build/build.ts
PengBoUESTC: yes , U are right =_=

PengBoUESTC:
@samijaber

PR #1115, opened 2023-04-12
fix: no config file found for Mitosis

ohkimur:
## Description

Add a short description of:
When passing a custom config file, either with `js` or `cjs` extension, to the CLI, the "No config file found for Mitosis" was shown.

This is a fix, that preserves the original approach with the additional edge case of a custom config file handled properly.

NOTE: The imports were automatically organized by `vs-code` with `organizeImports` function.


samijaber (approved):

samijaber:
Thank you for contributing this @ohkimur 🙏🏽 !

PR #1118, opened 2023-04-13
Added sveltosis to compile.t

SupaSibs:
Added svelte to the compiler so you can compile from the svelte language.


samijaber (approved):

samijaber:
You need to lint your PR with `yarn run lint`

samijaber:
@SupaSibs sorry I had misspoken: it was `yarn run fmt`. I went ahead and made the change

SupaSibs:
Ohhh, ok, that makes more sense. I've never really used any of the mitosis package.json commands so i didn't know enough about that, Thanks!

Issue #526, opened 2022-06-28
BUG: Mitosis does camelCase HTML attributes in React


samijaber:
**Scope**
- [ ] This impacts ALL Mitosis outputs
- [x] This only impacts specific generators/outputs (please list them here): React

**Describe the bug**
Attributes like `srcset` are camelCased in React (possibly other frameworks? Not sure). However, Mitosis does not account for that

**To Reproduce**

A link to a https://mitosis.builder.io/ fiddle containing the bug: [link](https://mitosis.builder.io/?outputTab=react&code=import+%7B+useState+%7D+from+%22%40builder.io%2Fmitosis%22%3B%0A%0Aexport+default+function+MyComponent%28props%29+%7B%0A++const+%5Bname%2C+setName%5D+%3D+useState%28%22Steve%22%29%3B%0A%0A++return+%28%0A++++%3Cdiv%3E%0A++++++%3Cimg+srcset%3D%22http%3A%2F%2Fexample.com%22+%2F%3E++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D)

**Expected behavior**
`scrset` should become `srcSet` in the React output. Same goes for all HTML attributes that are multi-word: https://reactjs.org/docs/dom-elements.html

**Additional context**
- I am not sure what the comprehensive list of camelCased DOM elements is, but we can probably look at React's JSX types to determine what to do 
- I am also not sure if other frameworks do this camelCasing. We might need to do this elsewhere too

samijaber:
Technical information:

We already have a list of all HTML elements: https://github.com/BuilderIO/mitosis/blob/b8d083ca78bdaec332745d2a0f8c0890335072c4/packages/core/src/constants/html_tags.ts#L1

We want to go to the `blockToReact` function, and identify blocks that are native HTML elements, and rename any `json.bindings` or `json.properties` keys that should be camelCased.

https://github.com/BuilderIO/mitosis/blob/b8d083ca78bdaec332745d2a0f8c0890335072c4/packages/core/src/generators/react.ts#L164-L167

https://github.com/BuilderIO/mitosis/blob/b8d083ca78bdaec332745d2a0f8c0890335072c4/packages/core/src/generators/react.ts#L184-L186

In both of these for-loops, we want to add a check that looks like this:

```js
const isHtmlElement = VALID_HTML_TAGS.includes(json.name);

const checkKeyShouldBeCamelCased = (key: string) => {
  // ... TO-DO:
}

[...]

PR #1122, opened 2023-04-14
Map camle cased attributes to lower case

moaoa:
## This is a draft pr to try and solve #526

what I added :
- isFirstLetterLowerCase util, 
- mapCamleCasedAttribues to lowerCase attributes (srcSet => srcset)

I followed the steps mentioned in the issue but my solution is not working out. Can anyone help me?

### DIFF
--- a/packages/core/src/__tests__/camle-case.test.ts
+++ b/packages/core/src/__tests__/camle-case.test.ts
@@ -1,0 +1,11 @@
+import MyComponent from './data/map-camle-cased-attributes.raw?raw'
+import { parseJsx } from '../parsers/jsx';
+import { componentToSvelte } from '../generators/svelte';
+
+describe('map camle cased', () => {
+    test('map camle cased', () => {
+        const component = parseJsx(MyComponent);
+        const svelteComponent = componentToSvelte()({component})
+        expect(svelteComponent).toMatchSnapshot();
+    })
+})
--- a/packages/core/src/__tests__/data/map-camle-cased-attributes.raw.tsx
+++ b/packages/core/src/__tests__/data/map-camle-cased-attributes.raw.tsx
@@ -1,0 +1,12 @@
+import { useState } from "@builder.io/mitosis";
+import React from '@builder.io/react'
+
+export default function MyComponent(props) {
+  const [name, setName] = useState("Steve");
+
+  return (
+    <div>
+      <img srcSet="http://example.com" />    
+    </div>
+  );
+}
--- a/packages/core/src/generators/svelte/svelte.ts
+++ b/packages/core/src/generators/svelte/svelte.ts
@@ -33,6 +33,9 @@ import { stripStateAndProps } from './helpers';
 import { ToSvelteOptions } from './types';
 import { blockToSvelte } from './blocks';
 import { stripGetter } from '../../helpers/patterns';
+import { isFirstLetterLowerCase } from '../../helpers/is-first-letter-lower-case'
+import { mapCamelCasedHtmlAttributes } from 'src/helpers/map-camel-cased-html-attributes';
+import { MitosisNode } from 'src/types/mitosis-node';
 
 const getContextCode = (json: MitosisComponent) => {
   const contextGetters = json.context.get;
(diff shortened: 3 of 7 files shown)


packages/core/src/generators/svelte/svelte.ts
```diff
@@ -146,7 +149,9 @@ export const componentToSvelte: TranspilerGenerator<ToSvelteOptions> =
       .filter((x) => !props.includes(x));
 
     json = runPostJsonPlugins(json, options.plugins);
-
+    //@ts-ignore
+    mapCamelCasedHtmlAttributes(json)
```
samijaber, line 153: Ah, I missed 1 detail: you are not meant to run this on the `MitosisComponent` json. Instead, you want to run it on each node. You can reuse this helper to do so:

https://github.com/BuilderIO/mitosis/blob/df756ff8a690da079a56be098e5e78faf83d3677/packages/core/src/helpers/traverse-nodes.ts#L6

In fact, `mapCamelCasedHtmlAttributes` should be a _plugin_, as that's the preferred abstraction for functionality that we want to share across generators. You can look at this plugin which reuses that same helper:

https://github.com/BuilderIO/mitosis/blob/df756ff8a690da079a56be098e5e78faf83d3677/packages/core/src/helpers/plugins/process-code.ts#L61

Your plugin should look like:

```ts
export const CAMEL_CASE_PLUGIN: Plugin = () => ({
  json: {
    post: (json) => {
      traverseNodes(json, (node) => {
        mapCamelCasedHtmlAttributes(node);
      });
    },
  },
})
```

and then add `CAMEL_CASE_PLUGIN` to `DEFAULT_OPTIONS.plugins` for each generator where this applies

packages/core/src/__tests__/camle-case.test.ts
```diff
@@ -0,0 +1,11 @@
...
+import { parseJsx } from '../parsers/jsx';
+import { componentToSvelte } from '../generators/svelte';
+
+describe('map camle cased', () => {
+    test('map camle cased', () => {
+        const component = parseJsx(MyComponent);
+        const svelteComponent = componentToSvelte()({component})
+        expect(svelteComponent).toMatchSnapshot();
+    })
+})
```
samijaber, line 11: instead of adding a single test, add it to the `shared.ts` like here:

https://github.com/BuilderIO/mitosis/blob/df756ff8a690da079a56be098e5e78faf83d3677/packages/core/src/__tests__/shared.ts#L25

and add it to the `BASIC_TESTS` array, which will run it for every generator.

packages/core/src/__tests__/data/map-camle-cased-attributes.raw.tsx
```diff
@@ -0,0 +1,12 @@
+import { useState } from "@builder.io/mitosis";
```
samijaber, line 1: you mispelled `camel` as `camle`

samijaber (changes requested):

packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap.rej
```diff
@@ -0,0 +1,143 @@
+diff a/packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap b/packages/core/src/__tests__/__snapshots__/qwik.test.ts.snap	(rejected hunks)
```
samijaber, line 1: this file needs to be removed

packages/core/src/generators/vue/vue.ts
```diff
@@ -276,11 +277,31 @@ const componentToVue: TranspilerGenerator<Partial<ToVueOptions>> =
...
+  api: 'composition',
+  vueVersion: 3,
+  prettier: true,
+  plugins: [CAMEL_CASE_PLUGIN],
+};
+
+export const componentToVue3 = (vueOptions?: VueOptsWithoutVersion) =>{
+  const options = mergeOptions<ToVueOptions>(DEFAULT_OPTIONS_VUE_3, { ...vueOptions, vueVersion: 3 })
+  return componentToVue(options);
+}
```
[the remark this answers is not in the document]
moaoa: yes i used that one

Issue #960, opened 2022-12-19
Angular standalone output still generates NgModule


R-Bower:
### I am interested in helping provide a fix!

No

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

N/A

### Expected Behaviour

The `standalone` option for the Angular output target should prevent an NgModule from being generated.

### Actual Behaviour

When the `standalone` option is provided to the Angular output target, the Angular code still generates an NgModule.  

As per the [Angular docs](https://angular.io/guide/standalone-components#the-standalone-flag-and-component-imports):
> Components, directives, and pipes can now be marked as standalone: true. Angular classes marked as standalone do not need to be declared in an [NgModule](https://angular.io/api/core/NgModule) (the Angular compiler will report an error if you try).

samijaber:
Thanks for filing. We want to only run the following line if `options.standalone !== true`: https://github.com/BuilderIO/mitosis/blob/637ecbddf6d8082c1fbde34a8ba0c245a32874b7/packages/core/src/generators/angular.ts#L571

harshilparmar:
I am on it!!

PR #1128, opened 2023-04-15
fix(angular): prevent NgModule generation in standalone components

alexonaci:
## Description

- Standalone output option prevents the generation of an NgModule. Fixes https://github.com/BuilderIO/mitosis/issues/960
- Added basic tests and generated snapshot

For testing I added a check on the output of the generator if it includes (or not) an NgModule or component metadata with imports, please let me know if this can be improved.


packages/core/src/__tests__/angular.standalone-component.test.ts
```diff
@@ -0,0 +1,86 @@
+import { parseJsx } from '..';
```
samijaber, line 1: You can safely remove this test, as the angular tests already compare `standalone` being true vs false:

https://github.com/BuilderIO/mitosis/blob/1f80185077cc354316dce7be04b0b40fd13b6856/packages/core/src/__tests__/angular.test.ts#L1-L13

alexonaci: now I got it, thank you! Changes added

samijaber (approved):
Thanks for contributing this! 🙏🏽

Issue #1035, opened 2023-02-13
Fiddle: autocomplete styles broken in light mode


samijaber:
Works fine in dark mode, but light mode is rough:

<img width="517" alt="CleanShot 2023-02-13 at 16 56 54@2x" src="https://user-images.githubusercontent.com/1393142/218572881-25e5a5f5-f4bf-4d28-b16d-7d64fcca154d.png">

PR #1129, opened 2023-04-17
css fix

Lokprakash-babu:
## CSS fix for autocomplete styles in light mode [issue](https://github.com/BuilderIO/mitosis/issues/1035)

Add a short description of:
- Changed the text color based on the theme. 
- Previously, text color was targeted globally through a>span. Seems like Monaco autocomplete uses same DOM structure for texts inside autocomplete. 
- With old CSS structure for styling the code, it might affect globally wherever we are using 'a' tags. Since, the text is white color using it with white background might not be intuitive. Hence moved the css which targets color to the header link.


ohkimur (approved):

samijaber:
Awesome! Thank you for taking a stab at this 🙏🏽 


I think there is still 1 place missing where the CSS needs to be changed: the color of some of the text is still `black` even though it should be `white`. 

See this  screenshot: 


CleanShot 2023-04-18 at 11 05 42@2x

Lokprakash-babu:
@samijaber Please check now. Fixed this in the latest commit.

samijaber (approved):
Wonderful, thank you for fixing this!

samijaber:
@Lokprakash-babu you need to lint your PR with `yarn run lint` at the root of the repo

PR #1133, opened 2023-04-18
fix(qwik): fix missing class attribute

mhevery:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/parsers/builder/builder.ts
+++ b/packages/core/src/parsers/builder/builder.ts
@@ -383,6 +383,8 @@ const componentMappers: {
       }),
     };
     const properties = { ...block.properties };
+    if (block.id) properties['block-id'] = block.id;
+    if (block.class) properties['class'] = block.class;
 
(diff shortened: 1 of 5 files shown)


packages/core/src/parsers/builder/builder.ts
```diff
@@ -383,6 +383,8 @@ const componentMappers: {
       }),
     };
     const properties = { ...block.properties };
+    if (block.id) properties['block-id'] = block.id;
+    if (block.class) properties['class'] = block.class;
```
samijaber, line 387: worth pointing out that this 2nd line did not add anything to snapshots. Is it redundant? Or do snapshot tests not cover the usecase for it?

mhevery: Yes that is the important bit. Will update the tests.

samijaber (approved):

PR #1135, opened 2023-04-19
feat: enhance build command

PengBoUESTC:
## Description

enhance build command ;
so we can use command like this: 
`mitosis build --config xxx --files 'src/components/badge/*.tsx'"`


PengBoUESTC:
@samijaber

PR #1144, opened 2023-04-23
global typescript option

### LINKED ISSUE #1114: Feature Request: Global TypeScript option in Mitosis Config
Currently you have to define `typescript: true`  inside `options` per target, but it would be nice if the config supported a global `typescript: true` flag that would apply to all relevant targets.

moaoa:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

closes #1114


samijaber:
Please lint your PR: https://github.com/BuilderIO/mitosis/tree/main/developer#pre-submit

moaoa:
Done 👍

packages/core/src/types/config.ts
```diff
@@ -15,6 +15,7 @@ export type GeneratorOptions = {
 };
 
 export type MitosisConfig = {
+  commonOptions: { typescript: boolean };
```
samijaber, line 18: We want all of this to be optional

```suggestion
  commonOptions?: { typescript?: boolean };
```

ohkimur:
@moaoa @samijaber I don't think this works with tsx files for some reason? I'm I missing something?

ohkimur:
@samijaber Can you create an issue for this particular one, please? I'll create issues when I'll find other bugs in the future.

Issue #1164, opened 2023-05-07
Add Preact Signals Option


crutchcorn:
It would be nice to add a [Preact Signals](https://preactjs.com/guide/v10/signals/) state implementation for the Preact generator

Issue #1163, opened 2023-05-07
createContext imports from React in Preact Generator


crutchcorn:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

N/A

### Expected Behaviour

Given this code in a Preact generator:

```jsx
import { createContext } from '@builder.io/mitosis';

export default createContext({
  foo: 'bar',
  get fooUpperCase() {
    return this.foo.toUpperCase();
  },
  someMethod() {
    return this.fooUpperCase.toLowercase();
  },
  content: null,
  context: {} as any,
  state: {},
});
```

It will output:

```jsx
import { createContext } from "react";

export default createContext<any>({
  foo: "bar",
  get fooUpperCase() {
    return this.foo.toUpperCase();
  },
  someMethod() {
    return this.fooUpperCase.toLowercase();
  },
  content: null,
  context: {},
  state: {},
});
```

### Actual Behaviour

It should import from `preact`  instead

Issue #1161, opened 2023-05-07
Preact Exposes React Options that Don't Work


crutchcorn:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

N/A

### Expected Behaviour

Because Preact reuses React's `ToReactProps`, it allows you to select state libraries, `type`, and even style libraries that are not supported in Preact

### Actual Behaviour

Preact only exposes the properties that Preact only can support

Issue #1160, opened 2023-05-07
ForwardRef is not properly supported in Preact


crutchcorn:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=A4JwpghgxgLkA%3D%3D%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAwhALaR7B4wAUYUEYAzgJRwA3gCg4cKMBjoo5NqLFwAPAmIA3cSgC8gztx4A6GMB4wAvgD4AKsZiKA9CtXn5fANzDTQA%3D

### Expected Behaviour

Using a `ref` inside of a Preact component adds `forwardRef`, per its shared rendering code with React.

But `forwardRef` isn't needed or available in Preact unless you're using the compat mode, which we're hoping to avoid.

### Actual Behaviour

Ideally, we should simply pass the `ref` to the component without changing anything else.

PR #1165, opened 2023-05-07
Split Preact Rendering to Dedicated Renderer

crutchcorn:
> First, let me start by apologizing: I generally try to avoid opening large PRs that shift code and functionality after discussing them with the authors first. This PR started with me playing with the codebase to see how viable it was and ended up with me playing with the codebase to learn it.
> 
> Should this work not align with the Mitosis project goals, I'd understand and wouldn't hold it against the project to throw away this work.

BLUF: **This PR splits Preact into its own generator and fixes two bugs along the way** (1163, 1160)

After looking through Mitosis' Preact support, I found a few issues, namely:

- https://github.com/BuilderIO/mitosis/issues/1163
- https://github.com/BuilderIO/mitosis/issues/1160
- https://github.com/BuilderIO/mitosis/issues/1155 (Fixed in another PR)

In addition to the bugs, there appear to be properties that are unintentionally added to the Preact generator options:

- https://github.com/BuilderIO/mitosis/issues/1161

And an opportunity to add a few options/properties that primarily/only apply to Preact over React:

- https://github.com/BuilderIO/mitosis/issues/1158
- https://github.com/BuilderIO/mitosis/issues/1164

After thinking through all of these issues, I realized that Preact should not be a strict derivative of React's generator but rather own its own generation code and simply share much of its logic from React.

This PR:

- Splits up the generator code from Preact
- Fixes an issue with createContext not importing the right path
- Fixes issues with properties that don't really exist in Preact (but do for React)

### DIFF
--- a/packages/cli/src/build/helpers/context.ts
+++ b/packages/cli/src/build/helpers/context.ts
@@ -8,6 +8,7 @@ import {
   MitosisConfig,
   parseContext,
   Target,
+  contextToPreact,
 } from '@builder.io/mitosis';
 import { readFile } from 'fs-extra';
 import { upperFirst, camelCase, last } from 'lodash';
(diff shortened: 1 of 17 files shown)


crutchcorn:
Closing, as prior discussions with @samijaber on Discord indicate this is not the route we want to go in

KevinEdry:
@crutchcorn I am trying to run Storybook with the generated Preact components and having some trouble because of the pragma, how did you solve this issue?

Issue #1167, opened 2023-05-08
Default Slot content extract defaultProps will lose variable scope


kingzez:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [X] React
- [X] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKNroDRyYA2E8AvnAGZQQhwDkAAgEbLBEAm6UAdMBAHoQwHKmCp6AKEnoAHpFhwulAIbIi8SsgB2AYxj9tcAHKlgu9ACEVUABRgaYVAEpEkuHF0RtqeL5Uw6HAAvCgY2Li2CNIeHgAWKtocROgAwiQYAFxwtugAbujaMAD82QCyEGjoAKIFRa7BAHxusbEOEE483ukQGMU8ts7usWR4w2RDw1DoMMhQRrbDHgA8HMB5nkQqqKjBCAAGqMiUlMCyALTm3vtkjUuxy8SkcNoqIOjBAERHJ2efd61Wqt1nBukRzABrPa5OowBrNfyBHgJJIpHoYGGFOG3e6Aza9dC4h4CNZ5AGA5YCJ4wckrEnrclDMiSIA%3D

### Expected Behaviour

Default Slot content extract defaultProps will lose variable scope, `handleClose  is not defined`, it should not extract defaultProps.

Mitosis JSX
```tsx
import { useStore, Slot } from '@builder.io/mitosis'

export default function NoticeBar(props) {
  const state = useStore({

    handleClose: (event?: MouseEvent) => {
      props.onClose?.()
    },
  })

  return (
    <div class={`suffix-icon`}>
      <Slot name="suffix">
        <div onClick={(event) => state.handleClose(event)}>
          close
        </div>
      </Slot>
    </div>
  )
}

```

React
```tsx
import * as React from 'react'

function NoticeBar(props) {
  function handleClose(event) {
    props.onClose?.()
  }

  return (
    <div className={`suffix-icon`}>
      {props.slotSuffix ?
        props.slotSuffix : (
        <div onClick={(event) => handleClose(event)}>close</div>
      )}
    </div>
  )
}

export default NoticeBar
```

[...]

PR #1168, opened 2023-05-08
fix: Slot default content no extract defaultProps

kingzez:
## Description

fixed https://github.com/BuilderIO/mitosis/issues/1167


samijaber (approved):
👍🏽

PR #1170, opened 2023-05-08
fix: remove incorrect forwardRef from Preact

crutchcorn:
## Description

This PR removes the `forwardRef` functionality in Preact, where it's not supported [without a `preact/compat` package import](https://preactjs.com/guide/v10/switching-to-preact/#forwardref).

Fixes #1160

PR #1172, opened 2023-05-09
React renderer

crutchcorn:
> This PR builds on top of #1170, please merge that first

## Description

This PR ensures that `forwardRef`s only get passed in React when:
     - `props.ref` is passed
     - `useMetadata` has a passed name that matches a ref


samijaber:
can you handle the conflicts in this branch and merge main so we see a cleaner diff?

packages/core/src/helpers/get-props.ts
```diff
@@ -14,7 +14,8 @@ const prohibitedKeywordRE = new RegExp(
       'extends,finally,continue,debugger,function,arguments,typeof,void,' +
       // both below conditions added based on https://github.com/vuejs/vue/blob/e80cd09fff570df57d608f8f5aaccee6d7f31917/src/core/instance/state.ts#L89-L96
       // below line added from https://github.com/vuejs/vue/blob/8880b55d52f8d873f79ef67436217c8752cddef5/src/shared/util.ts#L130
-      'key,ref,slot,slot-scope,is,' +
```
samijaber, line 17: This logic is used by many generators, and we need to forbid these reserved words for Vue as per the comment above: https://github.com/vuejs/vue/blob/8880b55d52f8d873f79ef67436217c8752cddef5/src/shared/util.ts#L130

crutchcorn: Alternative method:

Can we break this into two distinct regex keywords so that they only apply to the specific generator?

samijaber: Yeah.

I'd do that by adding a `target` argument to `getProps`, and toggle parts of the regex based on the need. Something like:

```ts
target === 'react' ? '' : 'ref'
```

samijaber: I just realized this function is never used by the React generatore, so I'm not sure your changes are doing anything.

packages/core/src/generators/react/refs.ts
```diff
@@ -0,0 +1,30 @@
...
+export function getReactPropsRef(
+  json: MitosisComponent,
+  shouldRemove?: boolean,
+): [string, boolean] {
+  let has = false;
+  let prop = '';
+  traverse(json).forEach(function (item) {
+    if (isMitosisNode(item)) {
+      const binding = item.bindings.ref;
+      const regexp = /(.+)?props\.(ref)( |\)|;|\()?$/;
```
samijaber, line 14: We want to avoid adding new regexes to the codebase moving forward and rely instead on Babel transformers. Regexes are very error-prone and brittle, and hard to understand. I can tell you copy/pasted similar helpers used by other generators but this stuff causes lots of bugs.

To use Babel, you can rely on https://astexplorer.net/ to figure out how the AST is structured, or look at some examples in the codebase:

https://github.com/BuilderIO/mitosis/blob/d5feb3047d04671a4d41a7e6bc7dc6131690ac28/packages/core/src/helpers/map-refs.ts#L13-L100



I made a quick example that should do what you want here: https://astexplorer.net/#/gist/5b7fcfe92b39eb399aea7236758befa2/latest


You can reuse `babelTransformExpression` to run the transform on the code

PR #1186, opened 2023-05-16
fix: useStyle for react output when using emotion or styled-components

### LINKED ISSUE #1124: `useStyle` hook doesn't work for some targets

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [X] React
- [X] React-Native
- [X] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN0A0KG2AngDbpwC+cAZlBCHAEQACARssOQCbpQAdMAgB6EMBgRUwVCwDcAKEXoAHpFhx+dHMnLw6yAHYBjGCKNwAsqQDCTSEfRGYACjCMwqAJSJFcOBMII1R4AG0jHBBCOAwYADko9ABdOABeYixcfFcWbHQAN3QWbyV/TLJKVwADcoCOCF5SPwDWwIhyaAAuOAByDnJkdF66mnLq72UAqHQYZChLV1GAHl5gAoA+UYDl4CMwZBht1pNUVDSEBGO2oM6oHpYZ3hYCa5pqN4KcQfQLyOiPm02sFbAALHBGADmvwQrkKzhgvjSG1is0S0ThRRcglwUGhMEEXx+3kBQNEWyBcAAEuhyJ0AIRwACSgQhcCgxjgezgACV0DgzEQAGpDIiYDrAXhEaBwAAywAAjtxePSVqI1ptyqVFNQgA

### Expected Behaviour

The `useStyle` hook should work for all targets. However, that's not the case for `react`, particularly with `styled components`, or `emotion`. Also, the same problem appears for react native, and solid targets.

When using the `useStyle` hook, the additional CSS must be included in the output target. When compiling to `react` with `styled-components` or `emotion` we could use something like global styles or similar.

### Actual Behaviour

The `useStyle` hook is ignored for `react` target when using `styled-components` or emotion. It also doesn't work with `react-native` and `solid` targets.

### Additional Information

The `useStyle` hook is also ignored when the output is Mitosis.

ohkimur:
This closes #1124

### DIFF
--- a/packages/core/src/helpers/styles/collect-css.ts
+++ b/packages/core/src/helpers/styles/collect-css.ts
@@ -80,8 +80,9 @@ const collectStyles = (
 export const collectCss = (json: MitosisComponent, options: CollectStyleOptions = {}): string => {
   const styles = collectStyles(json, options);
   // TODO create and use a root selector
-  let css = classStyleMapToCss(styles);
-  css += json.style?.length ? `\n${json.style}` : '';
+  let css = '';
+  css += !!json.style?.length ? `${json.style}\n` : '';
+  css += classStyleMapToCss(styles);
   return css;
 };
 
--- a/packages/core/src/generators/react/generator.ts
+++ b/packages/core/src/generators/react/generator.ts
@@ -320,11 +320,17 @@ const _componentToReact = (
     reactLibImports.add('useEffect');
   }
 
+  const hasCustomStyles = !!json.style?.length;
+  const shouldInjectCustomStyles =
+    hasCustomStyles &&
+    (options.stylesType === 'styled-components' || options.stylesType === 'emotion');
+
   const wrap =
     wrapInFragment(json) ||
     isRootTextNode(json) ||
     (componentHasStyles &&
       (options.stylesType === 'styled-jsx' || options.stylesType === 'style-tag')) ||
+    shouldInjectCustomStyles ||
     isRootSpecialNode(json);
 
   const [hasStateArgument, refsString] = getRefsString(json, allRefs, options);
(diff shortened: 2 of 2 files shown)


ohkimur:
@samijaber I researched a bit about how SolidJS output works and adapted the same solution as with React output.

ohkimur:
@samijaber Do you have any ideas why the e2e deps install is failing?

packages/core/src/helpers/styles/collect-css.ts
```diff
@@ -80,8 +80,9 @@ const collectStyles = (
 export const collectCss = (json: MitosisComponent, options: CollectStyleOptions = {}): string => {
   const styles = collectStyles(json, options);
   // TODO create and use a root selector
-  let css = classStyleMapToCss(styles);
-  css += json.style?.length ? `\n${json.style}` : '';
+  let css = '';
+  css += !!json.style?.length ? `${json.style}\n` : '';
+  css += classStyleMapToCss(styles);
```
samijaber, line 85: Is there a particular reason why the order was changed here? You moved `classStyleMapToCss` from first to last.

ohkimur: @samijaber As shown [here](https://css-tricks.com/the-order-of-css-classes-in-html-doesnt-matter/), the order or classes definition does matter. The last defined classes have higher priority, so I thought if someone is writing custom styles, and they apply other classes to an element, the custom styles should have lower precedence. However, I'm agnostic about this, if you think the previous approach was better, I could just revert it how it was.

samijaber (approved):

PR #1188, opened 2023-05-17
Sveltosis: Treat arrow functions as regular function declarations

### LINKED ISSUE #1182: Fat arrow functions and regular functions handled differently

Yes

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&inputTab=M4NwpgNgLmQ%3D&code=DwZwxgTglgDgLgPgFAAIVgPYDsR3QGyjAGsAJAQywBN8BTCARhQF4UAKWgN1qzgEoWCdNhAY6AOnwYA5mwBEVDClEBbWnAAWULNJQM5fVCiMAzAK5YwcKNgJEylGvQBMHbrwEBvYTjG1JMvKKyhhqmtq6zgYA3CgAvkjAAPTg0PDIiQBGZnBwttgAXGCEJMyexfYU1HSMcQgAQjl5WHrJ2bnYyMDtzSiFFaXlJQ7VLnWNHS3ObU2dQA%3D

### Expected Behaviour

Both should be handled equally.

### Actual Behaviour

Fat arrow functions are seen as data in mitosis.

### Additional Information

_No response_

dritter:
Hi there!

This is a clumsy approach to treat arrow functions as regular function declarations.

I usually write my functions as arrow functions like so (as a matter of personal taste, I find it more readable):
```Svelte
<script>
const myClickHandler = (event) => {
  console.log('do something')
}
</script>

<button on:click={myClickHandler}>Click Me</button>
```

Before this PR the generated code looked like this:
```Svelte
<script>
  let myClickHandler = null;
</script>

<button
  on:click={(event) => {
    myClickHandler(event);
  }}>Click Me</button
>
```

So the ArrowFunction was detected as VariableDeclaration, which is not wrong, but the generated code wouldn't work anyway.

With this PR, the generated code looks like this:
```Svelte
<script>
  function myClickHandler(event) {
    console.log("do something");
  }
</script>

<button
  on:click={(event) => {
    myClickHandler(event);
  }}>Click Me</button
>
```

This is not perfect, but I created this PR to start the discussion. :)

### Caveats

Because this is done by converting arrow functions into function declarations it comes with some drawbacks:
1. In the regular function `this`, `arguments` and `super` [have different bindings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), so the generated code might behave differently than intended. It would be much better to generate an ArrowFunction instead.
2. The code wouldn't work if the ArrowFunction does not contain a BlockStatement.
3. The code wouldn't work if there is not left hand statement (VariableDeclaration).

Fixes #1182


dritter:
Okay.. The current version of [fatArrow.svelte](https://github.com/BuilderIO/mitosis/blob/4cea3707dd8a866f65e700469f15cfcec2457037/packages/core/src/__tests__/syntax/svelte/fat-arrow.raw.svelte) as input generates the following svelte file:
```svelte
<script>
  clickHandler1 = (event) => {
    console.log("do something 1");
  };

  function clickHandler2(event) {
    console.log("do something 2");
  }
</script>

<button
  on:click={(event) => {
    clickHandler1(event);
  }}>Button 1</button
>
<button
  on:click={(event) => {
    clickHandler2(event);
  }}>Button 2</button
>
```
(note the missing `const`).

In vue this looks worse:
```vue
<template>
  <div>
    <button @click="clickHandler1($event)">Button 1</button>
    <button @click="clickHandler2($event)">Button 2</button>
  </div>
</template>

<script>
import { defineComponent } from "vue";

export default defineComponent({
  name: "fat-arrow",

  methods: {
    clickHandler1: (this.clickHandler1 = (event) => {
      console.log("do something 1");
    }),
    clickHandler2: function clickHandler2(event) {
      console.log("do something 2");
    },
  },
});
</script>
```

PR #1194, opened 2023-05-23
Fix windows attempt 3.0

crutchcorn:
## Description

After #1192, I was still having issues with trying to get the Fiddle building on Windows. This PR gets my Windows machine to be able to build the Fiddle.


crutchcorn:
Huh... I rebooted my machine and wiped `node_modules` and it turns out Sophos was killing the `yarn` process early, sighting "disk encryption". This isn't the first time this has happened either and I don't know how to debug.

Once I disabled that, `main` works fine. Closing until something more comes up.

Issue #524, opened 2022-06-28
Tooling: cannot use path aliases in imports


samijaber:
Even though the typescript server accepts them just fine in my VSCode, the build fails if we use a path alias.

https://github.com/BuilderIO/mitosis/blob/b8d083ca78bdaec332745d2a0f8c0890335072c4/packages/core/src/generators/vue.ts#L3-L16

These all have to be relative paths. If I were to write:

```ts
 import { collectCss } from 'src/helpers/collect-styles'; 
```

Then the build crashes, even though it seems to be properly setup in `tsconfig.json`. It's probably some minor configuration issue! This is a great good-first-issue for newcomers

samijaber:
I think this impacts both `core` and our `cli` projects, but I am not entirely sure.

decadef20:
I'll look into this.

decadef20:
I find a related issue [tsc - doesn't compile alias paths](https://stackoverflow.com/questions/59179787/tsc-doesnt-compile-alias-paths).
Because tsc don't compile alias path, if we set an alias in `core`,  other projects in `packages` and `example` which use `core` also need to set alias.

If we set it in `example` like this:
```js
"paths": {
      "@builder.io/mitosis": ["../../packages/core/src"],
      "core/*": ["../../packages/core/src/*"]
 }
```
which compiles failed.

samijaber:
@hseoy feel free to take a stab at it!

PR #1196, opened 2023-05-24
Tooling: Install and apply tsc-alias package for path alias in imports

hseoy:
## Description

I found the [tsc-alias](https://www.npmjs.com/package/tsc-alias) package to resolve the #524 issue, and I made modifications to the cli and core packages to enable the usage of path aliases in import statements.

I have initially added a path alias called 'src' and 'core', but I also welcome other suggestions. 😊

To test the application of path aliases, I made modifications to the following two files as examples. I believe it would be beneficial to gradually apply path aliases in the future

- packages/cli/src/commands/compile.ts
- packages/core/src/generators/vue/vue.ts

#### Build Result: packages/cli/src/commands/compile.ts

screenshot for result 1

#### Build Result: packages/core/src/generators/vue/vue.ts

screenshot for result 2


hseoy:
I'll figure out the cause of the failure 🥲

samijaber:
Closed in favor of #1291. I appreciate the effort you put into trying to put this together!

PR #1208, opened 2023-06-21
fix(angular): import paths for child components

### LINKED ISSUE #1205: bug: Angular import paths are incorrect with child components

Yes

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=IYOw5grgNsBOQ%3D%3D%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFATQkWIjgBhABbMWCipAB26TfGJkKcAOQceAYxWsjkiegAes%2BGxI5kTeCWSazMbprgAsgCe6uAQ2roAFGDkYKgAlIgScHBm4ajwANqaOCDoADRwGDAAcrnoALpwALwoGNh46JFU2OgAbuhU8TYpUOgwyFD%2BkckpcAA8LMBtAHyjYxPAmmDIMPMLqaio1QgI6xupEEzQAFzUfSxU%2BftjhIQ3KW04TMjoOzl59wcp4co4mgBzN4ISLtHQwRLVGZFfplPKgjq6Di4KBAmAcJ4vdDxL4bHhzA7jZSqUJacFwfE3AAS6CYxwAhHAAJKpf5wKBeOBLOAAJXQOB8hQAaq9CpgjsAWIVoHAADLAACOjBY9PW4x4U1mo26EkIQA%3D%3D%3D 

### Expected Behaviour

An additional import statement (or the existing one replaced) should import the generated module of the component, instead of the component class.

```diff
+import { ChildComponentModule } from "./child";
-import { ChildComponent } from "./child";
```

### Actual Behaviour

The generated output is missing a necessary import for the Angular module:

```ts
@NgModule({
  declarations: [MyComponent],
  imports: [CommonModule, ChildComponentModule],
  exports: [MyComponent],
})
```

`ChildComponentModule` is not imported in the generated output. 

### Additional Information

Angular standalone components could be a solid replacement for modules to simplify the generated output.

sean-perkins:
## Description

Provides a default `importMapper` to the Angular generator to generate valid import statements for Angular module projects.

For example, when importing a child component called `ChildComponent`, the import path will be generated as:

```ts
import { ChildComponentModule } from './child';
```

I opted against assigning the `importMapper` to the `DEFAULT_OPTIONS`, since the function needs context to whether the user has selected to generate for standalone components or not. With standalone enabled, the import mapper should not rewrite the import value to include the `Module` postfix. 

Resolves #1205

### DIFF
--- a/packages/core/src/generators/angular.ts
+++ b/packages/core/src/generators/angular.ts
@@ -260,6 +260,37 @@ const processAngularCode =
       (newCode) => stripStateAndPropsRefs(newCode, { replaceWith }),
     );
 
+const defaultImportMapper = (standalone?: boolean) => {
+  return (_component: any, theImport: any, importedValues: any, componentsUsed: any) => {
+    let importPath = theImport.path;
+
+    const { defaultImport, namedImports, starImport } = importedValues;
+
+    let importValue;
+
+    if (starImport) {
+      importValue = ` * as ${starImport} `;
+    } else {
+      importValue = [defaultImport, namedImports]
+        .filter(Boolean)
+        .map((importName) => {
+          if (!standalone) {
+            for (const usedComponentName of componentsUsed || []) {
+              if ((importName || '').indexOf(usedComponentName) > -1) {
+                // replace the component name with the module name
+                return (importName || '').replace(usedComponentName, `${usedComponentName}Module`);
+              }
+            }
+          }
+          return importName;
+        })
+        .join(', ');
+    }
+
+    return `import ${importValue} from '${importPath}';`;
+  };
+};
+
 const DEFAULT_OPTIONS: ToAngularOptions = {
   preserveImports: false,
   preserveFileExtensions: false,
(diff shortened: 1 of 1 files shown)


packages/core/src/generators/angular.ts
```diff
@@ -276,6 +271,42 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
+                    );
+                  }
+                }
+              }
+              return importName;
+            })
+            .join(', ');
+        }
+
+        return `import ${importValue} from '${importPath}';`;
```
samijaber, line 306: I think you're missing wrapping brackets `{ }`, right now the PR imports the same default output, just under a different name. I don't think this is what you want?

Is this what's needed instead?

```suggestion
        return `import {${importValue}} from '${importPath}';`;
```

PR #1210, opened 2023-06-29
chore: remove empty line in vue option api

shengxinjing:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 


packages/core/src/generators/vue/optionsApi.ts
```diff
@@ -179,7 +179,7 @@ export function generateOptionsApiScript(
     }
     return `${str},`;
   };
-
+  // prettier-ignore
```
samijaber, line 182: Can you make this PR work without needing this `prettier-ignore`? It's going to cause issues keeping code consistent down the line

shengxinjing: Prettier changes spaces and line breaks, which affects the final code formatting, hence the disabled setting.

PR #1219, opened 2023-07-08
fix(core): :bug: Preserve meaningful whitespace

### LINKED ISSUE #1218: Meaningful whitespace is not preserved

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [X] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnABLCaYQAUYUEYAzgJRwDeAUHHFMBjoo5Orz5wAPAmIA3AHziJlarS6NmLAHR5UAW2BwA%2FEbgByAOrRMCMwF8lUgPQyF4tgG5udoA%3D%3D

### Expected Behaviour

The meaningful whitespace (`" "`) should be preserved in the react output.

### Actual Behaviour

The space is converted into a line break, causing the output to be `HelloWorld` (without a space)

### Additional Information

I did notice a few things:

1. Adding `{' '}` causes the output to have a space, although not escaped with braces. Unfortunately, prettier also removes this literal space from the `.lite.tsx` file, because it's redundant.
2. Adding `&nbsp;` has no effect on the output. The `&nbsp;` escape is removed.

TzviPM:
Fixes #1218

## Description

Some code generators were joining blocks with newlines (`\n`), which, when run through the `dedent` helper, would then trim any trailing whitespace. This ended up causing problems with significant whitespace in some scenarios.

This PR simply changes the join operation to use the empty string. The parser for mitosis already preserves the significant whitespace, so this makes the output components more closely resemble their mitosis source code. Prettier can still be used to format the output code, which will use a syntax tree that preserves meaningful whitespace while avoiding excessively long line lengths and adding stylistic line breaks.

The main focus for review should be:

1. Ensure that the large changes to snapshots are correct.
2. Considering that many generators have the same structure in terms of a `componentToX` function that calls a `blockToX` function recursively and joins children together, would it make sense to have a helper method for the join operation so that this type of change can be made more accurately across generators?

### DIFF
--- a/packages/core/src/parsers/jsx/__snapshots__/types-identification.test.ts.snap
+++ b/packages/core/src/parsers/jsx/__snapshots__/types-identification.test.ts.snap
@@ -2,15 +2,9 @@
 
 exports[`Signals type parsing > findSignals 1`] = `
 {
-  "context": Set {
-    "context",
-  },
-  "props": Set {
-    "k",
-  },
-  "state": Set {
-    "n",
-  },
+  "context": Set {},
+  "props": Set {},
+  "state": Set {},
 }
 `;
 
--- a/packages/core/src/__tests__/__snapshots__/alpine.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/alpine.test.ts.snap
@@ -21,14 +21,11 @@ exports[`Alpine.js > jsx > Javascript Test > AdvancedRef 1`] = `
 
       <select name=\\"cars\\" id=\\"cars\\">
         <option value=\\"supra\\">GR Supra</option>
-
         <option value=\\"86\\">GR 86</option>
       </select>
     </div>
   </template>
-
   Hello
-
   <span x-html=\\"lowerCaseName()\\"></span>
   ! I can run in React, Qwik, Vue, Solid, or Web Component!
 </div>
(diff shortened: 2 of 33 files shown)


packages/core/src/parsers/jsx/__snapshots__/types-identification.test.ts.snap
TzviPM, line 1: This seems unrelated to my changes. It must have been regenerated when I updated the snapshots. Please confirm if this is correct, or if I should revert this file

samijaber: Are you on the latest `main`? This change shouldn't happen.

samijaber:
You also need to run `yarn run fmt` to lint your PR

samijaber:
Awesome PR, thanks for raising it! I believe this also helps us close https://github.com/BuilderIO/mitosis/issues/310, which was waiting on React/Solid generator to stop creating new lines for each text block. :)

samijaber:
Closing this PR due to inactivity. Feel free to create a new PR with these changes if you're ever able to pick this up again.

PR #1222, opened 2023-07-17
Fix: vue2 components naming

samijaber:


kingzez:
@samijaber Hi~, I don't understand why this is changed, can you explain why?

samijaber:
We were already transforming all component names in the `<template>` logic to kebab-case, but we were not doing that transformation in the `components` export inside `<script>`. 

This was fine because Vue does a transformation in the background, and lets you call the same component using kebab-case or camelCase syntax. But, I found a bug where this broke for single-word component names, like `Block`. It would get exported as `components: { Block }`, but the usage would be `<template> <block /> </template>`, which actually does not work.

So this makes everything consistent between imports and usages.

Issue #1207, opened 2023-06-20
Custom event compile incorrect in child component


kingzez:
### I am interested in helping provide a fix!

Yes，need guide

### Which generators are impacted?

- [ ] All
- [X] Angular
- [X] HTML
- [ ] Preact
- [X] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [X] Svelte
- [X] Vue
- [X] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKNrrgXzgDMoIQ4ByAAQCNlgAbAE3SgDpgIB6EYHVYKgoAoUJFiI4ASQB2YZPACGqOAAVFs%2BfEIkylNmy7A5CkcPQAPcfBZFFyBvCLIZAYxicZcbOjBhWABRgpGCoAJSIwnBwrhAyqPAJijB4ALwoGNi4AQhR0XAAFooyTAzoAEIMyFAB6ABu6DIwAFxwAGIQrmgAog1NEbn5BAA0wnnRRSVlAMKTAObotX0tcAlQxnMD4%2Fmx8RBlbAwQcwEUrvPoFMNw9Y0wYdv4o9H4D3lQ6DDVXgHbADzqTQKbbROIdLqoVIIJZ3CKpAB8cGCEFCbDBnTQMP6%2BBBcDis2KCyhWPucARqxgyXQbEmpXQBJkCxJYRxQ1BMkq1WJt36ZMRSRSNOKdM5NR591Z%2BS48LG0QeOLGXAAVEqokq1ED4II4ABZPgQAQqWaMJhwaZkSAyO5qvEycFoa74i7XADyHKqUDg2ohOHI4ptNtkcAAash0NcfXpxXBlEiQiowMoVDgYha4nc4DIIPAqOKIwUTanwOmmqsChAHKbFAxUBAi2BGHgU8jQgG4H9EwBaYxabboiGpABE%2FcxABJxWFB32ZAyiYPaTMLgFx8tJ9PRUOFxUPcuJ1PovC%2Flwuz2FDK8kHswVWMQoIoQOgAO7QADWmezqagH3cNwsrl88CDgASugig%2Fq495gPuSpcEAA%3D%3D%3D

### Expected Behaviour

```jsx
import { useStore } from '@builder.io/mitosis'
import { Input as PaInput } from '../input'

[...]

PR #1228, opened 2023-07-19
fix: custom event as props for child component

kingzez:
## Description

Add a short description of:
- custom event as props for child component
-  use arrow function wrap custom event

Fixed  https://github.com/BuilderIO/mitosis/issues/1207 
- Vue 
- Svelte


samijaber (approved):
@kingzez Thank you for your patience, took me a while to get around to looking at this.

Makes perfect sense, thanks for digging into it and finding a fix 🙏🏽

kingzez:
Great, you finally have time to look this, this PR only fixes Vue and  Svelte to prove my thought, other frameworks will be fixed later~

PR #1274, opened 2023-10-19
Fix srcset typo

samijaber:
## Description

Add a short description of:
- what changes you made, 
- why you made them, and 

### DIFF
--- a/packages/core/src/plugins/compile-away-builder-components.ts
+++ b/packages/core/src/plugins/compile-away-builder-components.ts
@@ -349,7 +349,7 @@ export const components: CompileAwayComponentsMap = {
       properties: noUndefined({
         loading: 'lazy',
         sizes: node.properties.sizes,
-        srcSet: srcSet || null,
+        srcset: srcSet || null,
         alt: node.properties.altText,
       }),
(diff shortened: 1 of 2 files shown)


packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -349,7 +349,7 @@ export const components: CompileAwayComponentsMap = {
       properties: noUndefined({
         loading: 'lazy',
         sizes: node.properties.sizes,
-        srcSet: srcSet || null,
```
steve8708, line 352: this was intentional - for react it has to be capitalized. can you verify that we'll still output `srcSet` for react with this change?

steve8708: iirc we have logic that will lowercase attributes like this for non-jsx frameworks, but don't have the reverse for jsx (convert `srcset` to `srcSet` for react)

samijaber: Yeah, I realized that after raising the PR, that's why I closed it 👍🏽

PR #1282, opened 2023-10-25
only add "use client" when needed

steve8708:
for react, we are blanket adding "use client" nearly everywhere. in most cases, we do not want this. react native and preact should never have it, and any component without client only hooks or event listeners shouldn't have it either

### DIFF
--- a/packages/cli/src/__tests__/__snapshots__/cli-integration.test.ts.snap
+++ b/packages/cli/src/__tests__/__snapshots__/cli-integration.test.ts.snap
@@ -343,73 +343,46 @@ function MyComponent(props) {
       }}
     >
       <div
-        className=\\"builder-columns\\"
         css={{
+          gap: \\"20px\\",
           display: \\"flex\\",
-          \\"@media (max-width: 999px)\\": {
+          \\"@media (max-width: 991px)\\": {
             flexDirection: \\"column\\",
             alignItems: \\"stretch\\",
+            gap: \\"0px\\",
           },
         }}
       >
         <div
-          className=\\"builder-column\\"
           css={{
             display: \\"flex\\",
             flexDirection: \\"column\\",
             alignItems: \\"stretch\\",
             lineHeight: \\"normal\\",
-            width: \\"calc(33.333333333333336% - 13.333333333333334px)\\",
+            width: \\"33%\\",
             marginLeft: \\"0px\\",
-            \\"@media (max-width: 999px)\\": {
+            \\"@media (max-width: 991px)\\": {
               width: \\"100%\\",
               marginLeft: 0,
             },
           }}
         >
-          <div
+          <img
+            loading=\\"lazy\\"
+            sizes=\\"(max-width: 638px) 100vw, 27vw\\"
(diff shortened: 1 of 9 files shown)


packages/core/src/generators/react/generator.ts
```diff
@@ -284,6 +286,31 @@ const getPropsDefinition = ({ json }: { json: MitosisComponent }) => {
...
+    if (isMitosisNode(node)) {
+      if (Object.keys(node.bindings).filter((item) => item.startsWith('on')).length) {
+        foundEventListener = true;
+        this.stop();
+      }
+    }
+  });
+
+  return foundEventListener;
+};
```
samijaber, line 312: This logic already exists here (minus the `on*` bindings):
 https://github.com/BuilderIO/mitosis/blob/7de6a2d94e69475bc04c7bd77db4bffbee410b6c/packages/core/src/generators/rsc.ts#L55-L66

packages/cli/src/__tests__/__snapshots__/cli-integration.test.ts.snap
```diff
@@ -326,15 +325,11 @@ export default MyComponent;
...
-"\\"use client\\";
-import * as React from \\"react\\";
+"import * as React from \\"react\\";
 
 function MyComponent(props) {
   return (
     <div
-      stackColumnsAt=\\"tablet\\"
-      space={36}
-      reverseColumnsWhenStacked={false}
```
samijaber, line 337: I think something went wrong in your PR, seeing as it's dropping valid bindings now.

steve8708: this is actually correct, these should not be applied to a `div`

samijaber: Gotcha. Doesn't make sense that attributes are appearing/disappearing in this PR though, since the only difference is `use client` logic. I think your branch is maybe not on latest `main`?

samijaber (approved):

packages/core/src/generators/react/generator.ts
```diff
@@ -284,6 +286,31 @@ const getPropsDefinition = ({ json }: { json: MitosisComponent }) => {
...
 
+// Test if a react component is a client component and needs
+// "use client" by seeing if it has any client-only features
+// like refs, effects, state, or event listeners
+const needsUseClient = (json: MitosisComponent) => {
+  if (json.hooks.onMount?.code) return true;
+  if (json.hooks.onUnMount?.code) return true;
+  if (json.hooks.onUpdate?.length) return true;
+  if (Object.keys(json.refs).length) return true;
+  if (Object.keys(json.state).length) return true;
```
samijaber, line 297: this check is excessive. we only need to check `property` types, a.k.a. `useState`s. The rest of `json.state` is functions/methods, which are OK.

packages/cli/src/__tests__/__snapshots__/cli-integration.test.ts.snap
samijaber: Seems like the CLI snapshots have been broken for a while, not sure how I missed that:

 CleanShot 2023-10-26 at 14 25 04@2x

You might want to run `yarn snapupdate` in `packages/cli` (not `core`) to update the failing ones.

samijaber: went ahead and did it

steve8708: yeah weirdly thats what i did, no clue why it didn't give me the same output, but thanks for updating 🙏

steve8708:
thanks for the feedback @samijaber - made the updates asked and fixed another couple things i spotted, can you give this one more look over and if looks good we can merge?

samijaber (approved):

PR #1306, opened 2023-11-27
chore: stop GitHub highlighting comments in json as errors

epreston:
GitHub also highlights comments in json as errors by default.  Adding the following to  `.gitattributes` stops that.

```
*.json  linguist-language=JSON-with-Comments
```

Makes project readme prettier on github.

Other possibilities discussed here:
https://github.com/microsoft/rushstack/issues/1088

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
+*.json linguist-language=JSON-with-Comments
+
 /.yarn/releases/** binary
 /.yarn/plugins/** binary


samijaber:
I don't know if it makes sense to do this. Many JSON files (like `package.json`) cannot have comments, and so the editor should reflect that by marking them as errors. Only special JSON files that allow comments should have this setting enabled.

epreston:
This is a cosmetic setting that changes how Github's rendering behaviour, when viewing the website.
It will not allow comments in json files.

samijaber (approved):
Thanks for clarifying! I mistook this file for `.editorconfig`

epreston:
Happy holidays.

PR #1314, opened 2023-12-05
feat: add native attributes to angular metadata to fix [attr.XXX] issues

### LINKED ISSUE #1313: Angular [attr.disabled]

Yes

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=IYOw5grgNsBOQ%3D%3D%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFAT0AD0iw4bEjmRN4JZADsAxjG6a4AWQCeAYQqRN6TTAAUYcmFQBKRFLhwo6GMigHbEh4eADwswABuAHyBQSHAmmDIMDGxHmGoOHRM6CwAvAgOEE4c6ZnZLIQpQTzRqXAAEuhMTBAAhHAAkiRwqBToEFZwYDioGKhwAAYA2qVZOQC6uVTKTBhUE3CCcCDoADSKImU5cCJwOEwA7jjG4zBQyOgccADymkzGk5qqTBs4miyTLRKeI5X5eTzoEAQcI5ABccAAFjAYE5YTweH8AOaqHCcbg8bHANgY5FQYAMfAAWjo8TCmkxAGJUMYbDgZK0qsEeGEojFnJJCEA%3D%3D

### Expected Behaviour

`<my-component [disabled]="false">` shouldn't disable the `<input>` inside `my-component`.

### Actual Behaviour

All `[attr.XXX]` attributes will be rendered to the DOM by Angular: https://angular.io/guide/attribute-binding#syntax

Only `null` and `undefined` are removed.

### Additional Information

-> There should be another check [here](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L214) to check if the key is a valid attribute for the native HTML tag.

I have some solutions in mind: 

1. add a mapping which [tag](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/constants/html_tags.ts) should keep which attribute like `input: ["disabled", ...]`
2. Use the `useMetadata`, so the user could decide which prop is "native"

nmerget:
## Description

closes #1313

- add new `nativeAttributes` option to `useMetadata.angular` to skip generating native attributes like [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) with the `attr` prefix
- in angular all `attr.XXX` attributes are passed to the DOM if they are not `null` or `undefined`


samijaber (approved):
Thanks @nmerget !

Issue #613, opened 2022-08-03
React generator fails when there is a comment with a single quote


ja-rivas:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [X] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFAT0AD0iw4bEjmRN4JZADsAxjG6a4AWQCeAYQqRN6TTAAUYcmFQBKRBLhxtETangBtTRwQdAAaOAwYADkg9ABdOABeFAxsPHRbKmx0ADd0KmdJd1ItXX1FCEwKdBgAC2BNAHNbVwQijx4eOAAVOtRPHE0Acng6Ai9wZnQWNs9vVAgmdA4mCCbBnpE4TZw4fF9BgqLCKQ8oauQoA1sZgB4WYGyAPhmPG%2FqwZBgXj09UVASEK0fsCfl4VlAAFzUOhMZB5ELfDyEY4gn7ZHCw9AAwLBFGouDeUw1AYNLEIWw5awwVwJR7harRYIU3I2Di4KCkmAcdGY5x44E8Z6ogAS6CYKwAhHAAJL9AxQLRbAwAJXQOF0YQAanCwpUmMAWGFoHAADLAACOjBYEtuPHuTyKh0IQA%3D%3D%3D

### Expected Behaviour

```jsx
import * as React from "react";
import { useState } from "react";

export default function MyComponent(props) {
  const [name, setName] = useState(() => "Steve");

  function doSomething() {
    // This can't be compiled
    console.log("This is a test");
  }

  return (
    <>
      <div>
        <input
          className="input"
          value={name}
          onChange={(event) => setName(event.target.value)}
        />
        Hello! I can run in React, Vue, Solid, or Liquid!
      </div>
      <style jsx>{`
        .input {
          color: blue;
        }
      `}</style>
    </>
  );
}

```

### Actual Behaviour

```jsx
import * as React from "react";
import { useState } from "react";

export default function MyComponent(props) {
[...]

Issue #250, opened 2022-03-28
Comments within `useState` block cause crash


samijaber:
**Describe the bug**
comments right above _function_ declarations in `useState` cause endless loop/crash in Mitosis

**To Reproduce**
<img width="1920" alt="image" src="https://user-images.githubusercontent.com/1393142/160486630-a4c0cf5d-4de8-4178-ac7c-ad77ebe2d560.png">
```
import { useState } from "@builder.io/mitosis";

export default function MyComponent(props) {
  const state = useState({
    // a comment
    name: 'steve',

    // another comment
    handleFn() {}
  })
  
  
  return (
    <div>
      <input
        css={{
          color: "red",
        }}
        value={state.name}
        onChange={(event) => state.name = (event.target.value)}
      />
      Hello
      {name}! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
}
```

If possible, a link to a https://mitosis.builder.io/ fiddle containing the bug: Fiddle causes infinite loop, so I put a code block here instead.

**Expected behavior**
Expect comments to be ignored/stripped before any Mitosis logic runs, and not cause any issues

**Additional context**
The comment above a constant property works fine, but one above a function declaration isn't

samijaber:
Similarly, comments within JSX cause issues: [see](https://mitosis.builder.io/?outputTab=react&code=export+default+function+MyBasicComponent%28%29+%7B%0A++return+%28%0A++++%3Cdiv%0A++++++%2F%2F+%40ts-ignore%0A++++++foo%3D%7B%27bar%27%7D%0A++++%3E%0A++++++Hello%21+I+can+run+in+React%2C+Vue%2C+Solid%2C+or+Liquid%21%0A++++%3C%2Fdiv%3E%0A++%29%3B%0A%7D%0A)

They end up making it to the final version of the code. The parser thinks that `div //@tsignore` is the name of a component.

As a first step, we might want to trim all `//...` comments from the end of every code block.

SamiKamal:
I'm working on this 👍🏾

samijaber:
I think one of the main reasons we have these issues is that the JSX parser works by updating the AST and replacing it with the appropriate JSON:

https://github.com/BuilderIO/mitosis/blob/8980fd5d4130ffa719d42469677d1667bce0953a/packages/core/src/parsers/jsx/jsx.ts#L148

[...]

PR #1320, opened 2023-12-17
Fix/comments with quotes

SamiKamal:
## Description

Add a short description of:
- Passed `{ comments: false }` to `parseCode`, 
- The reason was adding quotes in a comment would break the parser, more context check #613 and #250 (#250 is more broad, this fix is focused on #613)

### DIFF
--- a/packages/core/src/__tests__/data/comments-with-quotes.raw.tsx
+++ b/packages/core/src/__tests__/data/comments-with-quotes.raw.tsx
@@ -1,0 +1,29 @@
+import { useStore } from "@builder.io/mitosis";
+
+export default function MyComponent(props) {
+  const state = useStore({
+    name: "Steve",
+    // can't be compiled
+    handleFn() {}
+  });
+
+    function doSomething() {
+    console.log("This is a aa");
+  }
+
+  
+  return (
+    <div>
+      <input
+        css={{
+          color: "red",
+        }}
+        value={state.name}
+        onChange={(event) => { state.name = event.target.value }}
+        onClick={() => doSomething()}
+      />
+      Hello
+      {state.name}! I can run in React, Vue, Solid, or Liquid!
+    </div>
+  );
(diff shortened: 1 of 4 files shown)


samijaber:
This is great! You just need to update test snapshots with your latest test:

```
cd packages/core && yarn snapupdate
```

samijaber:
You also need to lint using `yarn fmt:prettier`

SamiKamal:
> You also need to lint using `yarn fmt:prettier`

All done now 👍🏾

packages/core/src/__tests__/__snapshots__/vue.test.ts.snap
```diff
@@ -722,9 +722,7 @@ export default defineComponent({
 
   methods: {
     findAndRunScripts() {
-      // TODO: Move this function to standalone one in '@builder.io/utils'
       if (this.$refs.elem && typeof window !== \\"undefined\\") {
-        /** @type {HTMLScriptElement[]} */
```
samijaber, line 727: One unwanted side-effect of your solution is that it removes all comments from all blocks. That would be a lot of useful content that we don't want to lose...

I will investigate a better approach.

SamiKamal: Opps, I guess I misunderstood this part, I thought it meant removing the comments from the whole code.

> #### Expected behavior
> Expect comments to be ignored/stripped before any Mitosis logic runs, and not cause any issues

samijaber: Sorry about that. The ideal expected behaviour is that comments never cause issues and that most comments make it to the final output.

SamiKamal: No worries, I'll look for another solution

PR #1324, opened 2023-12-27
fix: update esbuild to fix typescript "satisfies" not supported issue

rqzheng2015:
## Description

The current `Mitosis` will throw out error when compiling the `typescript` codes that include `satisfies` operator. 

```
ERROR: Expected ";" but found "satisfies"
```

Need to upgrade `esbuild` to fix that.

## Reference:

https://github.com/remix-run/remix/issues/4639


samijaber (approved):
Sweet, thanks!

PR #1334, opened 2024-01-24
feat(mitosis): code gen support for named slots

### LINKED ISSUE #1301: React named slots getting prefixed with the word 'Slot'

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [X] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=JYWwDg9gTgLgBAbzgZQDYRgGjgVwM4CmyMAhjAdvkTNAXAL5wBmUEIcARAAIBGOwqACYEoAOmAQA9CGA08wPBwDcAKBWhIsTqMnAAdsIAeogMZ5FqlQUOb4%2B8lCYkTdAMIkog120h6CemAAFVjA8OGtyAzCAIRJCYIhQxBU4OAALAhJhKAB%2BAC44Ej0AT1VUnghBYvzCkrLmCAwRGqLSlXo1KxtoeGEnHFR4Jhw9ExgJPTh3T29wCD8AgAowELwC6a8fef8g1YBKZJS4KAIYHChJxaPUgB5BYAA3a9S4E1Q4vABeBA4HggBaEweQQcDovOAAPmet3uD1e73MADkSCACJ8OBksiIOFDweCbmgMHA9Ci0RjMtkOHBJLi8TdJLDafjYfCPsjUeiKlUcdCXgT0PASRyOFzilSabz6YzJSy3mzSeimI0HDy8XzCYKFRwlU0oOKmXyGY8DbcNdSDVLjUc9qoOkA%3D%3D%3D

### Expected Behaviour

The name of the prop should be kept as it is. 

```js
import "./index.css";

function CardComponent(props) {
  return (
    <div className="ve-card">

      <div className="header">{props.header}</div>    
      <div className="body">{props.body}</div>           
      <div className="footer">{props.footer}</div>      

      {props.children}

    </div>
  );
}

export default CardComponent;

```

### Actual Behaviour

The name of the prop is prefixed with the word 'Slot' and due to this the prop types are mismatched

```js
import "./index.css";

function CardComponent(props) {
  return (
    <div className="ve-card">

      <div className="header">{props.slotHeader}</div>

      <div className="body">{props.slotBody}</div>

      <div className="footer">{props.slotFooter}</div>

      {props.children}
    </div>
  );
}

export default CardComponent;

```

### Additional Information

These are the dependencies that I'm using :

```js
"dependencies": {
    "@builder.io/mitosis": "^0.0.106",
    "@builder.io/mitosis-cli": "^0.0.64",
    "@ctrl/tinycolor": "^3.6.0",
    "@vitejs/plugin-react": "^4.0.4",
    "@vitejs/plugin-vue": "^4.3.3",
    "kill-port": "^2.0.1",
    "postcss-cli": "^10.1.0"
  },
```

For the time being, I have written a small script to remove the 'Slot' prefix after transpilation of Mitosis code. This problem does not occur in Angular and Vue3. I looked into the codebase and found that it was [this piece of code](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/react/blocks.ts) that was injecting this 'Slot' prefix :

```js
if (!slotName) {
      // TODO: update MitosisNode for simple code
      const key = Object.keys(json.bindings).find(Boolean);
      if (key && parentSlots) {
        const propKey = camelCase('Slot' + key[0].toUpperCase() + key.substring(1));      <-------
        parentSlots.push({ key: propKey, value: json.bindings[key]?.code });
        return '';
      }

      const children = processBinding('props.children', options);
      return `{${children} ${hasChildren ? `|| (${renderChildren()})` : ''}}`;
    }

    let slotProp = processBinding(slotName as string, options).replace('name=', '');

    if (!slotProp.startsWith('props.slot')) {
      slotProp = `props.slot${upperFirst(camelCase(slotProp))}`;       <-------
    }
```

I also tried naming the slots with the 'slot' prefix like this : 

```js
import { Slot, useState, useStore } from "@builder.io/mitosis";

import "./index.css";

export interface CardComponentProps extends BaseProps {
  slotHeader?: any;
  slotBody?: any;
  slotFooter?: any;
}


export default function CardComponent(props: CardComponentProps) {

  return (
    <div
      class={"ve-card"}
    >
      <div className="header">
        <Slot name="slotHeader" />
      </div>
      <div className="body">
        <Slot name="slotBody" />
      </div>
      <div className="footer">
        <Slot name="slotFooter" />
      </div>
      <Slot />
    </div>
  );
}

```

But then again, it's getting prefixed :

```js
import "./index.css";

function CardComponent(props) {
  return (
    <div className="ve-card">
      <div className="header">{props.slotSlotHeader}</div>

      <div className="body">{props.slotSlotBody}</div>

      <div className="footer">{props.slotSlotFooter}</div>

      {props.children}
    </div>
  );
}

export default CardComponent;

```

Is this the intended output ? If so, could you please tell me how I could tackle the type mismatches ? Also, is there a proper way to declare the prop types here ? Should we leave it as `any` or is `JSX.IntrinsicElements` more appropriate ? 

Any help or guidance on this matter would be greatly appreciated.

Thanks in advance !

mrkoreye:
## Description

This PR adds support for named slots in custom component code gen: https://www.loom.com/share/f3ad5eee72f9486a9130ce1f09ceafa0

Fixes #1301

### DIFF
--- a/packages/core/src/generators/react/types.ts
+++ b/packages/core/src/generators/react/types.ts
@@ -15,5 +15,4 @@ export interface ToReactOptions extends BaseTranspilerOptions {
    */
   contextType?: 'context' | 'prop-drill';
   addUseClientDirectiveIfNeeded?: boolean;
-  noPrefixSlots?: boolean;
 }
--- a/packages/core/src/generators/react/types.ts
+++ b/packages/core/src/generators/react/types.ts
@@ -15,4 +15,5 @@ export interface ToReactOptions extends BaseTranspilerOptions {
    */
   contextType?: 'context' | 'prop-drill';
   addUseClientDirectiveIfNeeded?: boolean;
+  noPrefixSlots?: boolean;
 }
--- a/docs/components.md
+++ b/docs/components.md
@@ -232,14 +232,14 @@ export default function MyComponent(props) {
 
 ### Slot
 
-When you want to register a named slot you do so using the `slot` prop.
+When you want to register a named slot you do so using a prop.
 
 ```jsx
 <div>
   <Layout
-    slotTop={<NavBar/>}
-    slotLeft={<Sidebar/>}
-    slotCenter={<Content/>}
+    top={<NavBar/>}
+    left={<Sidebar/>}
+    center={<Content/>}
   />
     anything else
   </Layout>
@@ -254,9 +254,9 @@ If the `Layout` component was also a Mitosis component then we simply use the re
 export default function Layout(props) {
   return (
     <div className="layout">
-      <div className="top">{props.slotTop}</div>
-      <div className="left">{props.slotLeft}</div>
-      <div className="center">{props.slotCenter}</div>
+      <div className="top">{props.top}</div>
+      <div className="left">{props.left}</div>
+      <div className="center">{props.center}</div>
       {props.children}
     </div>
   );
@@ -286,7 +286,7 @@ export default function Layout(props) {
 }
 ```
 
-For vue component `slot` prop will be compiled into named slot
+For vue component a `slot` prop will be compiled into named slot
 
 ```html
 <div class="layout">
@@ -347,9 +347,9 @@ useMetadata({
 export default function Layout(props) {
   return (
     <div className="layout">
-      <div className="top">{props.slotTop}</div>
-      <div className="left">{props.slotLeft}</div>
-      <div className="center">{props.slotCenter}</div>
+      <div className="top">{props.top}</div>
+      <div className="left">{props.left}</div>
+      <div className="center">{props.center}</div>
       {props.children}
     </div>
   );
--- a/packages/core/src/__tests__/__snapshots__/preact.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/preact.test.ts.snap
@@ -600,7 +600,7 @@ function ContentSlotJsxCode(props) {
             className={cls()}
           >
             {showContent && props.slotContent ? (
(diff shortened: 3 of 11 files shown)


packages/core/src/generators/react/types.ts
```diff
@@ -15,4 +15,5 @@ export interface ToReactOptions extends BaseTranspilerOptions {
    */
   contextType?: 'context' | 'prop-drill';
   addUseClientDirectiveIfNeeded?: boolean;
+  noPrefixSlots?: boolean;
```
samijaber, line 18: Given https://github.com/BuilderIO/mitosis/issues/1301, I think we want the default to be not prefixing.

Would you mind switching this to: 

```suggestion
  prefixSlots?: boolean;
```

and only prefixing if the option was explicitly provided?


This will require updating the snapshot tests, which you can do by running:

- `yarn ci:build` in the root, and then
- `yarn snapupdate` in `package/core`

samijaber: Scratch that: if you don't mind, completely remove the prefixing logic altogether! We don't need it.

mrkoreye: 👍

mrkoreye:
@samijaber updated tests and readme to reflect new slot behavior

samijaber (approved):
nice 👍🏽

PR #1335, opened 2024-01-24
fix: less css for compiled away columns

manucorporat:


samijaber:
@manucorporat need to run `yarn snapupdate` in `package/cli` to update the snapshot for this logic

samijaber:
@manucorporat do you still need this PR merged?

PR #1351, opened 2024-02-01
[fix] `preventDefault` in Qwik

sidmohanty11:
## Description

This PR updates the `preventDefault` logic for Qwik by:
- removing `=""` from `preventdefault:eventName=""`
- removes any occurrences of `(event*).preventDefault();` inside the handler

For example test,

```jsx
import { useState } from '@builder.io/mitosis';

  const [name, setName] = useState('Steve');

  return (
    <div>
      <input
        value={name}
        onChange={(event) => {
          event.preventDefault();

          setName(event.target.value);
        }}
      />
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
}
```

gives output,

```jsx
import { $, Fragment, component$, h, useStore } from "@builder.io/qwik";

export const MyComponent = component$((props) => {
  const state = useStore({ name: "Steve" });

  return (
    <div>
      <input
        preventdefault:change
        value={state.name}
        onChange$={$((event) => {
          state.name = event.target.value;
        })}
      />
      Hello! I can run in React, Vue, Solid, or Liquid!
    </div>
  );
});

export default MyComponent;

```

### DIFF
--- a/packages/core/src/generators/qwik/src-generator.ts
+++ b/packages/core/src/generators/qwik/src-generator.ts
@@ -359,6 +359,10 @@ export class SrcBuilder {
         if (key === 'innerHTML') key = 'dangerouslySetInnerHTML';
         if (key === 'dataSet') return; // ignore
         if (self.isJSX) {
+          if (key.includes(':')) {
+            self.emit(' ', key);
+            return;
+          }
           self.emit(' ', key, '=');
           if (typeof value == 'string' && value.startsWith('"') && value.endsWith('"')) {
             self.emit(value);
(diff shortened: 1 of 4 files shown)


packages/core/src/generators/qwik/src-generator.ts
```diff
@@ -359,6 +359,10 @@ export class SrcBuilder {
         if (key === 'innerHTML') key = 'dangerouslySetInnerHTML';
         if (key === 'dataSet') return; // ignore
         if (self.isJSX) {
+          if (key.includes(':')) {
```
samijaber, line 362: To be safe, I would make sure the `value` is empty too:

```suggestion
          if (key.includes(':') && value == '') {
```

samijaber (approved):
Awesome! Just one small thing

packages/core/src/generators/qwik/src-generator.ts
sidmohanty11: Sure, thanks a lot! I've added it

samijaber (approved):

Issue #1350, opened 2024-01-30
Slots: Angular and Vue behave diffently


nmerget:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [X] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=G4VwpkA%3D&code=JYWwDg9gTgLgBAbzgZQDYXgXzgMyhEOAIgAEAjAV2FQBMBTKAOmAgHoRgYIBnYbogNwAoIXQAekWHHo4AhhVTwcFAHYBjGCxVwAsgE8AwgUgq6KmAAow%2BMNwCUiEXDhDnUOjApRtF187gAPDTAAG4AfH7%2BgWgYcCqyIHQAvEQwdNwwMTBEcKwRUQGsweF%2BdsKYQA

### Expected Behaviour

If you use a named slot it should rather be the original name or `kebabCase` with dashes.

### Actual Behaviour

Currently, the generator for [Vue](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/vue/blocks.ts#L302) uses `toLowerCase()`.
[Angular](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L150) is using `kebabCase`.

It should be aligned. It should be `kebabCase` for Vue as well.

PR #1368, opened 2024-02-18
 fix for issue No #1350

GhoshAnup:
formatting slot name to kebab casing to make it consistent across Angular and Vu js

## Description

- I have made changes in slots.ts file. I have added a function to format the slot name such a way, so it stays consistent after compilation by Angular or Vu js. 

- As with current implementation the slots names were not consistent, Anguler js compiling to kebab case where vu js wasn't. Also, as per VU js documentation it's a good practice to name as per kebab casing. So, with this new function we will format the slot name to kebab casing. 
-
- I have tried to implement minimum changes to achieve the desired result of maintaining consistency in the slot name across Angular and VU js. So with this new method I am able to fix the issue. I have tested my changes in the provided repro steps. 

### DIFF
--- a/packages/core/src/helpers/slots.ts
+++ b/packages/core/src/helpers/slots.ts
@@ -8,7 +8,7 @@ export const isSlotProperty = (key: string, slotPrefix: string = SLOT_PREFIX): b
   key.startsWith(slotPrefix);
 
 export const stripSlotPrefix = (key: string, slotPrefix: string = SLOT_PREFIX): string =>
-  isSlotProperty(key, slotPrefix) ? key.substring(slotPrefix.length) : key;
+  isSlotProperty(key, slotPrefix) ? convertToKebabCase(key.substring(slotPrefix.length)) : convertToKebabCase(key);
 
 export function replaceSlotsInString(code: string, mapper: SlotMapper) {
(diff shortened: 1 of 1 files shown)


packages/core/src/helpers/slots.ts
```diff
@@ -8,7 +8,7 @@ export const isSlotProperty = (key: string, slotPrefix: string = SLOT_PREFIX): b
   key.startsWith(slotPrefix);
 
 export const stripSlotPrefix = (key: string, slotPrefix: string = SLOT_PREFIX): string =>
-  isSlotProperty(key, slotPrefix) ? key.substring(slotPrefix.length) : key;
+  isSlotProperty(key, slotPrefix) ? convertToKebabCase(key.substring(slotPrefix.length)) : convertToKebabCase(key);
```
nmerget, line 11: I think this shouldn't be applied to all generators.

Instead, it should be changed inside the [vue generator](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/vue/blocks.ts#L112).

Moreover, you could use the same code like the [angular generator](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L150), which is using `kebabCase` from `lodash`.

Issue #1322, opened 2023-12-19
Support for custom generators / targets


commax89:
### I am interested in helping provide a feature!

Yes

### Which generators are impacted?

- [X] All
- [ ] Angular
- [ ] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### What problem does this feature solve?

Some projects may need to customize, rewrite or write generators from scratch in order to achieve some specific business requirements that are not compatible with current coding standards.

### What does the proposed API look like?

From an initial high level investigation it should be possibile to load custom targets directly from the config file, in the cli package it should be enough to merge the current targets with the ones from the options. This should also enable generator replacements.

samijaber:
I'm open to adding this! How to do so:

- Add a `generators` dictionary prop to the `MitosisConfig`:
https://github.com/BuilderIO/mitosis/blob/e862056eca675887bd975bf5696c61666d0f3532/packages/core/src/types/config.ts#L19

- Update this call to take into account `options.generators`, by checking for `options.generators[target]` before relying on the default generators:
https://github.com/BuilderIO/mitosis/blob/e862056eca675887bd975bf5696c61666d0f3532/packages/cli/src/build/build.ts#L278

JerryWu1234:
I will do it

PR #1370, opened 2024-02-21
Support for custom generators / targets

JerryWu1234:
## Description

#1322

Add a short description of:
- Support for custom generators / targets
- #1322


packages/cli/src/build/build.ts
```diff
@@ -269,7 +251,7 @@ const getTargetContexts = (options: MitosisConfig) =>
   options.targets.map(
     (target): TargetContext => ({
       target,
-      generator: getGeneratorForTarget({ target }) as any,
```
samijaber, line 272: This is the default behaviour, we still want it. Otherwise every user has to provide every generator every time.

```ts
generator: options.generators[target] || getGeneratorForTarget({ target })
```

samijaber (changes requested):
Need to preserve default behaviour

packages/cli/src/build/build.ts
JerryWu1234: I set a default behavior in `packages/cli/src/build/build.ts`
my way [`generators: Object.assign(targets, config?.generators)`](https://github.com/BuilderIO/mitosis/blob/71511a8668f46a3058b8f725abb0943bc49f1d61/packages/cli/src/build/build.ts#L57)

JerryWu1234: @samijaber I just want to confirm if we still have to tweak?

JerryWu1234:
> Need to preserve default behaviour

@samijaber please check

packages/core/src/types/config.ts
```diff
@@ -1,22 +1,30 @@
...
+export type generatorsOption = {
+  [K in Target]: NonNullable<Targets[K]>;
+};
 
 export type MitosisConfig = {
+  generators?: generatorsOption;
+  /**
+   * link https://github.com/BuilderIO/mitosis/issues/1322
+   * to achieve some specific business requirements that are not compatible with current coding standards.
+   */
```
samijaber, line 27:
```suggestion
```

samijaber:
Thanks for putting this together @JerryWu1234 ! Thank you for your patience in getting it merged as well.

PR #1384, opened 2024-03-18
fix: fix ref issue when transforming to Solid.js code

rqzheng2015:
## Description

Fix ref issue when mitosis build target is `solid`

## Issue

I used the `ref example` provided on the official [doc](https://github.com/BuilderIO/mitosis/blob/84645d4ee333996716c1a9ac8e8625b283ade624/docs/hooks.md#forwardref-for-react).

Playground: [Click here](https://mitosis.builder.io/?outputTab=E4UwhgxgLkA%3D&code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAknmOjABRhQRgDOAlHAN4BQcOFGAx0UcgB5i9RsJQBeXq3YcAdDIYwASigC%2BcAPQA%2BANz89QA) 

It is correctly transformed on React,  which will present the output codes like these below:

```tsx
import { forwardRef } from "react";

const MyInput = forwardRef(function MyInput(props, inputRef) {
  return <input ref={inputRef} />;
});

export default MyInput;
```

But, it will cause issue on Solid, and we could see the issue at console panel.

image

## Reason
The issue is caused by the `getRefsString` method in Solid.js generator, which can be read [here](https://github.com/BuilderIO/mitosis/blob/c5ac90d85071d12948ff9b0ccbc04b434123876f/packages/core/src/generators/solid/index.ts#L64-L70).

```tsx
const getRefsString = (json: MitosisComponent, options: ToSolidOptions) =>
  Array.from(getRefs(json))
    .map((ref) => {
      const typeParameter = (options.typescript && json['refs'][ref]?.typeParameter) || '';
      return `let ${ref}${typeParameter ? ': ' + typeParameter : ''};`;
    })
    .join('\n');
```

And with the [insertion](https://github.com/BuilderIO/mitosis/blob/c5ac90d85071d12948ff9b0ccbc04b434123876f/packages/core/src/generators/solid/index.ts#L194) ` ${getRefsString(json, options)}` in solid code, eventaully, the generated solid.js code will become something like this:

```tsx
function MyComponent(props:any) {
  let props.inputRef;
  return (
    <input  ref={props.inputRef!}  />
    )
}
export default MyComponent;
```

And when these code met with [prettier format function](https://github.com/BuilderIO/mitosis/blob/c5ac90d85071d12948ff9b0ccbc04b434123876f/packages/core/src/generators/solid/index.ts#L254-L259) in Solid.js generator, it will throw out the error, which leads to the break of Solid.js code  bundle.

We could easily reproduce it in a online prettier [editor](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAsgJ4DCEAtgA7QIwAUVAThFQM5ICGUxAlPsAA6UfPgA2cGPmas2AOhxQq6GACU4qANzDRTSeiYj6O0fgA8i5VPx7UAXmAz2CpSvWoAhAF9RAegB8JqK8wl7CcAAeNExSACYanOhiUiTk1LSwmiAANCCsuNBsyKCcTCwA7gAKpQhFKJxi5ZzERbkARkycYADWkgDKnBRwADKKcMioDWxw7Z09-VRdigDmyDBM6DMg0xQ4axtbkVRwTDhDsA0AKidQpThwdZNi07lsKxIAiugQ8BNTWwArNgRPrvOBfH7jJBPF4gACO33glRY7GQIE4bAAtFA4HB4rEciB1pwcGIVmkKJw0Q0xIS3lBlhIAIIwdY4NoqOCVE6jHF-Z5bAAWMAoYgA6oKcPA2IswHA+rUpTgAG5S4hosBsVogZWbACSUHisD6YFOVBgTMNfRgxAk-NhMmmYs6VDRzAeJ2V41yimmMWRnGWlPtW0WTD9aJtxzYppw5sJzEUMDFOFiMEFyAAHAAGXJ6BE4PQBoNU6H-XIwThtFNpjNIABMuXQ00uVce5ZAcAobTx+OG3GW6EDcAAYhAmJTWStqSoICAvF4gA).
image

## Solution

Updated at May, 4th:

Remove ref definition when it's passed by props, such as `props.inputRef`.
```typescript
const getRefsString = (json: MitosisComponent, options: ToSolidOptions) =>
  Array.from(getRefs(json))
    .map((ref) => {
      if (ref.includes('.')) return '';
      const typeParameter = (options.typescript && json['refs'][ref]?.typeParameter) || '';
      return `let ${ref}${typeParameter ? ': ' + typeParameter : ''};`;
    })
    .join('\n');
```

Then it will behave in two ways:

- Ref is created by `useRef`, will still remain.
- Ref is passed by `props`, will not remain after this PR.

~~Remove the getRefsString method calling.~~

~~Then the Solid.js generator will generae Solid.js code like these below, which works as expected and will not throw out any error.~~
```tsx
function MyComponent(props:any) {
  return (
    <input ref={props.inputRef!}  />
    )
}

export default MyComponent;
```


rqzheng2015:
Hi, @samijaber! Can you help me with the pr test failed issue? 

I've been tested on my local macOS and update snapshot, and it all worked well. But the test just failed on github action, any particular reason why? Thanks.


Same test failed issue happens on my other [PR](https://github.com/BuilderIO/mitosis/pull/1385) as well.

rqzheng2015:
Hi, @samijaber, do you have a moment to see the reason of why vitest snapshot not correctly generated? That would be great help for my MR, thanks.

samijaber:
Hi @rqzheng2015 , sorry for the delay. There's an issue with our test suite causing tests to not work properly.

It is causing all PRs to break unless additional instructions are followed. 

I will fix this issue first and then let you know to rerun your tests locally to get them fixed

packages/core/src/generators/solid/index.ts
```diff
@@ -61,14 +60,6 @@ function getContextString(component: MitosisComponent, options: ToSolidOptions)
   return str;
 }
 
-const getRefsString = (json: MitosisComponent, options: ToSolidOptions) =>
```
samijaber, line 64: this is not the right fix. you are getting rid of all `ref` declarations/

what we need to do is fix `getRefs` so it does not include `ref` attributes whose contents are javascript expressions. https://github.com/BuilderIO/mitosis/blob/5b04b032c6c8dffba89923f5c85c0f7f4c3573ab/packages/core/src/helpers/get-refs.ts#L9-L12

an extra check needs to be made here, maybe exclude anything with a `.` as a quick fix for now.

rqzheng2015: Okay, sorry for the late response. I have corrected it in the `getRefsString` method of `Solid.js` generator without removing the `getRefsString` method itself. Please check again, thanks.

rqzheng2015:
Hi @samijaber, I have modified my PR to fix the ref issue changes and vitest snapshot issues, please check when you are free, thanks.

samijaber (approved):

PR #1389, opened 2024-03-27
[v0] angular generator for gen2 angular sdk

sidmohanty11:
## Description

This PR handles:
- context for angular
- rename extension support for all files in mitosis
- handle props correctly so 'js'/random prop doesn't get added
- handle dynamic components using ngComponentOutlet and templates (supports children)
- handle spread and some other js stuff inside template
- handle template strings inside template using concatenation

**Jira**
https://builder-io.atlassian.net/browse/ENG-5475

### DIFF
--- a/.changeset/cool-dogs-appear.md
+++ b/.changeset/cool-dogs-appear.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Fix: reduce false positive props found in getProps
--- a/.changeset/cool-dogs-appear.md
+++ b/.changeset/cool-dogs-appear.md
@@ -2,4 +2,4 @@
 '@builder.io/mitosis': patch
 ---
 
-Fix: reduce false positive props found in getProps
+Fix: reduce false positive props found in `getProps`
--- a/.changeset/healthy-toys-wait.md
+++ b/.changeset/healthy-toys-wait.md
@@ -3,4 +3,4 @@
 '@builder.io/mitosis-cli': patch
 ---
 
-Feat: explicitImportFileExtension now adds/removes .js extensions to all import files
+Feat: remove all explicit `.js` import extensions when `explicitImportFileExtension` config is `false`
--- a/.changeset/healthy-toys-wait.md
+++ b/.changeset/healthy-toys-wait.md
@@ -1,0 +1,6 @@
+---
+'@builder.io/mitosis': patch
+'@builder.io/mitosis-cli': patch
+---
+
+Feat: explicitImportFileExtension now adds/removes .js extensions to all import files
--- a/packages/cli/src/build/helpers/context.ts
+++ b/packages/cli/src/build/helpers/context.ts
@@ -51,10 +51,48 @@ export const generateContextFile = async ({
         return contextToQwik()({ context });
       default:
         console.warn('Context files are not supported for this target. Outputting no-op');
-        return `
-        // Noop file
-        export default {};
-      `;
+        if (context.name === 'Builder') {
+          return `
+          import { Injectable } from '@angular/core';
+
+          @Injectable({
+            providedIn: 'root'
+          })
+          export default class BuilderContext {
+            content: any = null;
+            context: any = {};
+            localState: any = undefined;
+            rootState: any = {};
+            rootSetState: any = undefined;
+            apiKey: any = null;
+            apiVersion: any = undefined;
+            componentInfos: any = {};
+            inheritedStyles: any = {};
+            BlocksWrapper: string = 'div';
+            BlocksWrapperProps: any = {};
+
+            constructor() { }
+          }
+          `;
+        } else if (context.name === 'Components') {
+          return `
+          import { Injectable } from '@angular/core';
+
+          @Injectable({
+            providedIn: 'root'
+          })
+          export default class ComponentsContext {
+            registeredComponents: any = {};
+
+            constructor() { }
+          }
+
+          `;
+        } else {
+          return `// No op
+          export default {};
+          `;
+        }
     }
   }
 };
--- a/packages/cli/src/build/helpers/context.ts
+++ b/packages/cli/src/build/helpers/context.ts
@@ -1,13 +1,14 @@
 import {
+  MitosisConfig,
+  Target,
   checkShouldOutputTypeScript,
+  contextToAngular,
   contextToQwik,
   contextToReact,
   contextToSolid,
   contextToSvelte,
   contextToVue,
-  MitosisConfig,
   parseContext,
-  Target,
 } from '@builder.io/mitosis';
 import { readFile } from 'fs-extra';
 import { camelCase, last, upperFirst } from 'lodash';
@@ -49,50 +50,13 @@ export const generateContextFile = async ({
         });
       case 'qwik':
         return contextToQwik()({ context });
+      case 'angular':
+        return contextToAngular()({ context });
       default:
         console.warn('Context files are not supported for this target. Outputting no-op');
-        if (context.name === 'Builder') {
-          return `
-          import { Injectable } from '@angular/core';
-
-          @Injectable({
-            providedIn: 'root'
-          })
-          export default class BuilderContext {
-            content: any = null;
-            context: any = {};
-            localState: any = undefined;
-            rootState: any = {};
-            rootSetState: any = undefined;
-            apiKey: any = null;
-            apiVersion: any = undefined;
-            componentInfos: any = {};
-            inheritedStyles: any = {};
-            BlocksWrapper: string = 'div';
-            BlocksWrapperProps: any = {};
-
-            constructor() { }
-          }
-          `;
-        } else if (context.name === 'Components') {
-          return `
-          import { Injectable } from '@angular/core';
-
-          @Injectable({
-            providedIn: 'root'
-          })
-          export default class ComponentsContext {
-            registeredComponents: any = {};
-
-            constructor() { }
-          }
-
-          `;
-        } else {
-          return `// No op
+        return `// No op
           export default {};
-          `;
-        }
+        `;
     }
   }
 };
--- a/packages/core/src/__tests__/__snapshots__/angular.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/angular.test.ts.snap
@@ -39,11 +39,11 @@ import { Component, ViewChild, ElementRef, Input } from \\"@angular/core\\";
     \`,
   ],
 })
-export class MyBasicRefComponent {
-  @Input() showInput: any;
+export default class MyBasicRefComponent {
+  @Input() showInput!: any;
 
-  @ViewChild(\\"inputRef\\") inputRef: ElementRef;
-  @ViewChild(\\"inputNoArgRef\\") inputNoArgRef: ElementRef;
+  @ViewChild(\\"inputRef\\") inputRef!: ElementRef;
+  @ViewChild(\\"inputNoArgRef\\") inputNoArgRef!: ElementRef;
 
   name = \\"PatrickJS\\";
   onBlur = function onBlur() {
@@ -53,6 +53,26 @@ export class MyBasicRefComponent {
   lowerCaseName = function lowerCaseName() {
     return this.name.toLowerCase();
   };
+
+  useObjectWrapper(...args: any[]) {
+    let obj = {};
+    args.forEach((arg) => {
+      obj = { ...obj, ...arg };
+    });
+    return obj;
+  }
+
+  useObjectDotValues(obj: any): any[] {
+    return Object.values(obj);
+  }
+
+  useTypeOf(obj: any): string {
+    return typeof obj;
+  }
+
+  useJsonStringify(obj: any): string {
+    return JSON.stringify(obj);
+  }
 
   ngAfterContentChecked() {
     console.log(\\"Received an update\\");
@@ -107,11 +127,11 @@ import { CommonModule } from \\"@angular/common\\";
   standalone: true,
   imports: [CommonModule],
 })
-export class MyBasicRefComponent {
-  @Input() showInput: any;
+export default class MyBasicRefComponent {
+  @Input() showInput!: any;
 
-  @ViewChild(\\"inputRef\\") inputRef: ElementRef;
-  @ViewChild(\\"inputNoArgRef\\") inputNoArgRef: ElementRef;
+  @ViewChild(\\"inputRef\\") inputRef!: ElementRef;
+  @ViewChild(\\"inputNoArgRef\\") inputNoArgRef!: ElementRef;
 
   name = \\"PatrickJS\\";
   onBlur = function onBlur() {
@@ -121,6 +141,26 @@ export class MyBasicRefComponent {
   lowerCaseName = function lowerCaseName() {
     return this.name.toLowerCase();
   };
+
+  useObjectWrapper(...args: any[]) {
+    let obj = {};
+    args.forEach((arg) => {
+      obj = { ...obj, ...arg };
+    });
+    return obj;
+  }
+
+  useObjectDotValues(obj: any): any[] {
+    return Object.values(obj);
+  }
+
+  useTypeOf(obj: any): string {
+    return typeof obj;
+  }
+
+  useJsonStringify(obj: any): string {
+    return JSON.stringify(obj);
+  }
 
   ngAfterContentChecked() {
     console.log(\\"Received an update\\");
@@ -159,7 +199,7 @@ export const DEFAULT_VALUES = {
     \`,
   ],
 })
-export class MyBasicComponent {
+export default class MyBasicComponent {
   DEFAULT_VALUES = DEFAULT_VALUES;
 
   name = \\"Steve\\";
@@ -168,6 +208,26 @@ export class MyBasicComponent {
   }
   age = 1;
   sports = [\\"\\"];
+
+  useObjectWrapper(...args: any[]) {
+    let obj = {};
+    args.forEach((arg) => {
+      obj = { ...obj, ...arg };
+    });
+    return obj;
+  }
+
+  useObjectDotValues(obj: any): any[] {
+    return Object.values(obj);
+  }
+
+  useTypeOf(obj: any): string {
+    return typeof obj;
+  }
+
+  useJsonStringify(obj: any): string {
+    return JSON.stringify(obj);
+  }
 }
 
 @NgModule({
@@ -205,9 +265,29 @@ import { Component } from \\"@angular/core\\";
     </div>
   \`,
 })
-export class MyBasicForShowComponent {
+export default class MyBasicForShowComponent {
   name = \\"PatrickJS\\";
   names = [\\"Steve\\", \\"PatrickJS\\"];
+
+  useObjectWrapper(...args: any[]) {
+    let obj = {};
+    args.forEach((arg) => {
+      obj = { ...obj, ...arg };
+    });
+    return obj;
+  }
+
+  useObjectDotValues(obj: any): any[] {
+    return Object.values(obj);
+  }
+
+  useTypeOf(obj: any): string {
+    return typeof obj;
+  }
+
+  useJsonStringify(obj: any): string {
+    return JSON.stringify(obj);
+  }
 }
 
 @NgModule({
@@ -249,7 +329,7 @@ export const DEFAULT_VALUES = {
   standalone: true,
   imports: [CommonModule],
 })
-export class MyBasicComponent {
+export default class MyBasicComponent {
   DEFAULT_VALUES = DEFAULT_VALUES;
 
   name = \\"Steve\\";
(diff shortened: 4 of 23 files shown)


packages/core/src/helpers/get-props.ts
```diff
@@ -38,7 +38,7 @@ export const getProps = (json: MitosisComponent) => {
           if (prop.match(prohibitedKeywordRE)) {
             throw new Error(`avoid using JavaScript keyword as property name: "${prop}"`);
           }
-          props.add(prop);
+          prop !== 'js' && props.add(prop);
```
samijaber, line 41: I would prefer that we find out why a `js` prop gets created to begin with. having a workaround like this in Mitosis is not great.

I know Svelte also adds that prop for some reason: 

https://mitosis.builder.io/?outputTab=M4NwpgNgLmQ%3D&code=JYWwDg9gTgLgBAbzgVwM4FMCy6YEMAmueANChgCq5QDmOcAvnAGZQQhwDkAAgEbLAAbfOigA6YBAD0IYDAipgqDgG4AUKEiw4AIQEQAxgGtUzVuw6jRky5P1tIAO3QOYqSTz1G3Hg8dEDZdFEAK1QADxV1cGh4JGEIMBgAZWBqB1wBBlM2ThsbJmQHfRgJBzd4xJClNQ0YuBgATzB0RB1%2BIRFdXyyWHIsrG0bm73bhKABaHyMqyNqtIZakbVGRABEiXAAFVjATRl7zPKsFkcEx8bAd1Bm1VXQwzXgFuFWdgC8IJ22EkwBeNrOaw2312cAAZIhVHA4OkQOgAFxwVAwKDABzUNTQogo4B8GDoVCI3AOBpqei3NBYHAEDYACgQULgUFQ%2BkRDOh0Ls0ScLnITQRnAwUAAbiIOMRGfQJfQAJS3e6PODCJi4ZACeAFIolT5wJJ6GC0y4%2FRGvBIfL5XGWQ6FQHDIKAOOC0xnQgA8%2BGAwpdHORDQE6F%2BCHZHJDkDR%2BKgAFFRS5CZxVXJxd7ofR6MnEJZRLSAIRG3aiPiAqAAYU%2B%2BLCMFEwoyyCCdhc9xgAH5RKgGiAeBABABJfDgiHBkPQjiFjoTVD6jiIvPXWHoCVDhgytMhgB86ddXS86ehYCozhggZnBZWJbLjarNbr54rLbbHa7vbguBMyNR6JXi7ge5gAAtAwABvenYCKIhB4KIAAkCDHnO9AAZ%2Bi5TMYgaUpQNA4PSO4cpIABUuHYRyuFwAA6tAhhUBAhR9oUJSZL%2BMAwLs8KSJI1CyL%2ByA8KIXKSMsRbdgA8pIACOADuwCGJIiioLWbgAKwAAwAIwAOxwIozDAGE6D4KIhHQrhkgGeJkmIhUySpOkAjOl%2BQ7HqOYylg2FaXgItaiKwEDJHg%2BItgA2rBuBwgAugZVovgCY5boY%2FkhQudlKugKpqjA05XCeRbOeWlbVu5QReT5RDoAFBm7hlc4GSFz4mPx0WeLF8XYbKiFDvWOVHhljkiNljatdCkjriGrqSB6wpDXAcqqGmQA%3D

it only happens when generating Typescript code...

packages/core/src/generators/angular.ts
```diff
@@ -518,10 +716,29 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
+        return Object.values(obj);
+      }
+
+      useTypeOf(obj: any): string {
+        return typeof obj;
+      }
+
+      useJsonStringify(obj: any): string {
+        return JSON.stringify(obj);
+      }
```
samijaber, line 741: can you explain why we need these workarounds exactly?

packages/cli/src/build/helpers/context.ts
```diff
@@ -51,10 +51,48 @@ export const generateContextFile = async ({
         return contextToQwik()({ context });
       default:
         console.warn('Context files are not supported for this target. Outputting no-op');
-        return `
-        // Noop file
-        export default {};
-      `;
+        if (context.name === 'Builder') {
```
samijaber, line 54: we need to add a `case 'angular':` that takes the passed `context` and uses it to generate the context, like how we do with React/Svelte/Vue/others

packages/core/src/helpers/get-props.ts
samijaber: Since the `js` prop doesn't exist in the Mitosis JSON, there must be a helper function used by both the Svelte and Angular generators that is adding it by mistake.

Potentially this one? https://github.com/BuilderIO/mitosis/blob/5b04b032c6c8dffba89923f5c85c0f7f4c3573ab/packages/core/src/helpers/get-props.ts#L29

packages/core/src/generators/angular.ts
sidmohanty11: yes! the angular templating doesn't allow spread objects to be passed as props directly (we are constantly spreading a lot of props in our mitosis sdk which doesn't work with angular templating). So `useObjectWrapper` spreads out the obj and creates an object with all the properties. We are handling this above where we are providing binding values:
https://github.com/BuilderIO/mitosis/pull/1389/files#diff-73edddbc4186d31b9bfa0abf3140c76eca97e76e0b1612cdd9805515bdf63de7R380

this basically removes the `...` and keeps that object as a key and passes it as an arg to this `useObjectWrapper`. Wanted to ask should it be here or we should create an override for this? Or if you have any other suggestions on how we can handle these things (handle JS inside angular templating)

For example,
Screenshot 2024-04-03 at 11 18 25 AM

Others are pretty similar,
- useObjectDotValues : no Object is there inside angular templating
- same with typeof
- same with JSON

packages/core/src/helpers/get-props.ts
sidmohanty11: yes let me debug it, I added it here to just test out the build

packages/core/src/generators/angular.ts
```diff
@@ -166,8 +172,103 @@ export const blockToAngular = (
...
+        allProps += `${key}: useObjectWrapper(${temp}) `;
+      } else if (code.startsWith('Object.values')) {
+        let stripped = code.replace('Object.values', '');
+        allProps += `${key}: useObjectDotValues${stripped} `;
+      } else if (key.includes('-')) {
+        allProps += `'${key}': ${code}, `;
+      } else {
+        allProps += `${key}: ${code}, `;
+      }
+    }
```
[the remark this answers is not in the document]
sidmohanty11: yup yup let me refactor it! Thanks

packages/cli/src/build/helpers/context.ts
sidmohanty11: yup! will work on this

packages/core/src/generators/angular.ts
```diff
@@ -189,58 +381,17 @@ export const blockToAngular = (
...
-        str += `[${key}]="${code}" `;
-      }
-    }
+    str += stringifiedBindings;
     if (SELF_CLOSING_HTML_TAGS.has(json.name)) {
       return str + ' />';
     }
     str += '>';
 
-    if (needsToRenderSlots.length > 0) {
```
sidmohanty11, line 240: this is not needed anymore right? As we are already handling it above?

https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L149

Not sure if this is needed,

https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/generators/angular.ts#L223

packages/core/src/helpers/get-props.ts
sidmohanty11: okay found out that the regex took anything that contained `props.*`

and there was an item named: '../../types/builder-props.js', that also contained props.js so assumed js as a prop. I've handled this case by adding a ^ before props so that it only matches anything that starts with `props.*`

`/^props\s*\.\s*([a-zA-Z0-9_\$]+)/`

sidmohanty11: found out that it matches because we have a file called builder-props.js
```
prop {
  prop: 'js',
  item: '../../types/builder-props.js',
  matches: [ 'props.js' ],
  match: 'props.js'
}
```

how should we handle this? Should we check that if the item is not a file ref (doesn't contain any relative path?)

https://github.com/BuilderIO/mitosis/blob/5b04b032c6c8dffba89923f5c85c0f7f4c3573ab/packages/core/src/helpers/get-props.ts#L32

samijaber: yeah, this `get-props` logic is matching false positives. Instead of `traverse(json)` and checking for all strings, we should check for what we know to be code strings.

You might be able to reuse https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/helpers/plugins/process-code/index.ts, which processes every code block in the JSON.

Look at how `createCodeProcessorPlugin` is reused in the codebase. In your case, you dont' need to worry about the `codeType` since you will handle them all in the same way.

See https://github.com/BuilderIO/mitosis/blob/5b04b032c6c8dffba89923f5c85c0f7f4c3573ab/packages/core/src/parsers/jsx/state.ts#L120-L134 for an example

sidmohanty11: Thanks for this! I've updated the `getProps` function using `createCodeProcessorPlugin` and passing the json there please take a look

packages/core/src/helpers/get-props.ts
```diff
@@ -28,21 +28,20 @@ const prohibitedKeywordRE = new RegExp(
...
-      const matches = item.match(allPropsMatchesRegex);
-      if (matches) {
-        for (const match of matches) {
-          const prop = match.match(propsRegex)![1];
-          if (prop.match(prohibitedKeywordRE)) {
-            throw new Error(`avoid using JavaScript keyword as property name: "${prop}"`);
-          }
-          props.add(prop);
+
+  createCodeProcessorPlugin(() => (code) => {
```
samijaber, line 32: I would add a changeset for this: 

reduce false positive props found in `getProps`

packages/core/src/helpers/component-file-extensions.ts
```diff
@@ -133,3 +148,27 @@ export const getComponentFileExtensionForTarget = (args: Args): string => {
...
+    case 'react':
+    case 'reactNative':
+    case 'rsc':
+    case 'solid':
+    case 'stencil':
+    case 'marko':
+    case 'preact':
+    default:
+      return '.js';
+  }
```
samijaber, line 173: can we change this helper so it receives `explicitImportFileExtension` like `getComponentFileExtensionForTarget` does and uses that to decide whether to preserve the `.js` or not?

Also, let's add a changeset for this new logic

packages/core/src/generators/angular.ts
```diff
@@ -125,13 +125,183 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+      return ` #${code} `;
+    } else if (BINDINGS_MAPPER[key]) {
+      if (code.startsWith('{')) {
+        return `[${BINDINGS_MAPPER[key]}]="useObjectWrapper(${handleObjectBindings(code)})" `;
+      } else if (code.startsWith('Object.values')) {
+        let stripped = code.replace('Object.values', '');
+        return `[${BINDINGS_MAPPER[key]}]="useObjectDotValues${stripped}" `;
+      } else {
+        return `[${BINDINGS_MAPPER[key]}]="${code}" `;
+      }
```
samijaber, line 220: instead of making this its own condition that duplicates a bunch of code, I think you can have a check at the start of this entire function:

```ts
const keyToUse = BINDINGS_MAPPER[key] || key;
```

and then use `keyToUse` everywhere in the function.

packages/core/src/generators/angular.ts
```diff
@@ -125,13 +125,183 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+      } else if (code.includes('JSON.stringify')) {
+        let obj = code.match(/JSON.stringify\([^)]*\)/g);
+        return `[${key}]="useJsonStringify(${obj})" `;
+      } else if (code.includes('as')) {
+        const asIndex = code.indexOf('as');
+        const asCode = code.slice(0, asIndex - 1);
+        return `[${key}]="$any${asCode})"`;
+      } else {
+        return `[${key}]="${code}" `;
+      }
```
samijaber, line 239: We can deduplicate this logic with the same one in `handleNgOutletBindings`  in a function called `processCodeBlockInTemplate` or similar

packages/core/src/generators/angular.ts
samijaber: Thanks this is useful to know. Can you add this info in a comment for the `processCodeBlockInTemplate` function I suggested you add?

sidmohanty11: Yup!

packages/core/src/helpers/get-props.ts
sidmohanty11: Done!

packages/core/src/helpers/component-file-extensions.ts
sidmohanty11: I've added this for all the targets, should this be the case? Or do you think we should handle it differently?

packages/core/src/generators/angular.ts
sidmohanty11: Yup makes more sense! I've updated it

packages/core/src/generators/angular.ts
sidmohanty11: Yup, I've refactored this now

packages/core/src/helpers/component-file-extensions.ts
sidmohanty11: https://github.com/BuilderIO/mitosis/pull/1389/files#diff-887260c407e0b24d17b0276939ac20344220853cc466e813d5fa6984bf4578b9R161

.changeset/cool-dogs-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Fix: reduce false positive props found in getProps
```
samijaber, line 5:
```suggestion
Fix: reduce false positive props found in `getProps`
```

.changeset/healthy-toys-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/mitosis': patch
+'@builder.io/mitosis-cli': patch
+---
+
+Feat: explicitImportFileExtension now adds/removes .js extensions to all import files
```
samijaber, line 6:
```suggestion
Feat: when `explicitImportFileExtension` config is `false`, `.js` extensions from all import files are `removed`.
```

packages/core/src/generators/angular/index.ts
```diff
@@ -118,6 +117,179 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+      events += ` (${event})="${finalValue}" `;
+    } else if (code.includes('{')) {
+      allProps += `${key}: useObjectWrapper(${handleObjectBindings(code)}) `;
+    } else if (code.startsWith('Object.values')) {
+      let stripped = code.replace('Object.values', '');
+      allProps += `${key}: useObjectDotValues${stripped} `;
+    } else if (key.includes('-')) {
+      allProps += `'${key}': ${code}, `;
+    } else {
+      allProps += `${key}: ${code}, `;
```
samijaber, line 278: if you change `processCodeBlockInTemplate` so it only receives `code` as input:

```ts
const processCodeBlockInTemplate = (code: string) => {}
```

then we should be able to reuse it here and replace all of this code. Something roughly like:

```suggestion
    } else {
      const codeToUse = processCodeBlockInTemplate(code);
      const keyToUse = key.includes('-') ? `'${key}'` : key;
      allProps += `${keyToUse}: ${codeToUse}, `;
```

packages/core/src/generators/angular/index.ts
```diff
@@ -118,6 +117,179 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+        event = 'input';
+      }
+      // TODO: proper babel transform to replace. Util for this
+      const eventName = cusArgs[0];
+      const regexp = new RegExp(
+        '(^|\\n|\\r| |;|\\(|\\[|!)' + eventName + '(\\?\\.|\\.|\\(| |;|\\)|$)',
+        'g',
+      );
+      const replacer = '$1$event$2';
+      const finalValue = removeSurroundingBlock(code.replace(regexp, replacer));
```
samijaber, line 268: can we also deduplicate this in a `processEventBinding` so it's reused between this and `stringifyBinding`?

packages/core/src/generators/angular/index.ts
```diff
@@ -520,10 +705,42 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
+        return Object.values(obj);
+      }
+
+      useTypeOf(obj: any): string {
+        return typeof obj;
+      }
+
+      useJsonStringify(obj: any): string {
+        return JSON.stringify(obj);
+      }
```
samijaber, line 743: Can we only add these if they are needed/used by the bindings logic? They are adding a ton of noise to the snapshot tests and generated components

sidmohanty11: thank you so much for this suggestion! you're right we don't need them if we don't use them. I've updated the PR

packages/core/src/generators/angular/index.ts
sidmohanty11: you're right! thanks! let me make the changes

packages/core/src/generators/angular/index.ts
sidmohanty11: done!

packages/core/src/generators/angular/index.ts
```diff
@@ -118,6 +118,167 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+  } else if (code.startsWith('Object.values')) {
+    let stripped = code.replace('Object.values', '');
+    return `"useObjectDotValues${stripped}" `;
+  } else if (code.includes('JSON.stringify')) {
+    let obj = code.match(/JSON.stringify\([^)]*\)/g);
+    return `"useJsonStringify(${obj})" `;
+  } else if (code.includes(' as ')) {
+    const asIndex = code.indexOf('as');
+    const asCode = code.slice(0, asIndex - 1);
+    return `"$any${asCode})"`;
```
samijaber, line 179: I think this is missing an opening parenthesis?

```suggestion
    return `"$any( ${asCode} )"`;
```

packages/core/src/generators/angular/index.ts
```diff
@@ -118,6 +118,167 @@ const BINDINGS_MAPPER: { [key: string]: string | undefined } = {
...
+};
+
+const processCodeBlockInTemplate = (code: string) => {
+  // contains helper calls as Angular doesn't support JS expressions in templates
+  if (code.startsWith('{')) {
+    // Objects cannot be spread out directly in Angular so we need to use `useObjectWrapper`
+    return `"useObjectWrapper(${handleObjectBindings(code)})" `;
+  } else if (code.startsWith('Object.values')) {
+    let stripped = code.replace('Object.values', '');
+    return `"useObjectDotValues${stripped}" `;
```
samijaber, line 172: This is missing parentheses no?
```suggestion
    return `"useObjectDotValues( ${stripped} )" `;
```

samijaber: OH nevermind, the parentheses are included in `stripped`, since it was an `Object.values(x)` call before.

packages/core/src/__tests__/__snapshots__/angular.test.ts.snap
```diff
@@ -3752,17 +3764,7 @@ import { CommonModule } from \\"@angular/common\\";
...
-          height: '100%',
-          ...attributes?.style,
-          objectFit: fit,
-          objectPosition: position,
-          // Hack to get object fit to work as expected and
-          // not have the video overflow
-          borderRadius: 1
-        }\\"
-      [attr.key]=\\"video || 'no-src'\\"
+      [ngStyle]=\\"useObjectWrapper(attributes?.style, { width: '100%' }, { height: '100%' }, { objectFit: fit }, { objectPosition: position }, { borderRadius: 1 })\\"
```
samijaber, line 3767: IIUC `?.` syntax doesn't work in templates? I thought you had logic that replaced it, but it doesn't seem to be taking care of this scenario here.

Not a huge deal, just pointing it out

packages/core/src/generators/angular/index.ts
```diff
@@ -501,15 +680,15 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
         .map(([k, v]) => `${k}: ${v}`)
         .join(',')}
     })
-    export class ${json.name} {
+    export default class ${json.name} {
```
samijaber, line 683: I just realized: we need a changeset with a `minor` version upgrade explaining this is a breaking change for Angular output: all Angular component exports are now default exports instead of named exports.

samijaber: I went ahead and added that for you

samijaber (approved):

PR #1398, opened 2024-04-17
fix: add show `else` block to angular

sidmohanty11:
## Description

This PR adds support for `<Show when={/*condition*/} else={/*stuffs*/}> ... </Show>`'s else condition with a negation,

### Before
Screenshot 2024-04-23 at 6 45 01 PM

### After
Screenshot 2024-04-23 at 6 44 40 PM

**Jira**
https://builder-io.atlassian.net/browse/ENG-5719


packages/core/src/generators/angular/index.ts
```diff
@@ -327,12 +319,18 @@ export const blockToAngular = (
...
     }
     str += `<ng-container *ngIf="${condition}">`;
     str += json.children.map((item) => blockToAngular(item, options, blockOptions)).join('\n');
     str += `</ng-container>`;
+    // else condition
+    if (isMitosisNode(json.meta?.else)) {
+      str += `<ng-container *ngIf="!(${condition})">`;
+      str += blockToAngular(json.meta.else, options, blockOptions);
+      str += `</ng-container>`;
+    }
```
samijaber, line 333: I don't mind this workaround. 

If we realize it has limitations down the line, we can improve it so it generates:

```suggestion
    const elseConditionName = /*some unique identifier*/;

    str += `<ng-container *ngIf="${condition}; else ${elseConditionName}">`;
    str += json.children.map((item) => blockToAngular(item, options, blockOptions)).join('\n');
    str += `</ng-container>`;
    // else condition
    if (isMitosisNode(json.meta?.else)) {
      str += `<ng-template #${elseConditionName}>`;
      str += blockToAngular(json.meta.else, options, blockOptions);
      str += `</ng-template>`;
    }
```

samijaber (approved):

packages/core/src/generators/angular/index.ts
sidmohanty11: yes @samijaber this should be the best way to go

PR #1418, opened 2024-05-01
fix: `events` in dynamic components and pass them as a prop

sidmohanty11:
## Description

Pass `events` as props.

so for any type of event we pass it as:

```
inputs: { on{EventName}: nameOfFn }
```

we need to get them as an Input() inside the component and we can use it by calling them directly

### DIFF
--- a/packages/core/src/generators/angular/index.ts
+++ b/packages/core/src/generators/angular/index.ts
@@ -197,7 +197,10 @@ const processEventBinding = (key: string, code: string, nodeName: string, custom
   );
   const replacer = '$1$event$2';
   const finalValue = removeSurroundingBlock(code.replace(regexp, replacer));
-  return ` (${event})="${finalValue}" `;
+  return {
+    event,
+    value: finalValue,
+  };
 };
 
 const stringifyBinding =
@@ -217,7 +220,8 @@ const stringifyBinding =
     // TODO: proper babel transform to replace. Util for this
 
     if (keyToUse.startsWith('on')) {
-      return processEventBinding(keyToUse, code, node.name, cusArgs[0]);
+      const { event, value } = processEventBinding(keyToUse, code, node.name, cusArgs[0]);
+      return ` (${event})="${value}"`;
     } else if (keyToUse === 'class') {
       return ` [class]="${code}" `;
     } else if (keyToUse === 'ref') {
@@ -236,7 +240,6 @@ const stringifyBinding =
 
 const handleNgOutletBindings = (node: MitosisNode) => {
   let allProps = '';
-  let events = '';
   for (const key in node.bindings) {
     if (key.startsWith('"')) {
       continue;
@@ -250,9 +253,13 @@ const handleNgOutletBindings = (node: MitosisNode) => {
       allProps += `${key.replace('props.', '')}: ${code}, `;
     } else if (key.includes('.')) {
       // TODO: handle arbitrary spread props
-      allProps += `${key.split('.')[1]}: ${code},`;
+      allProps += `${key.split('.')[1]}: ${code}, `;
     } else if (key.startsWith('on')) {
-      events += processEventBinding(key, code, node.name, cusArgs[0]);
+      const { event, value } = processEventBinding(key, code, node.name, cusArgs[0]);
+      allProps += `on${event.charAt(0).toUpperCase() + event.slice(1)}: ${value.replace(
+        /\(.*?\)/g,
+        '',
+      )}, `;
     } else {
       const codeToUse = processCodeBlockInTemplate(code);
       const keyToUse = key.includes('-') ? `'${key}'` : key;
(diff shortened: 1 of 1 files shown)


samijaber (commented):
- needs a changeset
- is there any snapshot test that can show the change? if not can you add one?

sidmohanty11:
done, there is one bug here which i want to point out:

Screenshot 2024-05-02 at 6 28 28 PM

lets say we have any properties attached to the component, we are directly sending them as an input (notice: `hello: world`) and there is no straightforward way to know if its a string or any property/state inside the component. Any pointers on how we should handle this?

**Mitosis JSON**

Screenshot 2024-05-02 at 6 31 24 PM

samijaber:
properties (a.k.a. `<div hello="world" />`) are always raw strings: they cannot be javascript expressions. So whenever you generate them in a generator you must wrap them in quotes:

See examples

https://github.com/BuilderIO/mitosis/blob/4664fb4da2909e8cf29f2c931c17536c8fdf9771/packages/core/src/generators/svelte/blocks.ts#L294-L297

https://github.com/BuilderIO/mitosis/blob/4664fb4da2909e8cf29f2c931c17536c8fdf9771/packages/core/src/generators/vue/blocks.ts#L189-L191

packages/core/src/generators/angular/index.ts
```diff
@@ -250,9 +264,13 @@ const handleNgOutletBindings = (node: MitosisNode) => {
...
       // TODO: handle arbitrary spread props
-      allProps += `${key.split('.')[1]}: ${code},`;
+      allProps += `${key.split('.')[1]}: ${code}, `;
     } else if (key.startsWith('on')) {
-      events += processEventBinding(key, code, node.name, cusArgs[0]);
+      const { event, value } = processEventBinding(key, code, node.name, cusArgs[0]);
+      allProps += `on${event.charAt(0).toUpperCase() + event.slice(1)}: ${value.replace(
+        /\(.*?\)/g,
+        '',
+      )}, `;
```
samijaber, line 273: what is this regex replace doing exactly? why is it needed?

sidmohanty11: this removes the calling of a function: if we have `onClick()` we convert it to `onClick`

and pass it like `onClick: onClick` not `onClick: onClick()` which calls the function on initialization.

This strips away all the content within `(***)`

now that i think about this, the best way should be wrapping up the function and sending the entire thing as another function

for example,

like: `onClick: () => onClick()` but when i try this, this is actually not possible in angular

sidmohanty11:
> properties (a.k.a. `<div hello="world" />`) are always raw strings: they cannot be javascript expressions. So whenever you generate them in a generator you must wrap them in quotes:

got it! thanks, have updated them to use `'prop'` as the entire block is surrounded inside double quotes

```js
    str += `<ng-container *ngComponentOutlet="
      ${elSelector.replace('state.', '').replace('props.', '')};
      inputs: { ${allProps} };
      content: myContent;
      ">  `;
```

packages/core/src/generators/angular/index.ts
sidmohanty11: this will have a limitation,

if we want to directly pass in `onClick={() => console.log('hello')}`

this will strip out `('hello')`

and return `onClick: console.log` which we don't want

sidmohanty11: ahh so the above will not work at all, even if we inline the function we can't use `console.log` inside angular templates, so nvm. So basically with this approach we need to extract out functions into their own functions

Screenshot 2024-05-02 at 8 58 12 PM

packages/core/src/__tests__/data/angular/dynamic-component.raw.tsx
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  function onClick() {
+    console.log('hello');
+  }
+
+  return (
+    <obj.Component
+      hello="world"
+      onClick={() => onClick()}
```
samijaber, line 16: can you add another example that receives an argument? I want to see if that still works with your code

Something like:

```suggestion
      onClick={() => onClick()}
      onChange={(event) => onClick(event)}
```

sidmohanty11: sure!

this will not work I think as they are just passed as function and we cannot inline functions directly in angular template:

for example here we had to manually call the function with event: 

https://github.com/BuilderIO/builder/pull/3237/files#diff-12898be32207d680f53e79a6999052f6017b1b9f9069019aa778c9fdee649f07R17

samijaber: ok. that should get resolved once we cleanup the angular template workarounds as I described to you

sidmohanty11: yes!!

PR #1420, opened 2024-05-02
Full docs site [WIP]

steve8708:


samijaber (approved):

packages/docs/package.json
```diff
@@ -28,22 +28,38 @@
     "qwik": "qwik"
   },
   "devDependencies": {
+    "@builder.io/mitosis": "latest",
```
samijaber, line 31: this wont use the latest local version. is that on purpose?

```suggestion
    "@builder.io/mitosis": "workspace:*",
```

PR #1427, opened 2024-05-06
Fix: playground links

samijaber:
## Description

make links shareable

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -48,6 +48,7 @@
     "eslint": "^8.57.0",
     "eslint-plugin-qwik": "^1.5.2",
     "lodash-es": "^4.17.21",
+    "lz-string": "^1.5.0",
     "meilisearch-docsearch": "^0.6.0",
     "monaco-editor": "^0.48.0",
     "postcss": "^8.4.31",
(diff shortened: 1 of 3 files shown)


packages/docs/package.json
```diff
@@ -68,5 +68,8 @@
     "vite-plugin-replace": "^0.1.1",
     "vite-tsconfig-paths": "^4.2.1",
     "wrangler": "^3.0.0"
+  },
+  "dependencies": {
+    "lz-string": "^1.5.0"
```
steve8708, line 73: this should go into dev dependencies (idk why qwik does this but it expects that)

packages/docs/src/routes/playground/index.tsx
```diff
@@ -17,8 +18,13 @@ const defaultTopTab: OutputFramework = 'vue';
 const defaultBottomTab: OutputFramework = 'angular';
 const defaultInputTab = 'jsx';
 
+const decodeCode = (url: URL) => {
+  const code = url.searchParams.get('code');
+  return code ? lzString.decompressFromBase64(code) : defaultCode;
```
steve8708, line 23: old links would break with this right?

aka ideally we decode if needs decoding, but if is a url without decoding still would work

samijaber: the old fiddle compressed with lz-string. so the old links are currently broken without this.

steve8708: oh! that explains why i was seeing whacky stuff. cool

steve8708 (approved):

PR #1432, opened 2024-05-08
improve using Mitosis with libraries docs

steve8708:


packages/docs/src/routes/docs/using-libraries/index.mdx
```diff
@@ -10,13 +10,52 @@ You can use any JavaScript library in your Mitosis components just like you woul
...
+
+Just be careful of one gotcha - not all frameworks support directly referencing
+local variables in a template, so avoid referencing libraries directly in JSX as opposed to
+component level functions or [useStore](/docs/components/#state).
+
+```jsx
+import { kebabCase } from 'lodash';
+
+// ❌ won't work in all frameworks, such as Angular which uses string
+// templates that can't reference outside variables
```
samijaber, line 36: @steve8708 I have spoken to @sidmohanty11 about this and we have a permanent long-term fix to all of these Angular issues: 

Mitosis will automatically extract all JS expressions outside of the Angular template, and move them to class properties/methods instead so that there are no such restrictions on how to write your Mitosis source code.

samijaber: This fix will be made fairly soon, so I don't know if you want to have the docs updated as in this PR and then revert, or just hold off until the fix is live

steve8708: ah yes, but not implemented now right? I think this affects other frameworks too though like vue with the options API it won't "just work" either

samijaber: Correct, not implemented now.

Ah, correct. Vue Options API also has a bug where we don't include the import in the list of functions that the template can use.

steve8708: up to you then, i'm happy to keep or close. if you think angular will be here soon and vue options is not important to most ppl (maybe not?) and all the other frameworks support this (maybe they do?) i'm ok closing it

Issue #1390, opened 2024-04-02
bug: incorrect react native code generation


parwinders:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [X] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/?outputTab=E4UwhgxgLgcmUEsBuIg%3D&inputTab=FYZwHkA%3D&code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFAT0AD0iw4bEjmRN4JZADsAxjG6a4AWQCeAYQqRN6TTAAUYcmFQBKRBLhxtETangBtTRwQdAAaOAwYADkg9ABdOABeFAxsPHRbKmx0ADd0KmdJdzgodBhkKANbIo8AHhZgbIA%2Bao84GuBNMGQYFtbPVFQEhARevs8IJmgALmoSlioQ0Y9CQiW4bJwmZHQhwODVsdbvUwALHE0Acx2EWxzrGFcExvDS6ODb3JsOXCgrmA4Nlt0M4DmMeM1DgAJdBMSYAQjgAElPOdilo4B04AAldA4XRhABq2zCmAmwBYYWgcAAMsAAI6MFhwlo1Hj1JpFAoSQhAA%3D

### Expected Behaviour

when i provide <Input/> Tag in input i expected <TextInput/> in react native output

### Actual Behaviour

when i provide <Input/> Tag in input i get<View/> in react native output

samijaber:
I don't understand the bug. can you share the exact code that you expected and what you got instead?

parwinders:
> I don't understand the bug. can you share the exact code that you expected and what you got instead?

I am sorry what I meant is Input element is converted into View component in react native code which is incorrect... Correct behaviour is that <Input/> is mapped to <TextInput/>

parwinders:
I am sorry actually I tried giving it angle brackets like html tags but I think it is not possible... So I would restate the bug again.
Expected output is when I give Input component it maps to  TextInput in react native. But the current behaviour is that Input is mapped to View component which is definately wrong

samijaber:
[...]

PR #1435, opened 2024-05-08
fix: handle text input in react native generator

Othman2001:
## Description
adding text input in the react native generator, react native text input handling is missed, we should convert input tag from react to TextInput and add styles, onChange and so on

this is a fix for this issue:
https://github.com/BuilderIO/mitosis/issues/1390

### DIFF
--- a/packages/core/src/generators/react-native/index.ts
+++ b/packages/core/src/generators/react-native/index.ts
@@ -126,7 +126,7 @@ const PROCESS_REACT_NATIVE_PLUGIN: Plugin = () => ({
             if (node.bindings.onClick) {
               node.name = 'Pressable';
             }
-            if(node.bindings.onChange){
+            if(node.bindings.onChange && node.bindings.value){
               node.name = 'TextInput'
             }
           } else if (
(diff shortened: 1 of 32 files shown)


ShehabThndr (approved):

packages/core/src/generators/react-native/index.ts
```diff
@@ -126,13 +126,15 @@ const PROCESS_REACT_NATIVE_PLUGIN: Plugin = () => ({
             if (node.bindings.onClick) {
               node.name = 'Pressable';
             }
+            if(node.bindings.onChange && node.bindings.value){
+              node.name = 'TextInput'
+            }
```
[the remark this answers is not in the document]
Othman2001: @samijaber  yeah, makes sense i do it 👍

packages/core/src/generators/react-native/index.ts
```diff
@@ -120,19 +120,20 @@ const PROCESS_REACT_NATIVE_PLUGIN: Plugin = () => ({
         if (isMitosisNode(node)) {
           // TODO: handle TextInput, Image, etc
           if (isChildren({ node })) {
-            node.name = '';
```
samijaber, line 123: let's keep this logic here, removing it is unrelated to this PR

Othman2001: @samijaber yeah makes sense, I have reverted this line

Othman2001: I noticed that tests fails in marko.test.ts and solit.ts, i will check why this happened

Othman2001: @samijaber  I managed to fix the failed test by updating snapshots and re-running the tests locally to check on everything.

samijaber (approved):
Thanks for the contribution!

PR #1437, opened 2024-05-09
feat: `visuallyIgnoreHostElement` option for angular to apply `display: contents`

sidmohanty11:
## Description

In this PR, we are adding a new option called `visuallyIgnoreHostElement` which attaches `display: contents;` to the host angular element, which means they will now just be considered as decoration element and will not interfere with dom stylings in any way. 

> Why is this needed? Refer: https://stackoverflow.com/questions/34280475/remove-the-host-html-element-selectors-created-by-angular-component/52292365#52292365

> `display: contents` causes an element's children to appear as if they were direct children of the element's parent, ignoring the element itself. This can be useful when a wrapper element should be ignored when using CSS grid or similar layout techniques.

### Steps to reproduce

- Go to [angular.dev](https://angular.dev/) playground
- Use this code snippet,
```ts
import {Component} from '@angular/core';
import {bootstrapApplication} from '@angular/platform-browser';

@Component({
  selector: 'deep',
  standalone: true,
  template: `
  <style>
    .div {
      margin-left: auto;
      margin-right: auto;
    }
    </style>
    <div class="div"><ng-content></ng-content></div>
  `,
  //styles: [':host { display: contents; }']
})
export class Deep{}

@Component({
  selector: 'nested',
  standalone: true,
  template: `
    <deep>hello</deep>
  `,
  imports: [Deep],
  //styles: [':host { display: contents; }']
})
export class Nested{}

@Component({
  selector: 'hello',
  standalone: true,
  template: `
   <style>
    .div2 {
      display: flex;
      flex-direction: column;
    }
    </style>
  <div class="div2">
    <nested />
    <nested />
    <nested />
    </div>
  `,
  imports: [Nested],
  //styles: [':host { display: contents; }']
})
export class Hello{}

@Component({
  selector: 'app-root',
  standalone: true,
  template: `
    <hello></hello>
  `,
  imports: [Hello],
})
export class PlaygroundComponent {}

bootstrapApplication(PlaygroundComponent);
```
- inspect elements and notice the angular components act as elements on the DOM and interfere with margin:auto which should've been applied in normal case and angular components should just behave as fragments
- if you add the comments back (display: contents), content are now centered because angular elements don't hamper the stylings now

### DIFF
--- a/.changeset/witty-keys-flow.md
+++ b/.changeset/witty-keys-flow.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Feat: `visuallyIgnoreHostElement` option for angular to ignore angular components as elements in the DOM
--- a/.changeset/witty-keys-flow.md
+++ b/.changeset/witty-keys-flow.md
@@ -1,5 +1,5 @@
 ---
-'@builder.io/mitosis': patch
+'@builder.io/mitosis': minor
 ---
 
 Feat: `visuallyIgnoreHostElement` option for angular to ignore angular components as elements in the DOM
--- a/packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
@@ -33,6 +33,9 @@ import { Component, ViewChild, ElementRef, Input } from \\"@angular/core\\";
   \`,
   styles: [
     \`
+      :host {
+        display: contents;
+      }
       .input {
         color: red;
       }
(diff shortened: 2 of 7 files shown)


.changeset/witty-keys-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
```
samijaber, line 2: should make this a minor bump due to how different the generated code will be layout-wise

```suggestion
'@builder.io/mitosis': minor
```

samijaber (approved):

PR #1443, opened 2024-05-11
feat[Angular]: Allow passing explicit selectors

### LINKED ISSUE #1442: Angular selector does not follow angular selector names

Yes

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

[The default playground example using the Angular output](https://mitosis.builder.io/playground/)

### Expected Behaviour

```ts
@Component({
  selector: "my-component",
```

### Actual Behaviour

```ts
@Component({
  selector: "my-component, MyComponent",
```

https://angular.io/guide/styleguide#component-selectors

### Additional Information

Source Line: https://github.com/BuilderIO/mitosis/blob/1c800ed94e76f928a0c3ef834dce649711d59e6b/packages/core/src/generators/angular/index.ts#L640

See [angular style guide for components selector](https://angular.io/guide/styleguide#component-selectors)

> don't mind the specific prefix warning, the issues is regarding the ", Counter" 

image

I suggest we change make these changes

```diff
- selector: `'${kebabCase(json.name || 'my-component')}, ${json.name}'`,
+ selector: `'${kebabCase(json.name || 'my-component')}`,
```

seofernando25:
## Description

Adds an explicit, `selector` option, allowing to override the default,

```ts
selector: `'${kebabCase(json.name || 'my-component')}, ${json.name}'`
```

This doesn't quite work if you have more than one component per project since all of them would have the same name

```ts
selector: options.selector
        ? `'${options.selector}'`
       : `'${kebabCase(json.name || 'my-component')}, ${json.name}'`,
```

Another option could try is to keep as is and leave the warning there.

Closes #1442 

Also see: [angular's style guide](https://angular.io/guide/styleguide#component-selectors)
[Relevant PR that added it in the first place](https://github.com/BuilderIO/mitosis/issues/747)

EDIT: Ok I think I can see why it was added there in the first place...  might be wise to allow overriding the `selector`


seofernando25:
Oh this question might seem stupid but, what would be the command to update affected snapshots?
:sweat_smile:  I had just updated them manually but I'm not sure if this is optimal

seofernando25:
I'm having some issues running the snapupdates. 
Is it suppose that take this long 10+ minutes?

So far I've changed the expected results manually.

seofernando25:
https://github.com/BuilderIO/mitosis/issues/747

Relevant PR

packages/core/src/generators/angular/index.ts
```diff
@@ -637,11 +637,13 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
 
     // Preparing built in component metadata parameters
     const componentMetadata: Record<string, any> = {
-      selector: `'${kebabCase(json.name || 'my-component')}, ${json.name}'`,
+      selector: options.selector
+        ? `'${options.selector}'`
+        : `'${kebabCase(json.name || 'my-component')}, ${json.name}'`,
```
nmerget, line 642: I would suggest to remove the `${json.name}` as default.

If a users wants to overwrite the selector with non `kebabCase` they may use `options.selector`.

sidmohanty11 (commented):
just a suggestion, if we are adding a specific selector can we explicitly mention by using `useMetadata` instead?

```
useMetadata({
  angular: { selector: 'some-name' },
})
```

inside the component where we want to swap it? because `options` is for a full fledged project

seofernando25:
> can we explicitly mention by using useMetadata instead?

Sorry for taking so long to respond. 
I can try pushing some changes this Sunday. 

But I wouldn't mind if this PR was taken over.

nmerget:
@seofernando25 I can't select your fork as base: 
image

I add the `selector` option to useMetadata [here](https://github.com/nmerget/mitosis/tree/selector-template). Would you like me to open up a new PR from my fork? It's wired because I see other forks as base branches as well

image

seofernando25:
@nmerget oh yes please, feel free!

PR #1463, opened 2024-05-29
improve docs - fix spelling and grammar & clarify some points

szamanr:
fixes some spelling and grammar issues in the docs. also clarified the points about `<Show>` since it was confusing.

### DIFF
--- a/packages/docs/src/routes/docs/components/index.mdx
+++ b/packages/docs/src/routes/docs/components/index.mdx
@@ -182,7 +182,7 @@ Control flow in Builder is static like [Solid](https://github.com/ryansolid/soli
 
 ### Show
 
-Use `<Show>` for conditional logic. It takes a single `when` prop for the condition to match. When the condition is truthy, the children will render; otherwise, they will not. You can additionally provide an `else` prop for the content to be rendered in case the condition is falsy."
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/components/index.mdx
```diff
@@ -182,7 +182,7 @@ Control flow in Builder is static like [Solid](https://github.com/ryansolid/soli
 
 ### Show
 
-Use `<Show>` for conditional logic. It takes a singular `when` prop for a condition to match for. When the condition is truthy, the children will render, the `else` otherwise they will not.
+Use `<Show>` for conditional logic. It takes a single `when` prop for the condition to match. When the condition is truthy, the children will render; otherwise, they will not. You can additionally provide an `else` prop for the content to be rendered in case the condition is falsy."
```
samijaber, line 185:
```suggestion
Use `<Show>` for conditional logic. It takes a single `when` prop for the condition to match. When the condition is truthy, the children will render; otherwise, they will not. You can additionally provide an `else` prop for the content to be rendered in case the condition is falsy.
```

samijaber (approved):
Thanks for these doc fixes!

PR #1475, opened 2024-06-08
Add support for basic components in React native

Othman2001:
## Description

- What changes have you made:
Support button, touchable opacity, Image  and Linking
- Why you made them, and:
I think this change is very important to react native generator as it is currently is not supporting all core components.
- Any other useful context:
I think using a Button is much more convenient than pressable, that's why I migrated pressable usage to the button. 
the fiddle is not working properly with the Image component for some reason, but I managed to run the source code locally and tested all of my changes.

- [x] Format the codebase
- [x] Update all snapshots


packages/core/src/generators/react/blocks.ts
```diff
@@ -163,13 +163,45 @@ export const blockToReact = (
...
+    let onPress;
+    if (json.properties.href) {
+      onPress = `() => Linking.openURL('${json.properties.href}')`;
+    } else if (json.properties.onClick) {
+      onPress = `() => ${json.properties.onClick}`;
+    } else {
+      onPress = `() => {}`;
+    }
+    const childrenNodes: string = json.children
+      .map((item) => blockToReact(item, options, component, needsToRenderSlots))
```
samijaber, line 188: This code is crashing due to `needsToRenderSlots` being created after this line:


https://github.com/BuilderIO/mitosis/blob/215c3783cb2e22dbbf8912a6e6129640f87e119b/packages/core/src/generators/react/blocks.ts#L287

You need to move the initialization to the start of `blockToReact`

samijaber (changes requested):

packages/core/src/generators/react/blocks.ts
Othman2001: oh, makes sense i will fix this today 🙏

Othman2001: Done ✅

packages/core/src/generators/react-native/index.ts
```diff
@@ -124,8 +124,12 @@ const PROCESS_REACT_NATIVE_PLUGIN: Plugin = () => ({
           } else if (node.name.toLowerCase() === node.name && VALID_HTML_TAGS.includes(node.name)) {
             if (node.name === 'input') {
               node.name = 'TextInput';
-            } else if (node.bindings.onClick) {
-              node.name = 'Pressable';
```
samijaber, line 128: I just realized: we still want this condition in our logic. If an element has `onClick` but is not `a` or `button`, we need it to be `Pressable`. 

With your changes, it will become a `View` which cannot be pressed. This breaks `onClick` handling.

packages/core/src/generators/react/blocks.ts
```diff
@@ -172,17 +172,54 @@ export const blockToReact = (
...
 
+  if (options.type === 'native' && json.name === 'Image') {
+    const isUrl = json.properties.src
+      ? /^(http|https):\/\/[^ "]+$/.test(json.properties.src)
+      : false;
+    let source;
+    if (isUrl) {
+      source = `{ uri: '${json.properties.src}' }`;
+    } else {
+      source = `require('${json.properties.src}')`;
```
samijaber, line 189: What situation is this meant for exactly? generating `require` statements in our components does not seem like something we want to do in the general case.

samijaber: I think the generator should only do `{ uri: '${THE_SRC}' }`

packages/core/src/generators/react/blocks.ts
```diff
@@ -172,17 +172,54 @@ export const blockToReact = (
...
+        .map((item) => blockToReact(item, options, component, true, needsToRenderSlots))
+        .join('');
+    }
+    if (childrenNodes) {
+      return `<TouchableOpacity onPress={${onPress}}>
+      ${childrenNodes}
+    </TouchableOpacity>`;
+    }
+    return `<TouchableOpacity onPress={${onPress}}/> `;
+  }
```
samijaber, line 214: These code blocks have two issues:

- they are wiping out all props for `TouchableOpacity` and `Image`. See my comment in the snapshot.
- they only handle `json.properties.*`, not `json.bindings.*`. Properties are static, bindings are dynamic JS code blocks (`` vs ``). You need to handle both in the generator.

What you want to do instead is handle `src`, `href` and `onClick` in the for loops below which iterate over `json.bindings` and `json.properties`: https://github.com/BuilderIO/mitosis/blob/f6e6d95f06963a15902ad1eb9899918c1f8e3da6/packages/core/src/generators/react/blocks.ts#L202-L275

Let me know if you need any additional guidance around how to do that.

samijaber (changes requested):
Uncovered a few other issues with how the components are being generated.

packages/core/src/generators/react-native/index.ts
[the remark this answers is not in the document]
Othman2001: yeah sorry i missed this

packages/core/src/generators/react/blocks.ts
Othman2001: if image is local, it should be imported using require, if it is URI it should be` {uri:"url of image"}`

packages/core/src/generators/react/blocks.ts
Othman2001: i see, i will be working on this comments and let you know if i needed any help, Thank you a lot for your help.

packages/core/src/generators/react/blocks.ts
Othman2001: i think this is because of the nature of Image component in RN, if you are using local images, you generally need to use the require function instead of using a URI string directly. This is because local images are bundled with the app and require is used to resolve the image source at build time.

samijaber: I see. in the case of `json.bindings.src`, you will need  to use `uri:` for all cases because bindings are dynamic and cannot be parsed at compile-time.

packages/core/src/generators/react/blocks.ts
Othman2001: Hi @samijaber 
```
regarding the issue related to handling bindings and properties i think i managed to handle it by doing smth like this
    if (options.type === 'native' && json.name === 'Image') {
      let src;
      const imageSource = processBinding(String(json.bindings['src']?.code), options);
      if (json.properties.src) {
        const isUrl = json.properties.src
          ? /^(http|https):\/\/[^ "]+$/.test(json.properties.src)
          : false;
        let source;
        if (isUrl) {
          source = `{ uri: '${json.properties.src}' }`;
        } else {
          source = `require('${json.properties.src}')`;
        }
        return `<Image  source={${source}} />`;
      }
      if (imageSource) {
        src = `{ uri: ${imageSource} }`;
      }
      return `<Image source={${src}} />`;
    }
```
but regarding the issue of wiping out other props not sure what i need to do is there something like attributes for example where i can add all other props when i return Image like this, would appreciate help here plz 

`     return `<Image  source={${source}}  {...otherProps} />`;
`

Othman2001: @samijaber  I think this PR is ready for your review, Thank you.

Othman2001:
> Uncovered a few other issues with how the components are being generated.

@samijaber  This PR is ready for you review now, I have resolved your comments, can you please check it whenever you are free 
Thank you

packages/core/src/generators/react/blocks.ts
```diff
@@ -233,6 +236,45 @@ export const blockToReact = (
...
+        if (hrefValue) {
+          onPress = `() => Linking.openURL(${href})`;
+          str += ` onPress={${onPress}} `;
+          continue; // Skip further processing for 'href' in TouchableOpacity
+        }
+      } else if (key === 'target') {
+        // Remove 'target' prop handling for TouchableOpacity
+        continue;
+      }
+    }
```
samijaber, line 277: This has one issue: you are handling `json.bindings` and `json.properties` key/values in the same place. You need to handle them separately. This will require duplicating this logic.

What you need to do is:

- update this code block so that instead of referencing `json.properties.href`, `json.properties.src` etc. it is always _only_ referencing `json.bindings.*`.
- duplicate this block inside of the `for (const key in json.properties)` loop https://github.com/BuilderIO/mitosis/blob/f6e6d95f06963a15902ad1eb9899918c1f8e3da6/packages/core/src/generators/react/blocks.ts#L202-L220

and make sure that in this new case, it is only references `json.properties.*` key/values.

Othman2001: I see, I will work on this next week end hopefully 🙏  meanwhile I will move this pr to draft

Othman2001: @samijaber  Done, I have  moved  the logic to the two loops, for bindings and properties.
Sorry if were late resolving this.

Othman2001:
@samijaber  I have resolved the comments now, and separated the logic on two loops like we agreed, sorry if i were late.

samijaber (approved):
Amazing job. Well done, and thank you for contributing this big improvement to the RN generator 🎉

samijaber:
@Othman2001 you need to re-run the snapshots one last time for it to fix a merge conflict, and you're good to go!

Othman2001:
> @Othman2001 you need to re-run the snapshots one last time for it to fix a merge conflict, and you're good to go!

@samijaber  Done I have updated the snapshots, This PR is ready to go 🥳

Issue #1410, opened 2024-04-24
Angular: useObjectWrapper gets generated with incorrect input for object


lukjab71:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [X] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

Can't reproduce in fiddle, it probably uses previous version of Mitosis

### Expected Behaviour

Object passed as prop in mitosis jsx is generated correctly in Angular component. 

### Actual Behaviour

When I use object in mitosis jsx for some prop, example: 

`tooltipPlacementOptions={{
  "positions": ["left", "bottom", "bottom-left"],
  "offsets": {
      "left": {"topOffset": 0, "leftOffset": -24},
      "bottom": {"topOffset": 24, "leftOffset": 0},
      "bottom-left": {"topOffset": 24, "leftOffset": 0}
  }
}}`

in generated Angular code it gets generated as input to _useObjectWrapper_ as incorrect js object:

`[tooltipPlacementOptions]='useObjectWrapper({ "positions": ["left" }, { "bottom" }, { "bottom-left"] }, { "offsets": {
            "left": {
              "topOffset": 0 }, { "leftOffset": -24
            } }, { "bottom": {
              "topOffset": 24 }, { "leftOffset": 0
            } }, { "bottom-left": {
              "topOffset": 24 }, { "leftOffset": 0
            }
          } })'`
          

The first array is completely malformed and it makes the generated component impossible to use, as this throws compilation errors. 
          
On very initial investigation it looks like there is some bug in function _"handleObjectBindings"_ for Angular generator. 

### Additional Information

It looks like only versions 0.2+ are affected. I cannot reproduce this issue neither in Fiddle nor in version 0.1.7, which suggests that this is fairly new bug. "useObjectWrapper" is not generated at all in Fiddle or the mentioned older version.

samijaber:
@lukjab71 can you provide a link to a fiddle containing the broken input, even if the output isn't broken? This will help us reproduce with a concrete example.

samijaber:
cc @sidmohanty11

sidmohanty11:
[...]

PR #1490, opened 2024-07-05
fix: [Angular] fix logic used for parsing objects provided as input to useObjectWrapper

lukjab71:
## Description
This is propsed fix to issue: [1410](https://github.com/BuilderIO/mitosis/issues/1410)

- What changes you made:
The logic responsible for parsing an input to the method was incorrect. It used ',' comma as object separator. This gave weird results for any other constructs that would use comma, like arrays. Besides, this logic should probably be used only when input actually has spread operator in it. 

Regardless from object detection logic, I'm not convinced if logic responsible for concatenation `${spreadOutObjects.join(', ')}, ${otherObjs.join(', ')}` was correct. `spreadOutObjects` collects objects that use spread operator. `otherObjs` collect the rest of the objects from input. When we concatenate them using the initial logic, I believe we are completely mixing the order of objects, as there is no guarantee that `spreadOutObjects` in the original input, where at the begining. What if they were spread out somewhere else, in some nested object, of some complicated input? I believe this would either completely malform the object or in best case change the object structure. 

Now, about the propsed solution. As I understand, the whole purpose of this method is to get rid of `...` spread operator from syntax, as angular won't accept it. 

I tried to approach this problem in recursive way(to fix initial way of detecting objects - instead of using comma as object separator, try to figure start bracket and end bracket), but it is tricky, as regex are not good with recursive problems and even when I extracted recursion to separate method, it still was tricky for some complicated examples. And then I switched the approach - in the end, why do we need to detect separate objects? Our goal is to get rid of `...` from the string. So we can simply get rid of the `...` and then deal with potential consequences: 

 1) `someField: ...someObject` -> remove `...` and we are done
 2) `someField: {...someObject}` -> remove `...` and we are left with {object} result
 3) `someField: {...{nestedDeeper: true }}` -> remove `...` and we are left with {{ something: true }} result

I believe those cases will cover every possible variations on different levels of nesting. So after removing `...` we just have to cleanup what's left.

I prepared the PR using remarks from CONTRIBUTING.md file, however, I noticed that even without my changes there are tests failing in the `core` package(?).

### DIFF
--- a/.changeset/perfect-seas-bow.md
+++ b/.changeset/perfect-seas-bow.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': minor
+---
+
+fix logic responsible for parsing object provided to useObjectWrapper angular method
--- a/packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
@@ -5252,6 +5252,59 @@ export class MyComponentModule {}
 "
 `;
 
+exports[`Angular with Import Mapper Tests > jsx > Javascript Test > useObjectWrapper 1`] = `
+"import { NgModule } from \\"@angular/core\\";
+import { CommonModule } from \\"@angular/common\\";
+
+import { Component } from \\"@angular/core\\";
+
+@Component({
+  selector: \\"my-component, MyComponent\\",
+  template: \`
+    <div>
+      <CustomComponent
+        [tooltipPlacementOptions]='{
+          \\"positions\\": [\\"left\\", \\"bottom\\", \\"bottom-left\\"],
+          \\"offsets\\": {
+            \\"left\\": {
+              \\"topOffset\\": 0,
+              \\"leftOffset\\": -24
+            },
+            \\"bottom\\": {
+              \\"topOffset\\": 24,
+              \\"leftOffset\\": 0
+            },
+            \\"bottom-left\\": {
+              \\"topOffset\\": 24,
+              \\"leftOffset\\": 0
+            }
+          }
+        }'
+      >
+        test
+      </CustomComponent>
+    </div>
+  \`,
+  styles: [
+    \`
+      :host {
+        display: contents;
+      }
+    \`,
+  ],
+})
+export default class MyComponent {}
+
+@NgModule({
+  declarations: [MyComponent],
+  imports: [CommonModule, CustomComponentModule],
+  exports: [MyComponent],
+  bootstrap: [SomeOtherComponent],
+})
+export class MyComponentModule {}
+"
+`;
+
 exports[`Angular with Import Mapper Tests > jsx > Javascript Test > useTarget 1`] = `
 "import { NgModule } from \\"@angular/core\\";
 import { CommonModule } from \\"@angular/common\\";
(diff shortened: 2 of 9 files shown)


packages/core/src/__tests__/data/angular/use-object-wrapper.raw.tsx
```diff
@@ -0,0 +1,18 @@
+export default function MyComponent(props) {
+  return (
+    <div>
+      <CustomComponent
+        tooltipPlacementOptions={{
```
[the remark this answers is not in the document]
lukjab71: Yes, this is how I understood the initial logic for _useObjectWrapper_, that had the bug. Remove spread operator to input parameter, provided to _useObjectWrapper_ function, to avoid using spread syntax for angular components inputs. 

I updated the raw file and snaps with propsed example.

To be honest I don't fully understand the testing strategy in the repo. Even without my changes, when I run test command for _core_ package I get:

` Test Files  5 failed | 37 passed | 1 skipped (43)`
`      Tests  48 failed | 5914 passed | 3 skipped (5965)`

Those failures are completely unrelated to my changes. I hope that in general everything is working as expected.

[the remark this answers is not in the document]
sidmohanty11: thank you for explaining, looks good to me!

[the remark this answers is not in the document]
sidmohanty11: can you please try to pull the latest `main` and run `yarn test:update` from root and see if it works?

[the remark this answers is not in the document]
sidmohanty11: might have to lint the test file too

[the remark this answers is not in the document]
lukjab71: I've merged with current `main` branch, I've run `yarn test:update`. It went through with success. I used prettier and lint command.

packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
```diff
@@ -2300,7 +2291,14 @@ import { Component, Input } from \\"@angular/core\\";
   template: \`
     <video
       preload=\\"none\\"
-      [ngStyle]=\\"useObjectWrapper(attributes?.style, { width: '100%' }, { height: '100%' }, { objectFit: fit }, { objectPosition: position }, { borderRadius: 1 })\\"
+      [ngStyle]=\\"useObjectWrapper(
```
sidmohanty11, line 2294: im curious if this will actually work? this should fail right? because we cannot send params like this?

packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
```diff
@@ -5475,6 +5472,85 @@ export class MyComponentModule {}
...
+        )\\"
+        [val4]=\\"useObjectWrapper( attributes,
+          something,
+          anything: [1, 2, 3],
+          hello: 'hello',
+          attributes2
+        )\\"
+        [val5]=\\"useObjectWrapper( attributes,
+          something,
+          anything: [1, 2, 3],
```
sidmohanty11, line 5503: same thing here, will we be able to pass params like this?

packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
lukjab71: Good remark. My solution was too generic and it messed up the parameters. I updated it. It reuses both concepts - initial and mine. 

Spread operator is removed. If the object was presented in form of `{...obj}`, we replace it with `obj`. If there was some spread inside of object, we get rid of spread and deal with consequences. `{...obj1,test:true, ...obj2}` => obj1, {test: true}, obj2.

When I look at generated snaps, they seem to be ok.

packages/core/src/__tests__/__snapshots__/angular.mapper.test.ts.snap
lukjab71: Good remark. My solution was too generic and it messed up the parameters. I updated it. It reuses both concepts - initial and mine. 

Spread operator is removed. If the object was presented in form of `{...obj}`, we replace it with `obj`. If there was some spread inside of object, we get rid of spread and deal with consequences. `{...obj1,test:true, ...obj2}` => obj1, {test: true}, obj2.

When I look at generated snaps, they seem to be ok.

sidmohanty11: looks awesome!

packages/core/src/generators/angular/index.ts
```diff
@@ -124,29 +124,11 @@ const handleObjectBindings = (code: string) => {
...
-  if (temp.endsWith(', ')) {
-    temp = temp.slice(0, -2);
-  }
-
-  if (temp.startsWith(', ')) {
-    temp = temp.slice(2);
-  }
+  temp = temp.replace(/\{\s*\.\.\.(\w+)\s*}/g, '$1');
+  temp = temp.replace(/\.\.\./g, '');
+  temp = temp.replace(/(\s*\w+\s*:\s*((["'].+["'])|(\[.+])|([\w.]+)))(,|[\n\s]*)/g, `{ $1 },`);
```
sidmohanty11, line 131: could you please add some comments for each regex, explaining the purpose behind it for future reference?

sidmohanty11 (approved):
LGTM! Thank you very much for your contribution @lukjab71!

.changeset/perfect-seas-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': minor
+---
+
+fix logic responsible for parsing object provided to useObjectWrapper angular method
```
sidmohanty11, line 5:
```suggestion
Angular: Fix: `useObjectWrapper` logic to correctly handle spread and objects as arguments
```

.changeset/perfect-seas-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': minor
```
sidmohanty11, line 2:
```suggestion
'@builder.io/mitosis': patch
```

packages/core/src/generators/angular/index.ts
lukjab71: Ofc, I added comments for each step of this operation in [801cf8bf57](https://github.com/BuilderIO/mitosis/pull/1490/commits/801cf8bf57055a199d3b74f499c1df90aa17e537)

samijaber (approved):

PR #1503, opened 2024-07-17
fix: run `onMount` and `onUpdate` code in browser only in angular

sidmohanty11:
## Description

Moved all our initializations (handling generated state) to the `onInit` hook which runs inside `ngOnInit` now (runs both on ssr and csr) but `onMount` and `onUpdate` hooks which are in turn generated inside `ngOnInit` and `ngOnChanges` now run on CSR only (has a check if its running on browser), this is to prevent running any onUpdate/onMount hooks on the server side incase of angular.

Jira
https://builder-io.atlassian.net/browse/ENG-6195


packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
```diff
@@ -177,67 +178,12 @@ export class MyBasicForShowComponentModule {}
...
-    const change = this.myService.method(\\"change\\");
-    console.log(change);
-  };
-  node_0_div = null;
-
-  constructor(public myService: MyService) {
-    const hi = this.myService.method(\\"hi\\");
-    console.log(hi);
-  }
+"unknown: Unexpected token (1:8)
```
samijaber, line 181: something fishy is going on here

packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
```diff
@@ -2505,6 +2390,12 @@ export default class MyBasicForNoTagRefComponent {
   ngOnInit() {
     this.mergedInputs_0 = {};
 
+    this.mergedInputs_0 = {};
+
+    this.mergedInputs_0 = {};
+
+    this.mergedInputs_0 = {};
```
samijaber, line 2397: this duplication doesn't make sense. seems like we're looping over the wrong thing in the generator

packages/core/src/generators/angular/index.ts
```diff
@@ -569,14 +550,10 @@ const handleBindings = (
...
       if (item.bindings[key]?.type !== 'spread' && !key.startsWith('on')) {
         json.state[newBindingName] = { code: 'null', type: 'property' };
-        json.hooks['onMount'].push({
-          code: `state.${newBindingName} = ${item.bindings[key]!.code}`,
-          onSSR: false,
-        });
-        json.hooks['onUpdate'] = json.hooks['onUpdate'] || [];
-        json.hooks['onUpdate'].push({
-          code: `state.${newBindingName} = ${item.bindings[key]!.code}`,
-        });
```
samijaber, line 579: I'm confused why this change is part of this PR.

IIUC, by moving this code from `onMount` and `onUpdate` to `onInit`, you're essentially no longer making these `property` state fields reactive right? 

This feels like a different unrelated change to me. Can you explain why it's made here?

packages/core/src/generators/angular/index.ts
```diff
@@ -376,29 +376,10 @@ export const blockToAngular = ({
...
-        root.hooks.onUpdate &&
-        root.hooks.onUpdate.length > 0 &&
-        !root.hooks.onUpdate
-          .map((hook) => hook.code)
-          .join('')
-          .includes(inputsPropsStateName)
-      ) {
-        root.hooks.onUpdate.push({
-          code: `this.${inputsPropsStateName} = {${allProps}}`,
-        });
```
samijaber, line 400: Same question here on why this change is part of this PR

packages/core/src/generators/angular/index.ts
sidmohanty11: will revert

packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
sidmohanty11: yes found out that, this is because of not adding `\n` to onInit code it was mixing up everything: something like this - `{}state.something` have fixed this by adding new lines

packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
sidmohanty11: yes! have added a check to only add this state once if it's not added before (this happens because we loop through bindings and where this state is actually used, so this depends on however many times that specific binding is being used in the code).

samijaber (approved):
🙏🏽

PR #1511, opened 2024-07-25
feat: implement `trackBy` for angular

sidmohanty11:
## Description

This PR implements [`trackBy`](https://angular.dev/guide/directives#tracking-items-with-ngfor-trackby) in the angular generator

Jira
https://builder-io.atlassian.net/browse/ENG-6241

### DIFF
--- a/packages/core/src/generators/angular/index.ts
+++ b/packages/core/src/generators/angular/index.ts
@@ -866,6 +866,15 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
 
     stripMetaProperties(json);
 
+    const { components: dynamicComponents, dynamicTemplate } = traverseToGetAllDynamicComponents(
+      json,
+      options,
+      {
+        childComponents,
+        nativeAttributes: useMetadata?.angular?.nativeAttributes ?? [],
+      },
+    );
+
     const dataString = getStateObjectStringFromComponent(json, {
       format: 'class',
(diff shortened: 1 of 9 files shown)


packages/core/src/generators/angular/index.ts
```diff
@@ -340,9 +340,36 @@ export const blockToAngular = ({
...
-    str += `<ng-container *ngFor="let ${json.scope.forName} of ${json.bindings.each?.code}${
-      indexName ? `; let ${indexName} = index` : ''
-    }">`;
+    const forName = json.scope.forName;
+
+    // Check if "key" is present for the first child of the for loop
+    if (json.children[0].bindings && json.children[0].bindings.key?.code) {
+      const trackByFnName = `trackBy${
+        forName ? forName.charAt(0).toUpperCase() + forName.slice(1) : ''
+      }`;
```
samijaber, line 349: i think you need to use indices to create those function names. `forName` can be the same for multiple loops

CleanShot 2024-07-29 at 14 25 59@2x


in this case `forName` is `item` for both loops.

sidmohanty11: thanks @samijaber have handled this case by adding a local variable in `json.meta` which keeps the track of the index, please let me know if this approach looks fine to you

otherwise as `blockToAngular` is a recursive fn we can add another param and access it from there maybe?

samijaber (approved):

PR #1515, opened 2024-07-31
fix: state init in angular

sidmohanty11:
## Description

This PR handles state initialization for angular correctly by looking at the code value to check if it needs access to any prop value, if it does it initializes itself as `null` but on ngOnInit it correctly assigns the values (here `@Input` value is available for us to use

For only `state`access  inside of any other state angular generator works fine directly computing state, ex:
Screenshot 2024-07-31 at 7 41 41 PM
For `props.*` access it breaks if we directly try to access from `this`
Screenshot 2024-07-31 at 7 42 07 PM
Screenshot 2024-07-31 at 7 42 56 PM


samijaber (approved):
LGTM, don't forget the changesets

PR #1520, opened 2024-08-01
fix: state init in angular in correct sequence

sidmohanty11:
## Description

sequence matters first initialize state then binding states else we will not get the correct value as angular states are not reactive and not re-trigger any computed values

### DIFF
--- a/.changeset/rude-garlics-impress.md
+++ b/.changeset/rude-garlics-impress.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Fix: state initialization sequence in angular, init local states first so that bindings that depend upon them can use the correct values on ngOnInit
--- a/.changeset/rude-garlics-impress.md
+++ b/.changeset/rude-garlics-impress.md
@@ -2,4 +2,4 @@
 '@builder.io/mitosis': patch
 ---
 
-Fix: state initialization sequence in angular, init local states first so that bindings that depend upon them can use the correct values on ngOnInit
+Angular: Fix: state initialization sequence. Initialize states in `ngOnInit` first, followed by bindings that depend upon them.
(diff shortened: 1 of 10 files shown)


packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
```diff
@@ -5168,6 +5168,52 @@ export class MyComponentModule {}
...
+  ],
+})
+export default class MyComponent {
+  @Input() value;
+
+  val = null;
+  node_0_Comp = null;
+
+  ngOnInit() {
+    this.val = this.value;
```
sidmohanty11, line 5197: essentially here it used to be,

```
this.node_0_Comp = { ...this.val };
this.val = this.value;
```
where `this.val` was not yet set

packages/core/src/__tests__/__snapshots__/angular.state.test.ts.snap
```diff
@@ -11224,6 +11270,52 @@ export class MyComponentModule {}
...
+export default class MyComponent {
+  @Input() value!: any;
+
+  val = null;
+  node_0_Comp = null;
+
+  ngOnInit() {
+    this.val = this.value;
+    this.node_0_Comp = { ...this.val };
+  }
```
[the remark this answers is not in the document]
sidmohanty11: yup so we need all state init to happen at the top of onInit in sequence so that everything is initialized

.changeset/rude-garlics-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Fix: state initialization sequence in angular, init local states first so that bindings that depend upon them can use the correct values on ngOnInit
```
samijaber, line 5:
```suggestion
Angular: Fix: state initialization sequence. Initialize states in `ngOnInit` first, followed by bindings that depend upon them.
```

samijaber (approved):

PR #1522, opened 2024-08-06
feat: support destructuring of objects in Angular in HTML elements

sidmohanty11:
## Description

Adds support to destructure objects arbitrarily inside HTML elements in Angular.

https://builder-io.atlassian.net/browse/ENG-6294


packages/core/src/generators/angular/helpers.ts
```diff
@@ -19,6 +19,24 @@ export const HELPER_FUNCTIONS = (
...
+      if (!el) {
+        return;
+      }
+      Object.keys(value).forEach((key) => {
+        if (key.startsWith('on')) {
+          this.renderer.listen(
+            el,
+            key.replace('on', '').toLowerCase(),
+            value[key]
+          );
```
samijaber, line 33: we need to add logic to destroy these listeners:

- https://medium.com/claritydesignsystem/four-ways-of-listening-to-dom-events-in-angular-part-3-renderer2-listen-14c6fe052b59#e502
- https://stackoverflow.com/a/44454508


so we'll need to save the return value of every `this.renderer.listen()` call, and then call them on `ngOnDestroy`.

Not only that, we will also need to handle `ngOnChanges`, which AFAICS is going to keep adding listeners on every `ngOnChanges` call, quickly causing a memory leak.

samijaber (changes requested):
I tested this logic in the angular playground, and it doesn't actually set the attribute in `ngOninit` because the element reference is `undefined` at that point:

```ng
import {Component, ElementRef, Renderer2, SimpleChanges, ViewChild} from '@angular/core';
import {bootstrapApplication} from '@angular/platform-browser';

@Component({
  selector: 'app-root',
  standalone: true,
  template: `
    <div #elRef0>Hello world!</div>
  `,
})
export class PlaygroundComponent {
  @ViewChild("elRef0") elRef0!: ElementRef;

  attributes = {
    foo: 'bar',
    style: 'background-color: "red"'
  }
  
  setAttributes(el: HTMLElement, value: any) {
    console.log('setAttributes:', {el})
    if (!el) {
      return;
    }

    Object.keys(value).forEach((key) => {
      console.log('setting', {key, value})
      if (key.startsWith("on")) {
        this.renderer.listen(
          el,
          key.replace("on", "").toLowerCase(),
          value[key]
        );
      } else {
        this.renderer.setAttribute(el, key, value[key] ?? "");
      }
    });
  }

  constructor(private renderer: Renderer2) {}

  ngOnInit() {
    console.log('ngOninit:', {el: this.elRef0})
    this.setAttributes(this.elRef0?.nativeElement, this.attributes);
  }

  ngOnChanges(changes: SimpleChanges) {
      console.log('ngOnChanges:', {el: this.elRef0?.nativeElement})
    if (typeof window !== "undefined") {
      console.log('ngOnChanges (browser):', {el: this.elRef0?.nativeElement})
      this.setAttributes(this.elRef0?.nativeElement, this.attributes);
    }
  }
}

bootstrapApplication(PlaygroundComponent);
```

packages/core/src/generators/angular/helpers.ts
sidmohanty11: thank you! will add the check so that it destroys onUnMount.

> Not only that, we will also need to handle ngOnChanges, which AFAICS is going to keep adding listeners on every ngOnChanges call, quickly causing a memory leak.

oh yes, this will not be a mere replacement but a new event addition... but we still need a way to update event listeners on changes as any event binding may be updated while visual editing

packages/core/src/generators/angular/index.ts
```diff
@@ -1175,6 +1189,8 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       str = runPostCodePlugins({ json, code: str, plugins: options.plugins });
     }
 
+    delete json.compileContext?.angular;
```
samijaber, line 1196: btw, you don't need to `delete` anything since the `json` is always cloned at the start of each generator, and not used outside of it afterwards.

sidmohanty11: got it thanks, removed

samijaber (approved):
⚡

packages/core/src/generators/angular/helpers.ts
```diff
@@ -39,7 +39,7 @@ export const HELPER_FUNCTIONS = (
           target[key]
         ));
       } else {
-        this.renderer.setAttribute(el, key, target[key] ?? '');
+        this.renderer.setAttribute(el, key.toLowerCase(), target[key] ?? '');
```
sidmohanty11, line 42: this handles `autoPlay` -> `autoplay`, `playInline` -> `playinline` etc

packages/core/src/generators/angular/index.ts
```diff
@@ -455,10 +455,10 @@ export const blockToAngular = ({
           json.meta._spreadStateRef = (json.meta._spreadStateRef as number) + 1;
           makeReactiveState(root, name, `this.${name} = ${json.bindings[key]?.code};`);
           spreadCode = `this.${name}`;
-          changesCode = `changes['${spreadCode.replace('this.', '')}']?.currentValue || {}`;
+          changesCode = `changes['${spreadCode.replace('this.', '')}']?.currentValue`;
```
sidmohanty11, line 458: the `|| {}` only works if we are destructuring props directly but if we are using a state it wont be present in the `changes` object so we would need to re-send the object (by keeping this part `undefined`)

samijaber (approved):

PR #1525, opened 2024-08-07
Fix: Angular ngOnChanges

samijaber:
## Description

- What changes you made:

- Why you made them, and:

- Any other useful context:

### DIFF
--- a/packages/core/src/generators/angular/index.ts
+++ b/packages/core/src/generators/angular/index.ts
@@ -941,12 +941,18 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       return `const defaultProps = {${defalutPropsString}};\n`;
     };
 
+    const angularCoreImports = [
+      ...(outputs.length ? ['Output', 'EventEmitter'] : []),
+      ...(options?.experimental?.inject ? ['Inject', 'forwardRef'] : []),
+      'Component',
+      ...(domRefs.size || dynamicComponents.size ? ['ViewChild', 'ElementRef'] : []),
+      ...(props.size ? ['Input'] : []),
+      ...(dynamicComponents.size ? ['ViewChild', 'TemplateRef'] : []),
+      ...(!json.hooks.onUpdate?.length ? ['SimpleChanges'] : []),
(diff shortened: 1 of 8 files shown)


packages/core/src/generators/angular/index.ts
```diff
@@ -1076,7 +1082,7 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
       ${
         !json.hooks.onUpdate?.length
           ? ''
-          : `ngOnChanges() {
+          : `ngOnChanges(changes: SimpleChanges) {
```
sidmohanty11, line 1085: should we only add the type for TS? `changes${options.typescript ? ': SimpleChanges': ''}`

packages/core/src/generators/angular/index.ts
```diff
@@ -394,7 +394,7 @@ export const blockToAngular = ({
     if (options.state === 'class-properties') {
       const inputsPropsStateName = `mergedInputs_${hashCodeAsString(allProps)}`;
       root.state[inputsPropsStateName] = {
-        code: 'null',
+        code: '{}' + (options.typescript ? ' as any' : ''),
```
sidmohanty11, line 397: @samijaber what does this help with? Will it be useful for checking the object?

packages/core/src/generators/angular/index.ts
samijaber: good call! 🤦🏽

packages/core/src/generators/angular/index.ts
samijaber: `as any` helps with types, but switching `null` to `{}` resolves errors being thrown because angular is expecting an `inputs` object, and `null` is not allowed.

I can't remember exactly when it happens so unfortunately i can't reproduce it right now, but it would show up in my console as a thrown error.

sidmohanty11: oh yes, makes sense. Thank you for explaining. It doesn't happen for us because maybe I turned it off in ts config

sidmohanty11 (approved):

PR #1535, opened 2024-08-26
fix: state parser supports string literal keys

liamdebeasi:
### Bug

Mitosis does not support string literals when setting keys when setting the store. As a result, using them leads to [errors](https://mitosis.builder.io/playground/?code=JYWwDg9gTgLgBAbzgVwM4FMDKMCGN1wC%2BcAZlBCHAEQACARssADYAm6UAdMBAPQjAwIqYKioBuAFAT0AD0iw4bEjmRN4JZADsAxjG6a4AWQCeAYQqRN6TTAAUYcmFQBKRBLhxtETani%2B8BAC8KBjY0Oi2SFQkEBBUAFxwMFDIBITOku5wUOgwyFAGtlkeADwswABuAHzFHnAlwJpgyDC1dZ6oqIEICG3tnhBM0IlUOSxUADR9HoSE03AVOEyp3f74HDEQc%2F113qYAFjiaAObo3bboFdYwroFVcBgwAHI4IBGX1xy4UKcwHIvLdDObb9Hg1HYACXQTCGAEI4ABJTxHbJaOCaPCVaHGOCNOAAJXQOF0EzgADVUqTMFc1OhSQBFADuwAA1qSjiw4CAjjiQOFSFBXuhGdAWahYbUSjxytUshkJIQgA%3D%3D).

https://builder-io.atlassian.net/browse/ENG-6522

This PR adds support for parsing string literals as keys when setting the store.

Note that this fix only supports dot notation. For example, `store.foo` is supported, but `store['foo']` is not. There are other bugs related to keys that require bracket notion (such as `store['foo-bar']` that need to be fixed first.


packages/core/src/__tests__/data/string-literal-store.raw.tsx
```diff
@@ -0,0 +1,8 @@
+import { useStore } from '@builder.io/mitosis';
+
+export default function StringLiteralStore() {
+  /* prettier-ignore */
```
liamdebeasi, line 4: Otherwise prettier will remove the quotes around `foo`, defeating the purpose of the test

samijaber: could we actually have a key with dashes in this test to see if the rest of mitosis is able to handle it? something like `"bar-baz"`

samijaber: my guess is that some subtle things might break throughout the generators if we're trying to access `state['foo-baz']` in a component

liamdebeasi: Good idea! As we discussed internally, I've added a new failing test to note the `foo-bar` issue. I've also created an internal ticket to track work on this.

samijaber (approved):
thank you 👑

PR #1539, opened 2024-08-28
fix index in for loops between mitosis<->builder

steve8708:
also verify onMount works as expected

fixes: https://builder-io.atlassian.net/browse/ENG-6546

### DIFF
--- a/packages/core/src/generators/builder.ts
+++ b/packages/core/src/generators/builder.ts
@@ -1,3 +1,5 @@
+import { isMitosisNode } from '@/helpers/is-mitosis-node';
+import { replaceIdentifiers } from '@/helpers/replace-identifiers';
 import { BuilderContent, BuilderElement } from '@builder.io/sdk';
 import json5 from 'json5';
 import { attempt, mapValues, omit, omitBy, set } from 'lodash';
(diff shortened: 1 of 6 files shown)


packages/core/src/generators/builder.ts
```diff
@@ -91,6 +93,24 @@ const componentMappers: {
     return block;
   },
   For(_node, options) {
+    // rename `index` var to `state.$index`
+    const replaceIndex = (node: MitosisNode) => {
+      traverse(node).forEach(function (thing) {
+        if (isMitosisNode(thing)) {
+          for (const [key, value] of Object.entries(thing.bindings)) {
+            if (value?.code.includes('index')) {
+              thing.bindings[key]!.code = replaceIdentifiers({
```
samijaber, line 102: Any chance we can use `replaceNodes` instead? I've deprecated `replaceIdentifiers` because we frequently hit weird issues with it, and I added many patches to it before realizing it's never going to be a safe utility. The better approach is for `from` and `to` to be AST nodes themselves: it's a bit more work but reduces chances of things going wrong.

I think it'll look something like this:

```ts
replaceNodes({
    code: value.code,
    nodeMaps: [{
      from: types.identifier('index'),
      to: types.memberExpression(types.identifier('state'), types.identifier('$index')),
    }],
  });
```

Issue #1150, opened 2023-04-28
Prop types are not inferred for `lit` output


ohkimur:
### I am interested in helping provide a fix!

Yes

### Which generators are impacted?

- [ ] All
- [ ] Angular
- [X] HTML
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [X] Web components

### Reproduction case

[...]

PR #1546, opened 2024-09-04
[FIXUP] passing wrong object for transforming refs

newbie95:
Added fixup for the [issue](https://github.com/BuilderIO/mitosis/issues/1150)


samijaber:
please create pull requests using our pull request template. it includes instructions and steps that you need to follow for your PR to be mergable to the repo.

https://github.com/BuilderIO/mitosis/blob/main/.github/pull_request_template.md

samijaber (changes requested):
PR request template not followed: missing changeset and tests

packages/core/src/__tests__/test-generator.ts
```diff
@@ -244,6 +247,7 @@ const BASIC_TESTS: Tests = {
   useTarget,
   signalsOnUpdate,
   getterState,
+  ref,
```
samijaber, line 250: looking at the snapshot tests updates, I see many existing snapshots getting fixed by your PR.

You can remove this new `lit-ref.raw.tsx` test and its snapshot: there are existing tests that cover your fix! 🙏🏽

samijaber (changes requested):
Awesome, thanks! You can actually get rid of the new test: the existing snapshots cover your bugfix

packages/core/src/__tests__/test-generator.ts
newbie95: That's great to hear! I've removed the new test cases.

PR #1547, opened 2024-09-06
feat: refactor stencil generator to work properly with current version

### LINKED ISSUE #1251: bug: Error about `stenil`!

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [ ] React
- [ ] React-Native
- [ ] Solid
- [X] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

No link

### Expected Behaviour

```jsx
import { Component, h } from '@stencil/core'

@Component({
  tag: 'my-button',
})
export class MyButton {
  render() {
    return (
      <button>
        <slot />
      </button>
    )
  }
}
```

```jsx
import { Component, h } from '@stencil/core'

@Component({
  tag: 'my-example',
})
export class MyExample {
  render() {
    return <my-button>example</my-button>
  }
}
```

Outputs component content normally when used directly!

image

### Actual Behaviour

```jsx
import { Component, h } from '@stencil/core'

@Component({
  tag: 'my-button',
})
export default class MyButton {
  render() {
    return (
      <button>
        <slot />
      </button>
    )
  }
}
```

```jsx
import { Component, h } from '@stencil/core'

import MyButton from './button'

@Component({
  tag: 'my-example',
})
export default class MyExample {
  render() {
    return <MyButton>example</MyButton>
  }
}
```

`stencil` cannot export components using `export default` or the following error is reported!

image

`stencil` automatically registers the component, it doesn't need to be imported and can't be imported and used, it will report an error if it is imported and used!

image

### Additional Information

Let's see what kind of way it needs to be handled? I can contribute code!

nmerget:
## Description

Refactored [Stencil](https://stenciljs.com/) generator to work with current Version.
The previous generator didn't work with all tests/examples.

closes #1251


nmerget:
@samijaber Can I do anything else to support you with this?

packages/core/src/generators/angular/index.ts
```diff
@@ -824,16 +825,9 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
     const propsTypeRef = json.propsTypeRef !== 'any' ? json.propsTypeRef : undefined;
 
-    json.imports.forEach(({ imports }) => {
-      Object.keys(imports).forEach((key) => {
-        if (imports[key] === 'default') {
-          childComponents.push(key);
-        }
-      });
-    });
+    const childComponents: string[] = getChildComponents(json);
```
samijaber, line 830: can we add a changeset for angular? this impacts the angular snapshots

packages/core/src/generators/stencil/generate.ts
```diff
@@ -216,21 +249,29 @@ export const componentToStencil: TranspilerGenerator<ToStencilOptions> =
...
+        ${wrap ? '</Fragment>' : ''})
       }
     }
   `;
 
+    str = postCodeEvents(str, events);
+    str = postCodeChildComponentImports(str, childComponents);
+
+    // Some props aren't replaced by processBinding
+    str = str.replaceAll(`props.`, `this.`);
```
samijaber, line 273: we really want to avoid `replaceAll()`'s in the mitosis codebase. can you dig into what `props.*` instances aren't being handled properly by `processBinding`?

You can try using `replaceNodes` inside of `processBinding` (instead of, or in addition to `stripStateAndPropsRefs`). `stripStateAndPropsRefs` uses the deprecated `replaceIdentifiers` which has some issues. `replaceNodes` is more consistent

example of use: https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/parsers/svelte/helpers/post-process.ts#L45-L53

nmerget: Fixed. I just missed to use `processBinding` inside `blockToStencil`

packages/core/src/generators/stencil/collect-class-string.ts
```diff
@@ -20,11 +24,11 @@ export function collectClassString(
 
   const dynamicClasses: string[] = [];
   if (typeof json.bindings.class?.code === 'string') {
-    dynamicClasses.push(json.bindings.class.code as any);
+    dynamicClasses.push(processBinding(json.bindings.class.code, processBindingOptions));
```
samijaber, line 27: you shouldn't need to add `processBinding` all over a generator. This is exactly why I created the `CODE_PROCESSOR_PLUGIN` plugin: it groups all of this "code processing" logic into one centralized place. This means we will never forget to add our processing for one particular scenario.

I see we haven't added this plugin to stencil generator yet: we should do that and use `processBinding` there. Here's two examples in Svelte of how we use this plugin: https://github.com/BuilderIO/mitosis/blob/664f4a1342f91c1e96b088914420cd1074f8e35c/packages/core/src/generators/svelte/svelte.ts#L150-L185

samijaber (changes requested):

packages/core/src/generators/stencil/collect-class-string.ts
nmerget: Ahh perfect, I saw this but didn't know when/how I should use it. I'll change it 👍

nmerget: @samijaber I changed it, looks very clean <3

Maybe it's a good idea to structure every generator like this? 
image

This would help to find parts for a generator way better then using a single file for `toComponent`, `toBlock` and `plugins`. What do you think?

samijaber: yes, our newer generators have been improved to have a more flat folder structure, bu the old ones are still stuck in one big file. Typically we have `state.ts`, `blocks.ts`, `generator.ts` and `types.ts` and some helpers.

open to ideas. I like your improvement, we can stick with that 👍🏽

packages/core/src/generators/stencil/plugins/get-code-processor-plugins.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ToStencilOptions } from '@/generators/stencil/types';
+import { CODE_PROCESSOR_PLUGIN } from '@/helpers/plugins/process-code';
+
+export const getCodeProcessorPlugins = (
+  options: ToStencilOptions,
+  processBindingOptions: ProcessBindingOptions,
+) => {
+  return [
+    ...(options.plugins || []),
+    // Strip types from any JS code that ends up in the template, because Svelte does not support TS code in templates.
```
samijaber, line 11:
```suggestion
```

samijaber (approved):
Amazing! looking at the stencil snapshots, lokos like tons of stuff got fixed :)

PR #1532, opened 2024-08-19
fix: issue with "npm run build" for @builder.io/mitosis

nmerget:
## Description

Please provide the following information:

- What changes you made:

  - Solved issues running `npm run build` in workspace `builder.io/mitosis`

- Why you made them, and:

  - I tried to test locally, but it failed because of some Typescript errors.

- Any other useful context:

I don't know why I got those [issues](https://cloud.nx.app/runs/8SXytnRUCw), I thought it was about a new update caused by `IntelliJ` IDE, but I did a clean install with `VSCode` as well. 

Make sure to follow the PR preparation steps in [CONTRIBUTING.md](../CONTRIBUTING.md#preparing-your-pr) before submitting your PR:

- [x] format the codebase: from the root, run `yarn fmt:prettier`.
- [x] update all snapshots (in core & CLI): from the root, run `yarn test:update`

PR #1548, opened 2024-09-06
chore: add yarnPath to .yarnrc.yml if corepack didn't work

nmerget:
## Description

I found my problem from this [PR](https://github.com/BuilderIO/mitosis/pull/1532#pullrequestreview-2255356732):

The current yarn version requires [corepacl](https://yarnpkg.com/getting-started/install). 
But sadly I'm not able to enable it (I think it is because of my corporate computer). 
The fallback in this case is that yarn uses the global installed version which is `1.x.x`. 
This leads to massive problems while installing the repo.

I don't have any clue about `yarn` but a solution was to add the `yarnPath` to `.yarnrc.yml` to use the correct yarn version inside `.yarn/releases`. I don't know if there are any other issues if we add `yarnPath`, but I would be a little bit blocked developing locally.


samijaber (approved):
Good find 👍🏽

PR #1553, opened 2024-09-06
docs: update component styling docs

### LINKED ISSUE #1250: Readme need more info about using CSS
For mitosis, the docs seem to clearly go over how your template and state etc will be set up. But besides brief mention of a CSS prop in a blog post, I have found very little about utilizing any sort of CSS patterns or files in mitosis. Seeing as how that's kind of an essential part of HTML and CSS, The readme or docs really need updates regarding this. A few more examples would go a long way.

I understand the idea of that, "just uses a CSS or style prop" seems sufficient. But styling is super important too and _any_ reference to how mitosis might integrate with any slightly more complicated styling patterns would be nice.

nmerget:
## Description

closes #1250

Add some additional explanation for styling Mitosis components


packages/docs/src/routes/docs/components/index.mdx
```diff
@@ -170,7 +170,49 @@ export default function ResponsiveExample() {
...
+
+```tsx
+/* ./src/my-component/my-component.lite.tsx */
+export default function CSSExample() {
+  return <div class="my-component" />;
+}
+```
+
+The advantage of this is, that you can rely on the CSS ecosystem (sass, less, etc.).
+But you might encounter some problems when you want to use web-components with shadowDOM.
```
samijaber, line 215: I don't fully understand this note. What's a "standalone component"?

nmerget: I meant `headless`, I don't know why I wrote `standalone` in the first place.

Changed it.

packages/docs/src/routes/docs/components/index.mdx
```diff
@@ -170,7 +170,49 @@ export default function ResponsiveExample() {
...
```
samijaber, line 215:
```suggestion
Another approach is to develop your components without a fixed style inside the Mitosis component, a.k.a. "headless components".

You can do so by loading the styles inside the consuming webapp via a bundler (vite, webpack, etc.) or with a head stylesheet `<link rel="stylesheet" ...`.
You can then use `class` or `data-*` attributes inside your Mitosis component to apply those styles:

```tsx
/* ./src/my-component/my-component.lite.tsx */
export default function CSSExample() {
  return <div class="my-component" />;
}
```

The advantage of this is that you can rely on the CSS ecosystem (sass, less, etc.).

Note: you might encounter some problems with this approach if you want to use web components with shadowDOM.
```

samijaber (approved):

PR #1557, opened 2024-09-09
docs: note for useRef hook to avoid naming it "ref" -> conflict with vue

nmerget:
## Description

Add a note in documentation for `useRef`:
Avoid using `const ref = useRef(null);`, beacuse it might be a conflict with `Vue`.


samijaber (approved):
Nice, thanks!

Even better would be to:
- create an eslint rule that forbids creating any property with the name `ref`.
- OR improve the vue generator so that it renames any properties called `ref` (to `_ref`, `ref_state` or anything else to avoid a name collision)

PR #1558, opened 2024-09-09
fix: issues for vue composition api

nmerget:
## Description

There were some issues with vue `option.api="composition"`:

- `ref` wasn't imported when using `useRef` hook without using `useState`
- Composition api always use [`ref()`](https://vuejs.org/api/reactivity-core.html#ref) wihtout any `class` -> we don't need `this.`, but we always use `.value`
- `ref` could be `null` for `useRef` see: https://vuejs.org/guide/essentials/template-refs.html#accessing-the-refs


packages/core/src/generators/vue/compositionApi.ts
```diff
@@ -51,8 +51,10 @@ export function generateCompositionApiScript(
     functions: false,
     getters: false,
     format: 'variables',
-    valueMapper: (code, _, typeParameter) =>
-      isTs && typeParameter ? `ref<${typeParameter}>(${code})` : `ref(${code})`,
+    valueMapper: (code, _, typeParameter) => {
+      const cleanCode = code.replaceAll('this.', ''); // Composition api isn't a class we don't need "this." here
```
samijaber, line 55: Is this meant to handle situations like the one in your test?

```ts
 const state = useStore<MyStore>({
    _id: 'abc',
    _active: false,
    _do: (id?: string) => {
      this._active = !!id;
      if (props.onChange) {
        props.onChange(this._active);
      }
    },
  });
```

I always use `state.*` instead of `this.*` in these scenarios. 


I think having a `replaceAll()` only in the vue composition generator to be too narrow. A better place would be here: https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/parsers/jsx/element-parser.ts#L194-L267

this is where all bindings are parsed. It might make more sense to add some logic that will replace any `this.*` with `state.*` within a binding's code block. You can use `replaceNodes` to do this intelligently (by manipulating the AST) instead of a regex `replaceAll()`, which is error prone.

.changeset/perfect-cups-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+fix: issues with vue composition api
```
samijaber, line 5: I know this is tedious, but could you spell out each individual fix. This can be helpful for many folks when updating mitosis and seeing changes in their generated code.

packages/core/src/generators/vue/compositionApi.ts
```diff
@@ -61,6 +63,9 @@ export function generateCompositionApiScript(
     getters: false,
     functions: true,
     format: 'variables',
+    /*    valueMapper: (code) => {
+      return code.replaceAll('this.', ''); // Composition api isn't a class we don't need "this." here
+    },*/
```
samijaber, line 68:
```suggestion
```

packages/core/src/generators/vue/compositionApi.ts
```diff
@@ -51,8 +51,10 @@ export function generateCompositionApiScript(
     functions: false,
     getters: false,
     format: 'variables',
-    valueMapper: (code, _, typeParameter) =>
-      isTs && typeParameter ? `ref<${typeParameter}>(${code})` : `ref(${code})`,
+    valueMapper: (code, _, typeParameter) => {
+      const cleanCode = code.replaceAll('this.', ''); // Composition api isn't a class we don't need "this." here
+      return isTs && typeParameter ? `ref<${typeParameter}>(${cleanCode})` : `ref(${cleanCode})`;
+    },
```
samijaber, line 57:
```suggestion
```

packages/core/src/__tests__/__snapshots__/parse-jsx.test.ts.snap
```diff
@@ -3476,6 +3476,95 @@ exports[`Parse JSX > Javascript > InputParent 1`] = `
...
+  "name": "NestedStore",
+  "refs": {},
+  "state": {
+    "_id": {
+      "code": "\\"abc\\"",
+      "propertyType": "normal",
+      "type": "property",
+    },
+    "_messageId": {
+      "code": "this._id + \\"-message\\"",
```
samijaber, line 3559: it doesn't seem like your `replaceNodes` logic works properly. this should be `state.id`.

packages/core/src/__tests__/__snapshots__/svelte.test.ts.snap
```diff
@@ -1092,6 +1092,19 @@ exports[`Svelte > jsx > Javascript Test > InputParent 1`] = `
 />"
 `;
 
+exports[`Svelte > jsx > Javascript Test > NestedStore 1`] = `
+"<script>
+  let _id = \\"abc\\";
+
+  let _messageId = this._id + \\"-message\\";
```
samijaber, line 1099: another indicator that the `replaceNodes` isn't doing its job

samijaber (changes requested):
your `replaceNodes` call isn't working properly. 🤔 

might want to add some unit tests in https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/helpers/replace-identifiers.test.ts

to figure out why that is.

packages/core/src/generators/vue/compositionApi.ts
samijaber: oh sorry i meant to suggest reverting

```suggestion
    valueMapper: (code, _, typeParameter) =>
      isTs && typeParameter ? `ref<${typeParameter}>(${code})` : `ref(${code})`,
```

packages/core/src/helpers/replace-identifiers.ts
```diff
@@ -220,6 +216,9 @@ export const replaceNodes = ({
   };
 
   return babelTransformExpression(code, {
+    ThisExpression(path) {
+      searchAndReplace(path);
+    },
```
samijaber, line 221: 🙌🏽

samijaber (approved):
Amazing 🎉

samijaber:
Looks like you have some issues with your Solid and Marko tests: 
CleanShot 2024-10-02 at 10 46 20@2x
CleanShot 2024-10-02 at 10 47 28@2x


when tests go from throwing to no longer throwing, if `yarn test:update` doesn't handle this properly, what I tend to do is delete the snapshot file and recreate it altogether.

nmerget:
> Looks like you have some issues with your Solid and Marko tests: CleanShot 2024-10-02 at 10 46 20@2x CleanShot 2024-10-02 at 10 47 28@2x
> 
> when tests go from throwing to no longer throwing, if `yarn test:update` doesn't handle this properly, what I tend to do is delete the snapshot file and recreate it altogether.

@samijaber somehow I'm not able to get any different snapshot, even if I delete them :(

samijaber:
@nmerget very weird. I'm not sure why this is happening. I ran `yarn test:update` in `packages/core` on my end and it seems to have updated them on the first try.

nmerget:
Maybe some `yarn` cache or `windows` problem again ^.^

mfranzke:
> Maybe some `yarn` cache or `windows` problem again ^.^

get a Mac

PR #1565, opened 2024-09-26
Remove on prefix on Angular Events

bastiW:
Remove on prefix on Angular Events

## Description

- What changes you made: Remove on prefix on Angular Events
- Why you made them, and: Angular strictly recommends not to use prefix "on" on EventEmitters. [Rule 05-16: Don't Prefix Output Properties](https://angular.dev/style-guide#style-05-16)

### DIFF
--- a/.changeset/flat-days-protect.md
+++ b/.changeset/flat-days-protect.md
@@ -1,5 +1,0 @@
----
-'@builder.io/mitosis': patch
----
-
-Builder: include Column widths in generator
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @builder.io/mitosis-cli
 
+## 0.5.38
+
+### Patch Changes
+
+- Updated dependencies [469394f]
+  - @builder.io/mitosis@0.5.38
+
 ## 0.5.37
 
 ### Patch Changes
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/mitosis-cli",
-  "version": "0.5.37",
+  "version": "0.5.38",
   "description": "mitosis CLI",
   "types": "build/types/types.d.ts",
   "bin": {
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 0.5.38
+
+### Patch Changes
+
+- 469394f: Builder: include Column widths in generator
+
 ## 0.5.37
 
 ### Patch Changes
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -22,7 +22,7 @@
     "name": "Builder.io",
     "url": "https://www.builder.io"
   },
-  "version": "0.5.37",
+  "version": "0.5.38",
   "homepage": "https://github.com/BuilderIO/mitosis",
   "main": "./dist/src/index.js",
   "exports": {


packages/core/src/generators/angular/index.ts
```diff
@@ -865,13 +869,16 @@ export const componentToAngular: TranspilerGenerator<ToAngularOptions> =
...
       if (options?.experimental?.outputs) {
-        return options?.experimental?.outputs(json, variableName);
+        return options?.experimental?.outputs(json, outputName);
       }
-      return `@Output() ${variableName} = new EventEmitter()`;
+
+      return `@Output() ${removeOnFromAngularOutputEvent(outputName)} = new EventEmitter()`;
     });
 
+    outputs.forEach(console.log);
```
bastiW, line 880: remove

packages/core/src/helpers/strip-state-and-props-refs.ts
```diff
@@ -60,7 +61,7 @@ export const DO_NOT_USE_VARS_TRANSFORMS = (
   outputVars?.forEach((_var) => {
     // determine expression edge cases onMessage( to this.onMessage.emit(
     const regexp = '(^|\\s|;|\\()(props\\.?)' + _var + '\\(';
-    const replacer = '$1' + context + _var + '.emit(';
+    const replacer = '$1' + context + removeOnFromAngularOutputEvent(_var) + '.emit(';
```
nmerget, line 64: You shouldn't import a function from a generator inside `helpers`. Right now `DO_NOT_USE_VARS_TRANSFORMS` is used by `generators/html` as well

bastiW: So better do that mainipulation inside 

processAngularCode ?

nmerget: Yes do the manipulation inside `packages/core/src/generators/angular/*` there is a `helper.ts` as well

bastiW:
@nmerget Thanks a lot for your feedback. I have some more questions: 

1. I now copy pasted the code from the [packages/core/src/helpers/strip-state-and-props-refs.ts](https://github.com/BuilderIO/mitosis/pull/1565/files#diff-938da9e8c356cb3df3840ddd6dee3b2224739f697572e72eec2085b0f9bf2ed6) to the pipe transform inside [packages/core/src/generators/angular/index.ts](https://github.com/BuilderIO/mitosis/pull/1565/files#diff-9ad88ca883760bd275260d6d15f58dfc5093ab14135e005829fb8a1c351deff3) and removed the `on` in the code.
This is the easiest solution as the code has just a string in the JSON. I am thinking about better alternatives. We could make a JSON AST out of the JS Code String. Is there already an existing function for that?

2. In the Qwik E2E Test I get an error. Unfortunately, I know too little about Qwik to fix it.

nmerget:
> @nmerget Thanks a lot for your feedback. I have some more questions:
> 
> 1. I now copy pasted the code from the [packages/core/src/helpers/strip-state-and-props-refs.ts](https://github.com/BuilderIO/mitosis/pull/1565/files#diff-938da9e8c356cb3df3840ddd6dee3b2224739f697572e72eec2085b0f9bf2ed6) to the pipe transform inside [packages/core/src/generators/angular/index.ts](https://github.com/BuilderIO/mitosis/pull/1565/files#diff-9ad88ca883760bd275260d6d15f58dfc5093ab14135e005829fb8a1c351deff3) and removed the `on` in the code.
>    This is the easiest solution as the code has just a string in the JSON. I am thinking about better alternatives. We could make a JSON AST out of the JS Code String. Is there already an existing function for that?
> 2. In the Qwik E2E Test I get an error. My IDE tells me I should change
> 
> ```tsx
>  <EventChild
>         onConfirm$={$((name) => _onConfirm(props, state, name))}
>         onCancel$={$(() => _onCancel(props, state))}
>       ></EventChild>
> ```
> 
> to
> 
> ```tsx
>  <EventChild
>         onConfirm={$((name) => _onConfirm(props, state, name))}
>         onCancel={$(() => _onCancel(props, state))}
>       ></EventChild>
>       
> ```
> 
> Unfortunately, I know too little about Qwik to fix it. For some reason, the route `/event-listener/` is not shown in the overview when I start the Qwik project (I think this is because the qwik build failed). Furthermore, I don't know if removing the `$` is correct.

1. I think just copying won't be the best in this case, like it says `DO_NOT_USE_VARS_TRANSFORMS`. Probably the idea by using AST at some point might be the way to go. I just know that @samijaber tries to get rid of the regex replacements and wants to rely on Babel. 
2. Am I wrong or are the code snippets identical? I don't know much about qwik either, sorry

bastiW:
1) I would need a hint from @samijaber  how to do that. Or we do that with a later PR. 
2) IDE tells it is `onConfirm=` instead of `onConfirm$=`

nmerget:
> 1. I would need a hint from @samijaber  how to do that. Or we do that with a later PR.
> 2. IDE tells it is `onConfirm=` instead of `onConfirm$=`

Maybe this is a good hint I got in my [PR](https://github.com/BuilderIO/mitosis/pull/1547#discussion_r1777572330): 

https://github.com/BuilderIO/mitosis/blob/664f4a1342f91c1e96b088914420cd1074f8e35c/packages/core/src/generators/svelte/svelte.ts#L150-L185

` CODE_PROCESSOR_PLUGIN` uses `processOnEventHooksPlugin()` which uses `traverseNodes` transforms the stuff before generation. I guess this could be the way to go

bastiW:
@nmerget  but the onEvent is part of 

When I read the comment of component.hooks.onEvent

I see the following comments:


```
import {  onEvent } from '@builder.io/mitosis';

Adds an event listener to a given element ref.
  packages/ core/ src/ index. ts

```

It requires   elementRef: T which refers to HTML Element. 

So this is just for HTML refs to add event listeners, like touch, keydown... this is not the same as output emitters of the component. The output can be called from JS and must not be part of an HTML ref. So I think onEvents is not the way to go.

nmerget:
> @nmerget but the onEvent is part of
> 
> When I read the comment of component.hooks.onEvent
> 
> I see the following comments:
> 
> ```
> import {  onEvent } from '@builder.io/mitosis';
> 
> Adds an event listener to a given element ref.
>   packages/ core/ src/ index. ts
> ```
> 
> It requires elementRef: T which refers to HTML Element.
> 
> So this is just for HTML refs to add event listeners, like touch, keydown... this is not the same as output emitters of the component. The output can be called from JS and must not be part of an HTML ref. So I think onEvents is not the way to go.

Sorry I think it was a little bit confusing. I tought about using something **like** `processOnEventHooksPlugin()`. So you should write your own plugin but you could do similar to `processOnEventHooksPlugin()` by traversing trough the nodes.

PR #1566, opened 2024-09-26
feat: Serialize Personalized Container in a JSX friendly way

manucorporat:
## Description
Needed for the LLM to serialize personalized content


samijaber (commented):
Can you add some tests to https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/__tests__/builder.test.ts#L9 so we see what the generation and parsing of this component looks like?

PR #1595, opened 2024-10-16
add slots support for builder and mitosis jsx

steve8708:
that way the AI will work correctly with child nodes that are not direct children

needed for components registered with builder that are like

```<Layout top={...} left={...} />```

currently those don't go to mitosis JSX and see the AI cannot "see" these and things break

### CODE REVIEW

File: packages/core/src/generators/builder.ts [hunk 1]
```diff
@@ -260,7 +266,9 @@ export const blockToBuilder = (
 
   if (thisIsComponent) {
     for (const key in json.bindings) {
-      builderBindings[`component.options.${key}`] = json.bindings[key]!.code;
+      if (!json.slots?.[key]) {
```
samijaber, line 269: yup good catch. unfortunately need to keep a copy of slots in `json.bindings` as well until all generators have migrated to using `json.slots`

steve8708: ya, just no need in the builder json right

packages/core/src/__tests__/__snapshots__/mitosis.test.ts.snap
```diff
@@ -1214,7 +1214,10 @@ import ContentSlotCode from \\"./content-slot-jsx.raw\\";
 export default function SlotCode(props) {
   return (
     <div>
-      <ContentSlotCode slotTesting={<div>Hello</div>} />
+      <ContentSlotCode
+        slotTesting={<div>Hello</div>}
+        slotTesting={<div>Hello</div>}
+      />
```
samijaber, line 1220: i think you have a bug. you might have forgotten to exclude some of the slots in `json.bindings`

samijaber (approved):
LGTM.

you forgot the changeset, i added it

PR #1602, opened 2024-10-23
fix: support <Show else> in builder json

manucorporat:


packages/core/src/generators/builder/generator.ts
```diff
@@ -186,6 +186,57 @@ const componentMappers: {
...
+    const elseNode =
+      elseCase && filterEmptyTextNodes(elseCase)
+        ? el(
+            {
+              // TODO: the reverse mapping for this
+              component: {
+                name: 'Core:Fragment',
+              },
+              bindings: {
+                show: `!(${node.bindings.when?.code as string})`,
```
steve8708, line 217: this alternatively can just use `hide`
```suggestion
                hide: node.bindings.when?.code as string,
```

packages/core/src/generators/solid/blocks.ts
```diff
@@ -18,20 +19,26 @@ const transformAttributeName = (name: string) => {
...
-  if (json.properties._text) {
-    return json.properties._text;
-  }
-  if (json.bindings._text?.code) {
-    return `{${json.bindings._text.code}}`;
+export const blockToSolid = (
+  json: MitosisNode,
+  component: MitosisComponent,
+  options: ToSolidOptions,
+  insideJsx: boolean,
```
samijaber, line 26: can we keep the named arguments for this function? i find that much more readable than having to guess their order

samijaber (approved):
nice

packages/core/src/generators/solid/blocks.ts
manucorporat: it's a ton of new new objects to GC, and solid it's the only one doing it

PR #1605, opened 2024-10-24
fix multiline bindings in builder json

steve8708:


packages/core/src/generators/builder/generator.ts
```diff
@@ -481,6 +481,20 @@ export const componentToBuilder =
...
+          for (const [key, value] of Object.entries(el.bindings)) {
+            if (value.match(/\n|;/)) {
+              if (!el.code) {
+                el.code = {};
+              }
+              if (!el.code.bindings) {
+                el.code.bindings = {};
+              }
+              el.code.bindings[key] = value;
+              el.bindings[key] = `return ${value}`;
```
samijaber, line 494: heads up that this is only ok if its a one-line binding. multi line will break

samijaber (approved):

packages/core/src/generators/builder/generator.ts
steve8708: ah i see your example now, yeah all bindings from mitosis should be one line. you can't bind to a block. i.e. you can'ta do this in jsx

```jsx
<Component foo={
  const x = 123
  return x + 10
  } />
  ```
  
  all binding values must be expressions

samijaber: ok yes thats what i meant. it can't be a block of JS code, it has to be a single expression

PR #1610, opened 2024-10-26
fix: map() parser from jsx-lite

manucorporat:
- Handle a lot of new cases
  - `array.map(() => </br>);`
  - `array.map(function() { return <div/> })`
  - `Array.from({length: 10}, (_, index) => {});`

- Add indexName to builder generator
- Fix builder parser to get the indexName
- Fix vue/solid/angular/lit generators for indexName


steve8708:
if this requires the SDKs to have indexName then we need to avoid merging this - historic SDKs won't have that and we don't want things to be broken

is there any way to get these updates in without requiring an SDK update? that'll save us from a lot of problems, esp this last minute

samijaber (approved):

PR #1619, opened 2024-10-29
Fix: builder shadow var 2

samijaber:


packages/core/src/__tests__/data/blocks/index-in-for.raw.tsx
```diff
@@ -10,12 +10,14 @@ export default function MyComponent(props) {
     <div>
       {todos.map((todo) => (
         <div>
-          {state.todos.map((todo, index) => (
+          {state.todos.map((todo) => (
             <input
               type="checkbox"
               checked={todo.completed}
+              key={todo.id + '-' + index}
```
manucorporat, line 17: index is undefined, no?

samijaber: ah good point. i should have kept `index` in the map for this

packages/core/src/__tests__/data/blocks/index-in-for.raw.tsx
```diff
@@ -10,12 +10,14 @@ export default function MyComponent(props) {
     <div>
       {todos.map((todo) => (
         <div>
-          {state.todos.map((todo, index) => (
+          {state.todos.map((todo) => (
```
samijaber, line 13:
```suggestion
          {state.todos.map((todo, index) => (
```

packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
```diff
@@ -1363,8 +1363,10 @@ export default function MyComponent(props) {
                 <input
                   type=\\"checkbox\\"
                   checked={todo.completed}
+                  key={todo.id + \\"-\\" + index}
+                  value={index}
                   foo={{
-                    bar: 1,
+                    bar: 1 + state.$index,
```
samijaber, line 1369: hmm @manucorporat it's weird that the `key` and `value` here are not getting updated to `state.$index`

I think that's a bug no?

PR #1629, opened 2024-11-05
fix: do not crash with comment before method in store

liamdebeasi:
## Description

https://builder-io.atlassian.net/browse/ENG-7332

This PR fixes an issue where comments before a method in state would cause Mitosis to crash.

Given: 
```js
useStore({
  foo: true, // comment
  bar() {}
})
```

The comment would get pushed to its own line and become part of the `bar` node:

```js
// comment
bar() {}
```

The expression handling fallback would then try to do something like this:

```js
let _ = // comment
bar() {}
```

which is not a valid expression and would crash.

My PR fixes the issue by ensuring that the only the method code is used and any auxiliary data such as comments are removed from parsing.


samijaber (approved):
oh nice, this has been driving many people nuts!

PR #1630, opened 2024-11-06
fix: identifiers are not incorrectly scoped to state

liamdebeasi:
## Description

https://builder-io.atlassian.net/browse/ENG-7338

Mitosis transforms identifiers that have the same name as state variables to have `this.` or `state.` prefixed where necessary. This causes issues with shadowed variables that are unrelated to state.

In the following example, the `errors` in `foo` have nothing to do with the `errors` defined in state.

```js
const state = useStore({
  errors: {},
  foo(errors) {
    return errors;
  }
});
```

However, this is incorrectly prefixed with `state.` which causes runtime issues:

```js
const state = useStore({
  errors: {},
  foo(state.errors) {
    return state.errors;
  }
});
```

This transformation is only needed when moving functions into state or when using `useState`. To fix this, I updated Mitosis to only transform state values that were set using `useState` or were function that got moved to state.


.changeset/late-papayas-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+identifiers are not incorrectly scoped to state
```
samijaber, line 5: can we follow the changeset format here? what exactly got fixed, for which generators/parsers, etc.

liamdebeasi: updated in https://github.com/BuilderIO/mitosis/pull/1630/commits/ed6b5f81b5879abb525f3700e377c09bc5503472

Let me know if you have other ideas for how to word this

.changeset/late-papayas-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+shadowed variables in state methods are not scoped to state in all generators
```
samijaber, line 5: [All] Fix: scope renaming of state methods to not include shadow variables
[Angular]: Update `state.*` -> `this.*` transform to new AST transform approach

samijaber, line 5:
```suggestion
[All] Fix: scope renaming of state methods to not include shadow variables
[Angular]: Update `state.*` -> `this.*` transform to new AST transform approach
```

samijaber (approved):

PR #1634, opened 2024-11-08
Solid: fix Fragment w/ `key` prop

samijaber:

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
@@ -3743,25 +3743,55 @@ export default MyComponent;
 `;
 
 exports[`Solid > jsx > Javascript Test > basicForFragment 1`] = `
-"Property or signature expected. (26:17)
-  24 | {(option, _index) => {
-  25 |   const index = _index();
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/__snapshots__/solid.test.ts.snap
```diff
@@ -11957,25 +11987,55 @@ export default MyComponent;
...
+
+function BasicForFragment(props: any) {
+  return (
+    <>
+      <div>
+        <For each={[\\"a\\", \\"b\\", \\"c\\"]}>
+          {(option, _index) => {
+            const index = _index();
+            return (
+              <Fragment key={\`key-\${option}\`}>
```
sidmohanty11, line 12026: @samijaber is Fragment a global component in solid? Found out here that solid fragments are already keyed: https://github.com/solidjs/solid/discussions/366 do we need to handle it? or just `<>...</>` should be sufficient?

samijaber: Ah, good catch!

- `Fragment` is not a global component in solid, or actually a component at all! I don't know why I thought it was
- good find on the issue, can keep `<>`

fixed in https://github.com/BuilderIO/mitosis/pull/1639 to ignore all props in Fragments

PR #1642, opened 2024-11-15
fix: do not set width binding on Builder Column if value is undefined

liamdebeasi:
## Description

We always create a `width` binding even if the width is undefined. This causes issues when converting a Mitosis node to JSX as it will yield `<Column width={} />`. Empty expressions (`width={}`) are not valid in JSX, so this will crash and JSX parser.

I fixed the issue to only add a width binding when the value is defined.

### DIFF
--- a/.changeset/soft-moons-change.md
+++ b/.changeset/soft-moons-change.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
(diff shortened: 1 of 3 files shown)


smzelek (approved):

liamdebeasi:
Sami and I discussed offline, and we need both my change and the one Sami suggested.

PR #1644, opened 2024-11-18
fix: preserve bound media query styles when converting builder to mitosis

liamdebeasi:
## Description
https://builder-io.atlassian.net/browse/ENG-7491

Bound media query styles are not preserved when converting Builder to Mitosis. This PR addresses the problem by ensuring that bound media query styles are included in the Mitosis node when bound using `responsiveStyles.[SIZE].[CSS PROP]`.

### DIFF
--- a/packages/core/src/__tests__/builder.test.ts
+++ b/packages/core/src/__tests__/builder.test.ts
@@ -648,6 +648,7 @@ describe('Builder', () => {
               'responsiveStyles.small.left': 'state.left',
               'responsiveStyles.small.top': 'state.top',
               'responsiveStyles.large.color': 'state.color',
+              'style.fontSize': 'state.fontSize',
             },
           },
         ],
@@ -659,7 +660,7 @@ describe('Builder', () => {
       {
         "style": {
           "bindingType": "expression",
-          "code": "{ '@media (max-width: 640px)': {\\"left\\":\\"state.left\\",\\"top\\":\\"state.top\\"}, '@media (max-width: 1200px)': {\\"color\\":\\"state.color\\"}, }",
+          "code": "{ fontSize: state.fontSize, '@media (max-width: 640px)': {\\"left\\":\\"state.left\\",\\"top\\":\\"state.top\\"}, '@media (max-width: 1200px)': {\\"color\\":\\"state.color\\"}, }",
           "type": "single",
         },
       }
@@ -671,6 +672,7 @@ describe('Builder', () => {
         return (
           <div
             style={{
+              fontSize: state.fontSize,
               \\"@media (max-width: 640px)\\": {
                 left: \\"state.left\\",
                 top: \\"state.top\\",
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/builder.test.ts
```diff
@@ -637,6 +637,56 @@ describe('Builder', () => {
...
+    const content = {
+      data: {
+        blocks: [
+          {
+            '@type': '@builder.io/sdk:Element' as const,
+            bindings: {
+              'responsiveStyles.small.left': 'state.left',
+              'responsiveStyles.small.top': 'state.top',
+              'responsiveStyles.large.color': 'state.color',
+              'style.fontSize': 'state.fontSize',
```
liamdebeasi, line 651: I wanted to show that the presence of `responsiveStyles.` doesn't interfere with other styles bound using `style.`

packages/core/src/parsers/builder/builder.ts
```diff
@@ -108,17 +108,46 @@ const getActionBindingsFromBlock = (
...
+         * @media (max-width: 1200px): {
+         *   color: "state.color",
+         *   background: "state.background"
+         * }
+         */
+      } else if (key.includes('responsiveStyles')) {
+        const [_, size, prop] = key.split('.');
+        const mediaKey = `@media (max-width: ${sizes[size as Size].max}px)`;
+
+        const objKey = `'${mediaKey}'`;
```
smzelek, line 139: just out of curiosity more than anything, does wrapping this do anything as opposed to using `responsiveStyles[mediaKey]`?

liamdebeasi: The key needs to be surrounded in quotes because it has spaces/special characters. Here's what it looks like without the extra quotes:

```diff
- Expected
+ Received

  {
    "style": {
      "bindingType": "expression",
-     "code": "{ fontSize: state.fontSize, '@media (max-width: 640px)': {\"left\":\"state.left\",\"top\":\"state.top\"}, '@media (max-width: 1200px)': {\"color\":\"state.color\"}, }",
+     "code": "{ fontSize: state.fontSize, @media (max-width: 640px): {\"left\":\"state.left\",\"top\":\"state.top\"}, @media (max-width: 1200px): {\"color\":\"state.color\"}, }",
      "type": "single",
    },
  }
```

Though this is a good reminder that I should document why this is needed.

liamdebeasi: documented here: https://github.com/BuilderIO/mitosis/pull/1644/commits/8d7ae81bc548c6a84a9d1ac243096a443ec106ea

also switched to using double quotes for consistency with the rest of the object (shouldn't make any functional difference)

liamdebeasi:
Added the follow up ticket per our discussion earlier today: https://builder-io.atlassian.net/browse/ENG-7569

PR #1653, opened 2024-12-10
Generate from angular selectors. ENG-7482

STRd6:
## Description

Generate dom structure based on the actual angular selector for the component.

Previously we would only generate for very simple selectors in some circumstances ex: `some-component`.

Now we correctly parse the selector and add appropriate tag name, class names, id, and attributes.

Currently checking in selector-parser.js as a generated file which is not ideal but I don't know how to hook into the tsc build process to register a transpiler like esbuild can.

ENG-7482


packages/core/src/generators/angular/selector-parser.hera
```diff
@@ -0,0 +1,67 @@
+# Parser for angular selectors
+# https://angular.dev/guide/components/selectors#types-of-selectors
```
liamdebeasi, line 2: Out of curiosity, is Angular's selector parser publicly exported? (I.e. could we leverage that at all?)

packages/core/src/__tests__/angular.selector.test.ts
```diff
@@ -0,0 +1,16 @@
+import { parse } from '../../src/generators/angular/selector-parser.js';
+
+describe('Angular selectors', () => {
+  test('should parse gnarly selectors', () => {
+    expect(parse('ccc.c1#wat[co].c2[counter="cool"]#wat[x=\'y\'].c3')).toEqual({
```
liamdebeasi, line 5: Are support for the `:not` pseudo-class and multiple selectors (e.g. `foo, [bar]`) in scope? Both are supported by Angular: https://angular.dev/guide/components/selectors#types-of-selectors

packages/core/src/generators/angular/index.ts
```diff
@@ -427,10 +429,49 @@ export const blockToAngular = ({
...
+    if (id) {
+      str += `#${id} `;
+    }
+
+    // TODO: merge with existing classes/bindings
+    if (classes.length) {
+      str += `class="${classes.join(' ')}" `;
+    }
+
+    // TODO: Merge with existing properties
```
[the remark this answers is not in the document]
STRd6: I think it's mostly an FYI. Especially with classes and bindings it may be difficult to do in the general sense. Once we get a bug report about it it will help us to narrow the scope to practical examples.

packages/core/src/generators/angular/selector-parser.hera
STRd6: Good call, I think they do export the parser and it might not be too difficult to transform its output into a format we can use for code generation.

https://github.com/angular/angular/blob/main/packages/compiler/src/selector.ts

packages/core/src/__tests__/angular.selector.test.ts
STRd6: 👍 I should add parsing for those. They don't affect the code generation but will need to be accounted for in the parsing.

manucorporat (commented):
LGTM!

packages/core/src/__tests__/angular.selector.test.ts
liamdebeasi: Ok great! Only other thing I can think of here is to add a test for those cases for parsing.

liamdebeasi (approved):
One non-blocking suggestion, but otherwise looks good. Thanks for looking into the Angular parser!

packages/core/src/__tests__/angular.selector.test.ts
```diff
@@ -0,0 +1,16 @@
+import { parse } from '../generators/angular/parse-selector';
```
liamdebeasi, line 1: Oh one more thing I just noticed: Do you intend for this to be a version bump in Mitosis? If so, you'll need to generate a new changeset.

PR #1661, opened 2025-01-15
refactor: change defaultProps generation for react

### LINKED ISSUE #1471: React's defaultProps will be remove from function components in a future release

Yes

- [ ] All
- [ ] Angular
- [ ] HTML
- [ ] Preact
- [ ] Qwik
- [X] React
- [ ] React-Native
- [ ] Solid
- [ ] Stencil
- [ ] Svelte
- [ ] Vue
- [ ] Web components

### Reproduction case

https://mitosis.builder.io/playground/?code=JYWwDg9gTgLgBAbzgVwM4FMAi6BmBDZAGxgAUoIxU4BfOHckOAcgAEAjZYQgE3SgDpgEAPQhgMCKmComAbgBQ84ADsYffAGN0cMhSoJ5cOADc8hZOgD8ALjioYUFQHMF1RegAekWHF74i8DjIyhowQspwALIAngDCEOAQyuiqABRg5JS2upQAlIiGKBjY%2FsQ5qKkGRkam5ui2TAAW6ISEEEwANIXUuQqFUOgwyFARADzcwMYAfAgZevy1FtSjwhPTrkA

### Expected Behaviour

Default props are defined with ES6 default parameters.

```tsx

interface Props {
  value?: string;
}

function MyComponent({ value: "hello" }: Props) {
  return <div>{value}</div>;
}

export default MyComponent;
```

PS. The playground doesn't not support TypeScript at the monent.

### Actual Behaviour

Using `defaultProps` which will be removed in React 19.

```tsx

interface Props {
  value?: string;
}

function MyComponent(props: Props) {
  return <div>{props.value}</div>;
}

MyComponent.defaultProps = { value: "hello" };

export default MyComponent;
```

The error was also shown in the console.

image

### Additional Information

`defaultProps` will be removed in place of ES6 default parameters in React 19
https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops

nmerget:
## Description

closes #1471

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/react-state-variables.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/react-state-variables.test.ts.snap
@@ -1,63 +1,115 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`React - stateType: variables > jsx > Javascript Test > AdvancedRef 1`] = `
-"'(' expected. (18:34)
-  16 | const inputNoArgRef = useRef(null);
-  17 | const name= 'PatrickJS'
-> 18 | const onBlur = onBlur = function function onBlur() {
-     |                                  ^
-  19 | // Maintain focus
-  20 | inputRef.current.focus();
-  21 | }"
-`;
-
-exports[`React - stateType: variables > jsx > Javascript Test > Basic 1`] = `
 "\\"use client\\";
 import * as React from \\"react\\";
+import { useRef, useEffect } from \\"react\\";
 
-export const DEFAULT_VALUES = {
-  name: \\"Steve\\",
-};
-
-function MyBasicComponent(props) {
-  const name = \\"Steve\\";
-  const underscore_fn_name = function underscore_fn_name() {
-    return \\"bar\\";
+function MyBasicRefComponent(props) {
+  const inputRef = useRef(null);
+  const inputNoArgRef = useRef(null);
+  const name = \\"PatrickJS\\";
+  const onBlur = function onBlur() {
+    // Maintain focus
+    inputRef.current.focus();
   };
-  const age = 1;
-  const sports = [\\"\\"];
+  const lowerCaseName = function lowerCaseName() {
+    return name.toLowerCase();
+  };
+
+  useEffect(() => {
+    console.log(\\"Received an update\\");
+  }, [inputRef.current, inputNoArgRef.current]);
 
   return (
     <>
-      <div className=\\"test div\\">
-        <input
-          value={DEFAULT_VALUES.name || name}
-          onChange={(myEvent) => (name = myEvent.target.value)}
-        />
-        Hello! I can run in React, Vue, Solid, or Liquid! &gt;
+      <div>
+        {props.showInput ? (
+          <>
+            <input
+              className=\\"input\\"
+              ref={inputRef}
+              value={name}
+              onBlur={(event) => onBlur()}
+              onChange={(event) => (name = event.target.value)}
+            />
+            <label htmlFor=\\"cars\\" ref={inputNoArgRef}>
+              Choose a car:
+            </label>
+            <select name=\\"cars\\" id=\\"cars\\">
+              <option value=\\"supra\\">GR Supra</option>
+              <option value=\\"86\\">GR 86</option>
+            </select>
+          </>
+        ) : null}
+        Hello
+        {lowerCaseName()}! I can run in React, Qwik, Vue, Solid, or Web
+        Component!
       </div>
       <style jsx>{\`
-        .div {
-          padding: 10px;
+        .input {
+          color: red;
         }
       \`}</style>
     </>
   );
 }
 
-export default MyBasicComponent;
+export default MyBasicRefComponent;
 "
 `;
 
+exports[`React - stateType: variables > jsx > Javascript Test > Basic 1`] = `
+"'function' is not allowed as a variable declaration name. (18:7)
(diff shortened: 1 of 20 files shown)


samijaber (changes requested):
@nmerget Thanks a lot for tackling this long overdue refactor!

There is however one big drawback to your approach: by writing `props = { /* DEFAULT_PROPS */}`, you are overwriting any props that do not have defaults!

See this example:

CleanShot 2025-01-15 at 08 03 11@2x


```tsx
type Props = {
  foo?: string;
  bar?: string;
};

// BEFORE
const A = (props: Props) => <div>B: {JSON.stringify(props)}</div>;
A.defaultProps = { foo: 'bar' };

// AFTER
const B = (props: Props = { foo: 'bar' }) => (
  <div>A: {JSON.stringify(props)}</div>
);

// FIX
const C = (props: Props) => {
  props = { ...props, foo: 'bar' };
  return <div>C: {JSON.stringify(props)}</div>;
};

export default function App() {
  return (
    <>
      <A bar="another-prop" />
      <B bar="another-prop" />
      <C bar="another-prop" />
    </>
  );
}
```

I think we need to instead use the approach i provide in `C`

nmerget:
> ```tsx
> const C = (props: Props) => {
>   props = { ...props, foo: 'bar' };
>   return <div>C: {JSON.stringify(props)}</div>;
> };
> ```

This would always  overwrite the given props with the default value, wouldn't it? :P

nmerget:
Yes, I changed it to follow your suggestions. I also had to change other targets because the e2e tests were failing. ^.^

packages/core/src/__tests__/__snapshots__/rsc.test.ts.snap
```diff
@@ -482,68 +482,14 @@ export default Button;
...
-    </>
-  );
-}
-
-export default Column;
-"
+"'function' is not allowed as a variable declaration name. (21:9)
+  19 |   function Column(props) {
+  20 |
+> 21 |   const function getColumns() {
```
samijaber, line 488: This is a big regression. Seems to only happen for RSCs, and not other React generators.

samijaber: Looks like:

```ts
  const getColumns = function getColumns() {
    return props.columns || [];
  };
```

became

```ts
  const function getColumns() {
    return props.columns || [];
  };
```

which is invalid JS

packages/core/src/generators/react/helpers/state.ts
```diff
@@ -98,3 +100,74 @@ export const updateStateSettersInCode = (value: string, options: ToReactOptions)
...
+      ? `const state = useBuilderState(${getStateObjectStringFromComponent(json)});`
+      : options.stateType === 'variables'
+      ? getStateObjectStringFromComponent(json, {
+          format: 'variables',
+          keyPrefix: 'const',
+          valueMapper: (code, type, _, key) => {
+            if (key) {
+              if (type === 'getter') return getFunctionString(code.replace('get ', ''));
+              if (type === 'function')
+                return code.startsWith('async') ? code : getFunctionString(code);
```
samijaber, line 144: this refactor seems to have introduced the regression: `getFunctionString(code)` doesn't do the same thing as `${key} = function ${code}` in the original code

samijaber (changes requested):
Requesting changes because some of your changes caused a regression in the react generator

e2e/e2e-react/src/App.tsx
```diff
@@ -1,4 +1,5 @@
 // Mitosis output for not yet include .d.ts, so ignore the types when importing.
+// Trigger nx remote cache: 1
```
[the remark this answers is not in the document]
nmerget: Yes, somehow the cache for e2e-react was always used...

All the other e2e targets didn't use the cache, I don't know why react was the exception

packages/core/src/generators/react/helpers/state.ts
nmerget: I'll look at it

nmerget: Quick fix was to have an exception for `rsc` ^.^

packages/core/src/__tests__/__snapshots__/react-state-variables.test.ts.snap
```diff
@@ -1,63 +1,115 @@
...
-export default MyBasicComponent;
+export default MyBasicRefComponent;
 "
 `;
 
+exports[`React - stateType: variables > jsx > Javascript Test > Basic 1`] = `
+"'function' is not allowed as a variable declaration name. (19:7)
+  17 |
+  18 |   const name= 'Steve'
+> 19 | const function underscore_fn_name() {
```
samijaber, line 66: interesting...looks like your function refactor impacts `stateType: 'variables'` config on react.

It fixes a handful of cases where we used to have 

```ts
// before - invalid 
const onChange = onChange = function function onChange() {

// after - valid
const onChange = function onChange() {
```


but it also breaks the following:

```ts
//before - valid
const underscore_fn_name = function underscore_fn_name() {

//after - invalid
const function underscore_fn_name() {
```

Would this be fixed if your `rscPrefix` workaround applied all the time? If you don't mind testing this out, could be a huge benefit to the generator.

If it ends up causing more headaches, we can ignore and merge without the fix.

nmerget: I found the issue (I guess). 

The original issue had happend only for functions outside of the state/store

PR #1669, opened 2025-01-27
Null check

mrkoreye:
## Description

Adds null check for property values

https://www.loom.com/share/be7a4cd800614433aaa82b7675b87bd8

https://builder-io.atlassian.net/browse/ENG-8150

### DIFF
--- a/.changeset/itchy-chairs-collect.md
+++ b/.changeset/itchy-chairs-collect.md
@@ -2,4 +2,4 @@
 '@builder.io/mitosis': patch
 ---
 
-fix null check issue with localized values
+[Builder] Fix null check issue with localized values
--- a/.changeset/itchy-chairs-collect.md
+++ b/.changeset/itchy-chairs-collect.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+fix null check issue with localized values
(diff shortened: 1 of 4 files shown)


packages/core/src/parsers/builder/builder.ts
```diff
@@ -776,7 +776,11 @@ export const builderElementToMitosisNode = (
 
   const linkUrl = (block as any).linkUrl;
   if (linkUrl) {
-    if (typeof linkUrl === 'object' && linkUrl['@type'] === '@builder.io/core:LocalizedValue') {
+    if (
+      typeof linkUrl === 'object' &&
+      linkUrl !== null &&
```
samijaber, line 781: Weird that we need this since there's a `if (linkUrl)` right above?

.changeset/itchy-chairs-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+fix null check issue with localized values
```
samijaber, line 5: I try to have changesets specify which parser/generator they're impacting:

https://github.com/BuilderIO/mitosis/blob/main/CONTRIBUTING.md#changeset-format

```suggestion
[Builder] Fix null check issue with localized values
```

samijaber (approved):

packages/core/src/parsers/builder/builder.ts
mrkoreye: This particular null check might not be 100% necessary, but I wanted to keep it consistent with the other spots where we check for localized values

PR #1672, opened 2025-01-30
Correctly handle arrow function expression arguments when used with non-'on' properties

STRd6:
## Description

Handle arrow function expressions for non-on properties. See https://mitosis.builder.io/playground/?code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAwhALaR7B4wAUAlHAN4BQvcOFGAx0UcmwGDpAHmJ4w6GFOlwYwAM4wAvNzaoANHABGRgl20A%2BOKjgBqE%2FbgEAviukB6S1I4BuXi5AA%3D%3D%3D for an example of how it is currently broken.


.changeset/fresh-schools-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Fix arrow function expressions in non-on properties
```
samijaber, line 5: on how to format the changesets https://github.com/BuilderIO/mitosis/blob/main/CONTRIBUTING.md#changeset-format

samijaber (commented):
nice! just need to tweak the changeset format

packages/core/src/parsers/jsx/element-parser.ts
```diff
@@ -282,11 +282,9 @@ export const jsxElementToJson = (
         if (types.isStringLiteral(expression)) {
           // <Foo myProp={"hello"} />
           memo.properties[key] = expression.value;
-        } else if (types.isArrowFunctionExpression(expression)) {
+        } else if (key.startsWith('on') && types.isArrowFunctionExpression(expression)) {
```
kylefowler, line 285: so the fix for `non-on` properties is to not have them run this code for `isArrowFunctionExpression`?

STRd6: Non-on properties will now fall to the next `else` branch and treat them as `function() {...}` expressions which ends up working correctly. See the snapshot tests for more details.

kylefowler (approved):

PR #1676, opened 2025-02-04
[ENG-8183] Checking for a null value

seunss:
## Description

- What changes you made:
Checking for a null string
- Why you made them, and:
 Null values were being passed in for width
- Any other useful context:

### DIFF
--- a/.changeset/old-dots-design.md
+++ b/.changeset/old-dots-design.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+check for null string
--- a/packages/core/src/parsers/builder/builder.ts
+++ b/packages/core/src/parsers/builder/builder.ts
@@ -391,7 +391,7 @@ const componentMappers: {
            * If width if undefined, do not create a binding otherwise its JSX will
            * be <Column width={} /> which is not valid due to the empty expression.
            */
-          ...(col.width !== undefined && {
+          ...(col.width != null && {
             bindings: {
               width: { code: col.width.toString() },
             },


packages/core/src/parsers/builder/builder.ts
```diff
@@ -391,11 +391,12 @@ const componentMappers: {
            * If width if undefined, do not create a binding otherwise its JSX will
            * be <Column width={} /> which is not valid due to the empty expression.
            */
-          ...(col.width !== undefined && {
-            bindings: {
-              width: { code: col.width.toString() },
-            },
-          }),
+          ...(col.width !== undefined &&
+            col.width !== null && {
```
kylefowler, line 395: could this just be `col.width != null`?

seunss: No, as I believe Liam previously changed this to check for` undefined`, which led me to believe that values could be either `undefined` or `null`.

samijaber (commented):
You need to add changesets to your PR, see https://github.com/BuilderIO/mitosis/blob/main/CONTRIBUTING.md#preparing-your-pr (last bullet point) and the following section for the conventional format we use

packages/core/src/parsers/builder/builder.ts
kylefowler: `!= null` appears to evaluate the same way in these cases, though my javascript is rusty.

```
const t = undefined
const u = null
const v = "hello"
t != null
false
u != null
false
v != null
true
```

kylefowler (approved):

.changeset/old-dots-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+check for null string
```
samijaber, line 5: I push for more elaborate release notes that follow the format in https://github.com/BuilderIO/mitosis/blob/main/CONTRIBUTING.md#changeset-format

so that it's clear what part of Mitosis each change impacts.

In your case, it would be:

```suggestion
[Builder]: fix null column width crash.
```

PR #1689, opened 2025-02-17
[ENG-8368] 

armgjoka:
## Description

CodeGen Issue: invalid input jsx caused by invalid class names. 


manucorporat (approved):
This is an amazing PR! i know @samijaber will ask for a changeset, so we can release this though :)

packages/core/src/helpers/styles/collect-css.ts
```diff
@@ -43,8 +43,11 @@ const collectStyles = (
       if (nodeHasCss(item)) {
         const value = parseCssObject(item.bindings.css?.code as string);
         delete item.bindings.css;
-        const componentName = item.properties.$name
-          ? dashCase(item.properties.$name)
+
+        const cleanedName = item.properties.$name?.replace(/[^a-zA-Z0-9_-]/g, '');
```
samijaber, line 47: IIUC, this is making sure that we trim the start if it's not alphanumeric, i.e. an emoji? Can we comment that this is what it's doing?

armgjoka: @samijaber updated

samijaber (approved):

PR #1693, opened 2025-02-20
Remove broken emoji from text

armgjoka:
## Description

Noticed that when a text element has broken emoji it will cause issues when generating code for example: 
 "text": "<span style=\"font-weight: 600; font-size: 18px;\">Would you like to take your daily</span><br/><span style=\"font-weight: 600; font-size: 18px;\">recap quiz?</span> <span style=\"font-weight: 400; line-height: 29px; color: rgba(0,0,0,1);\">You have a 13 day steak**\ud83d**</span>**\udd25**"


.changeset/short-llamas-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+Remove broken emoji from text
```
samijaber, line 5: we like to specify the impacted generator/parser in the changelog so its clear what the change is for:


```suggestion
[Builder]: Remove broken emoji from text
```

samijaber (commented):
can we add a test to show the impact of this change?

.changeset/short-llamas-divide.md
armgjoka: Hey @samijaber, updated

armgjoka: @samijaber there are 2 changesets, should I delete the first one? Also added the test

samijaber (approved):

PR #1696, opened 2025-02-21
fix: issue with eslint plugin not allowing useDefaultProps

nmerget:
## Description

There was an issue with eslint plugin showing an error when using `useDefaultProps` outside of the component


packages/eslint-plugin/src/rules/only-default-function-and-imports.ts
```diff
@@ -56,7 +56,8 @@ const rule: Rule.RuleModule = {
             (!types.isExpressionStatement(child) ||
               !types.isCallExpression(child.expression) ||
               !types.isIdentifier(child.expression.callee) ||
-              child.expression.callee.name !== HOOKS.META_DATA)
+              (child.expression.callee.name !== HOOKS.META_DATA &&
```
samijaber, line 59: - can we add a test with `useDefaultProps`
- can we update the error message of this plugin

> disallow anything other than import declarations, the component itself (in a default export), and type declarations inside the component file.

to

> disallow anything other than import declarations, the component itself (in a default export), module-scope hooks, and type declarations inside the component file.

nmerget: done

packages/eslint-plugin/src/rules/only-default-function-and-imports.ts
```diff
@@ -7,12 +7,15 @@ import isMitosisPath from '../helpers/isMitosisPath';
 // Rule Definition
 // ------------------------------------------------------------------------------
 
+export const onlyDefaultFunctionAndImportsMessage =
+  'Mitosis component files should only contain import declarations, the component itself (in a default export), and type declarations';
```
samijaber, line 11:
```suggestion
  'Mitosis component files should only contain import declarations, the component itself (in a default export), module-scope hooks, and type declarations';
```

samijaber (approved):

PR #1697, opened 2025-02-21
fix: Builder parser does not strip falsey style values

liamdebeasi:
## Description

The Builder JSON parser currently strips out falsey style values (I.e. `padding: 0`). The original motivation for this was to cut down on the amount of CSS generated with codegen in Builder. However, our codegen process has improved since then where this is not a big problem anymore.

More importantly, stripping falsey values is causing functional changes to the code because we make the assumption that a value of `0` is the same as not having the CSS declaration to begin with which is not always true. For example, we remove values such as `flex-shrink: 0`. However, the [default value for `flex-shrink` is 1](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink#formal_definition).

Another example can be seen here: https://codepen.io/liamdebeasi/pen/gbOrgJm

In the first block `#text` has `z-index: 0` which causes it to show on top of the square. In the second block ,`#text` does not have `z-index: 0` which causes it to show below the square.

This PR resolves the issue by removing logic that removes falsey CSS values.

For added context, this logic was originally added [via a PR to Mitosis](https://github.com/BuilderIO/mitosis/pull/1273), but this PR was a port [from another repo](https://github.com/BuilderIO/builder-internal/pull/5130).

----

### DIFF
--- a/.changeset/rich-turkeys-smoke.md
+++ b/.changeset/rich-turkeys-smoke.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': patch
+---
+
+[Builder]: parser does not strip falsey style values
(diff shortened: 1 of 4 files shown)


STRd6 (approved):

kylefowler (approved):
ship it

PR #1702, opened 2025-02-26
feat: add dependency handling for onUpdate hook

nmerget:
## Description

Dependencies from `onUpdate` hook weren't used inside `componentDidUpdate`. By using `@Watch` we can generate a similar solution for `onUpdate` hook with dependencies.

### DIFF
--- a/packages/core/src/generators/stencil/helpers/index.ts
+++ b/packages/core/src/generators/stencil/helpers/index.ts
@@ -89,18 +89,27 @@ export const needsWrap = (children: MitosisNode[]): boolean => {
  * @param events
  * @param props
  * @param dataString
+ * @param watch
  */
-export const getStencilCoreImportsAsString = (
-  wrap: boolean,
-  events: string[],
-  props: string[],
-  dataString: string,
-): string => {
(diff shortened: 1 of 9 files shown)


packages/core/src/generators/stencil/helpers/index.ts
```diff
@@ -130,3 +139,16 @@ export const getImports = (
     },
   });
 };
+
+/**
+ * Converts the deps string into an array of strings
+ * @param deps The hook dependencies as string e.g.: "[this.a,this.b]"
+ */
+export const getDepsAsArray = (deps: string): string[] => {
```
samijaber, line 147: we do things like this so often in each generator...might be time to update the `hook.deps` so that it's an array of strings which gets concatenated by the individual generators without the need to parse the string, split and join

samijaber (approved):

packages/core/src/generators/stencil/helpers/index.ts
nmerget: Yes I thought about it as well. 👍

PR #1710, opened 2025-03-04
fix: Builder generator does not create duplicate component option mappings

liamdebeasi:
## Problem

The Builder generator maps options for Builder components to both `component.options` and `bindings`. [Example](https://mitosis.builder.io/playground/?code=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnALICeAwhALaR7B4wAUYUEYAzgJRwA3gCg4cKMBjoo5NqLFwAPAEl6qAObA4qHmGBEASqhIQAvIICMAXzgB6AHzy%2BAbmFWgA%3D%3D%3D)

In the example above, notice how `aspectRatio` appears on both `component.options` and `bindings['component.options.aspectRatio']`.

## Solution

The Builder generator loops through any component props and [either puts them on `component.options` or `bindings`](https://github.com/BuilderIO/mitosis/blob/57f371d409d2c0167aebe0ea82431b554efc3948/packages/core/src/generators/builder/generator.ts#L606-L611).

However, we later loop through the component props again and then [put them on `bindings`](https://github.com/BuilderIO/mitosis/blob/57f371d409d2c0167aebe0ea82431b554efc3948/packages/core/src/generators/builder/generator.ts#L654-L660), thus leading to the duplicate mapping problem.

I fixed the issue by removing the second loop.

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
@@ -1525,7 +1525,6 @@ exports[`Builder > Regenerate custom Hero 2`] = `
         "@type": "@builder.io/sdk:Element",
         "actions": {},
         "bindings": {
-          "component.options.height": "400",
           "component.options.multiBinding": " return {
   hello: state.message
 }",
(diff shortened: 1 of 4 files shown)


packages/core/src/__tests__/__snapshots__/builder.test.ts.snap
liamdebeasi, line 1: The GitHub preview doesn't show the full story here, but the `component.options.height` value is still there while the `bindings.component.options.height` value was removed.

smzelek (approved):

PR #1731, opened 2025-03-20
Feat angular signals

nmerget:
## Description

Angular v18+ uses [signals](https://angular.dev/guide/signals) as a new feature.
This allows the generator to match better with other targets.

This PR will rewrite the complete Angular generator to match all new features for Angular


nmerget:
@samijaber I'm excited to hear what you think about this one :D

samijaber (commented):
@nmerget This is a MONSTER PR 🙏🏽 

We have this exact need internally for outputting Signals in Angular and were going to start tackling it in a few weeks. Such a happy coincidence you just put this PR up!

Our team will review this shortly. Very excited to have this feature merged in!

cc @sidmohanty11

samijaber (commented):
Can we create a new angular e2e test that uses the `signals` API? Just to make sure it works at a basic level and renders some components.

If that turns out to be too tricky, then at the very least update the existing e2e-angular test to use `api: 'signals'`

nmerget:
> Can we create a new angular e2e test that uses the `signals` API? Just to make sure it works at a basic level and renders some components.
> 
> If that turns out to be too tricky, then at the very least update the existing e2e-angular test to use `api: 'signals'`

Yes, great idea. I'll copy paste the current `e2e-angular` to test both

nmerget:
> > Can we create a new angular e2e test that uses the `signals` API? Just to make sure it works at a basic level and renders some components.
> > If that turns out to be too tricky, then at the very least update the existing e2e-angular test to use `api: 'signals'`
> 
> Yes, great idea. I'll copy paste the current `e2e-angular` to test both

Hmm the copy&paste wouldn't work so easily. I'll convert it to `api=signals` it should be the default in the near future anyways: https://angular.dev/reference/releases#actively-supported-versions

relliv:
@nmerget Awesome! How about the ngModel support?

nmerget:
> @nmerget Awesome! How about the ngModel support?

I didn't implement this. 
We include it in our repo via a post-build [script](https://github.com/db-ux-design-system/core-web/blob/dependabot-npm_and_yarn-builder-io-22e5c146ba/packages/components/scripts/post-build/angular.ts#L48-L56) in additon to `ControlValueAccessor`. I would like to transfer it to mitosis as well. But it might be better to do it in an extra branch after this one is merged? But it won't be a big deal to come up with an `useMetadata` api to use [`model()`](https://angular.dev/api/core/model#) instead of `input()`

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -0,0 +1,349 @@
...
+      (item) => item.length && isUpperCase(item[0]) && !BUILT_IN_COMPONENTS.has(item),
+    );
+    const componentSettings: Record<string, any> = {
+      selector: `'${kebabCase(json.name)}'`,
+      standalone: 'true',
+      imports: `[${['CommonModule', ...componentsUsed].join(', ')}]`,
+      template: `\`${getTemplateFormat(template)}\``,
+    };
+    if (onPush) {
+      componentSettings.changeDetection = `'ChangeDetectionStrategy.OnPush'`;
```
sidmohanty11, line 161: http://localhost:3000/?outputTab=IYOw5grgNsBOQ%3D%3D%3D&code=FAVwzgpgshAuCGATeCAUBvYACL8B2A5iADbwBOAXFpjjgMYAW%2BBEAInBHbAJYD2eVAOQB5PAAVwDQdiwBfYLICUwYBAAeAB15lYWRBABm8EroMg8XPnixQAngGFeAWy14IeWKg1leGsIuoZMjgQMmtUGRwAHkRuADcAPkicBghiYl5kqIB6WMSZRQBuBSA%3D%3D
Screenshot 2025-04-02 at 3 28 22 PM

```suggestion
      componentSettings.changeDetection = 'ChangeDetectionStrategy.OnPush';
```

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -0,0 +1,349 @@
...
+        _: 'data' | 'function' | 'getter',
+        typeParameter: string | undefined,
+      ) => {
+        if (typeParameter && !code.length) {
+          console.error(`
+Component ${json.name} has state property without an initial value'. 
+This will cause an error in Angular.
+Please add a initial value for every state property even if it's \`undefined\`.`);
+        }
+        return `signal${typeParameter ? `<${typeParameter}>` : ''}(${code})`;
```
sidmohanty11, line 185: @nmerget would it be possible to update getters to `computed` of angular? This will fix the getter being called and allow us to only update when a dependent signal updates (also help us to not trigger the computation on every change detection cycle)

Screenshot 2025-04-02 at 3 31 56 PM

packages/core/src/helpers/event-handlers.ts
```diff
@@ -1,6 +1,6 @@
 import { camelCase } from 'lodash';
 
-export const checkIsEvent = (code: string) => code.startsWith('on');
+export const checkIsEvent = (code: string): boolean => code.startsWith('on');
```
sidmohanty11, line 3: if a prop starts with 'on', it strips on and updates it to an event/output
Screenshot 2025-04-02 at 3 35 11 PM

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -0,0 +1,349 @@
...
+          }
+          
+          ${withAttributePassing ? getAttributePassingString(options.typescript) : ''}
+          
+          ${
+            !json.hooks.onMount.length && !json.hooks.onInit?.code
+              ? ''
+              : `ngOnInit() {
+                  ${!json.hooks?.onInit ? '' : json.hooks.onInit?.code}
+                  ${json.hooks.onMount.length > 0 ? stringifySingleScopeOnMount(json) : ''}
```
sidmohanty11, line 310: in some cases, we don't want to call the function and just want to pass the function

Screenshot 2025-04-02 at 3 39 03 PM

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -0,0 +1,349 @@
...
+            return undefined;
+          },
+        })}
+    
+        @Component({
+          ${Object.entries(componentSettings)
+            .map(([k, v]) => `${k}: ${v}`)
+            .join(',')}
+        })
+        export class ${json.name} implements AfterViewInit {
```
sidmohanty11, line 251: mitosis components are by default- default exported, can we add an angular option: `defaultExportComponent` or something similar and allow exporting classes as default export? and should we have the `implements AfterViewInit` part? as we will dynamically add code to ngOnInit or any other hooks?

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -0,0 +1,349 @@
...
+      userOptions,
+    });
+    options.typescript = true; // Angular uses ts all the time
+    options.api = 'signals';
+
+    if (options.plugins) {
+      json = runPreJsonPlugins({ json, plugins: options.plugins });
+    }
+
+    const withAttributePassing = true; // We always want to pass attributes
```
sidmohanty11, line 81: @nmerget i wanted to understand if we should keep this behind the flag, this adds extra code in all of the components without us needing this feature. In some cases, we want to pass the input and use that instead of attaching them as HTML attributes. your thoughts?

packages/core/src/generators/angular/classic/plugins/get-class-properties-plugin.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+      } else {
+        makeReactiveState(
+          json,
+          newBindingName,
+          `state.${newBindingName} = {...(${item.bindings[key]!.code})}`,
+        );
+        item.bindings[newBindingName] = item.bindings[key];
+        item.bindings[key]!.code = `state.${newBindingName}`;
+        delete item.bindings[key];
```
sidmohanty11, line 105: for spread support, i had an idea and wanted to know your thoughts, would it be possible to convert any spread property or properties using backticks to a new computed property inside the class and use them as their bindings? similar to how we do it for `state: 'class-properties'` approach here. we can take this up in a follow-up PR

sidmohanty11 (commented):
Thanks so much @nmerget for going all in and adding Angular `signals` support. This is a HUGE improvement!!! Added a few comments, please take a look when you get a chance

samijaber (approved):
@nmerget I spoke with @sidmohanty11 and we agree that the best way to approach this is to merge this PR as-is, and let you iterate on the feedback in subsequent PRs.

Since this PR doesn't impact any existing generators, and doesn't seem to break the old Angular approach, there shouldn't be any issues.

Also this will make it easier to review the small fixes for us, and avoid you having to fix merge conflicts.

Thanks a ton for this massive new feature 🙏🏽 🚀 !!

packages/core/src/generators/angular/signals/component.ts
nmerget: Sure, we can use the flag again 👍

nmerget:
> Thanks so much @nmerget for going all in and adding Angular `signals` support. This is a HUGE improvement!!! Added a few comments, please take a look when you get a chance

Great feedback thank you. I tested the implementation against our repo, we don't use `gettters` at the moment - so there were no errors testing this, but you are right the `getters` shouldn't behave like other functions 👍 

Anyways there are some tasks left to do, I'll try to do them in a future branch :)

PR #1734, opened 2025-03-27
Fix: Vue Builder $tagName

samijaber:

### DIFF
--- a/packages/core/src/__tests__/builder.test.ts
+++ b/packages/core/src/__tests__/builder.test.ts
@@ -28,6 +28,7 @@ import lazyLoadSection from './data/builder/lazy-load-section.json?raw';
 import localization from './data/builder/localization.json?raw';
 import slotsContent from './data/builder/slots.json?raw';
 import slots2Content from './data/builder/slots2.json?raw';
+import tagNameContent from './data/builder/tag-name.json?raw';
 import textBindings from './data/builder/text-bindings.json?raw';
 import advancedFor from './data/for/advanced-for.raw.tsx?raw';
 import asyncBindings from './data/ref/basic-ref-assignment.raw.tsx?raw';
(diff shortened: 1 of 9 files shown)


packages/core/src/__tests__/builder.test.ts
```diff
@@ -511,6 +512,133 @@ describe('Builder', () => {
...
+        );
+      }
+
+      export default MyComponent;
+      "
+    `);
+
+    expect(vue).toMatchInlineSnapshot(`
+      "<template>
+        <ProgressBar :num=\\"10\\"><CounterComponent></CounterComponent></ProgressBar>
```
manucorporat, line 598: is it working?

packages/core/src/__tests__/builder.test.ts
```diff
@@ -511,6 +512,132 @@ describe('Builder', () => {
...
+        );
+      }
+
+      export default MyComponent;
+      "
+    `);
+
+    expect(vue).toMatchInlineSnapshot(`
+      "<template>
+        <progress-bar :num=\\"10\\"><counter></counter></progress-bar>
```
samijaber, line 598: @manucorporat works now ^

packages/core/src/plugins/compile-away-builder-components.ts
```diff
@@ -40,7 +41,7 @@ const wrapOutput = (
       ...omit(node.bindings, ...inputNames),
     },
     // TODO: forward tagName as a $tagName="..."
-    name: node.properties._tagName || node.properties.$tagName || 'div',
+    name: node.properties._tagName || getBuilderTagName(node) || 'div',
```
liamdebeasi, line 44: what's the difference between $tagName and _tagName? should both of those checks be in `getBuilderTagName` instead of only `$tagName`?

liamdebeasi (approved):
looks good! one non-blocking question

liamdebeasi:
okay technically I asked two non-blocking questions

packages/core/src/plugins/compile-away-builder-components.ts
samijaber: i actually have no clue. Not sure who's setting `_tagName` or where it's coming from 🤷🏽.

This is for Builder components that get compiled away when Mitosis runs, which is slightly different than the VCP flow _i think_.

samijaber: I can't find a reference to `_tagName` in any other codebase... 😅

PR #1736, opened 2025-03-28
feat: add escapeInvalidCode flag, drop invalid bindings

liamdebeasi:
## Description
The team would like to explore ways of fixing the input Builder JSON when there are things like syntax errors. To do this, we are adding an `escapeInvalidCode` to the Builder JSON parser. When enabled, invalid code bindings will be preserved in quotes so as to not cause Babel to crash. When disabled, invalid code bindings will be dropped with a console warning.

I also started a 2nd builder test file because the first one was getting quite large.


packages/core/src/parsers/builder/builder.ts
```diff
@@ -738,8 +751,17 @@ export const builderElementToMitosisNode = (
...
       const useKey = key.replace(/^(component\.)?options\./, '');
       if (!useKey.includes('.')) {
+        let code = (blockBindings[key] as any).code || blockBindings[key];
+
+        if (options.escapeInvalidCode) {
+          const verifyCode = verifyIsValid(code);
+          if (!verifyCode.valid) {
+            code = '`' + code + ' [INVALID CODE]`';
+          }
+        }
```
samijaber, line 761: one idea: if `escapInvalidCode` is false, would it be good to still verify if valid, and if it's not then maybe we drop the code altogether? In the spirit of failing gracefully?

not a blocker

packages/core/src/parsers/builder/builder.ts
```diff
@@ -116,11 +116,18 @@ const getStyleStringFromBlock = (block: BuilderElement, options: BuilderToMitosi
...
         continue;
       }
+
+      let code = block.code?.bindings?.[key] || block.bindings[key];
+      if (options.escapeInvalidCode) {
+        const verifyCode = verifyIsValid(code);
+        if (!verifyCode.valid) {
+          code = '`' + code + ' [INVALID CODE]`';
+        }
+      }
```
samijaber, line 126: same Q here

samijaber (approved):

packages/core/src/parsers/builder/builder.ts
liamdebeasi: added in https://github.com/BuilderIO/mitosis/pull/1736/commits/2303506d2708e2e1ed5e0279707832a5267c94a1

packages/core/src/parsers/builder/builder.ts
liamdebeasi: added in https://github.com/BuilderIO/mitosis/pull/1736/commits/2303506d2708e2e1ed5e0279707832a5267c94a1

PR #1739, opened 2025-04-01
fix: improve binding validation

liamdebeasi:
## Description

The invalid binding detection misses code that _is_ valid JS but is not a valid binding such as `export default foo`. This PR improves the detection by checking if the code can be used as a binding when passed into babel.

I'm also using the `parseCode` util now because it has presets that are needed to parse JSX/TSX.


packages/core/src/parsers/builder/builder.ts
```diff
@@ -779,12 +806,29 @@ export const builderElementToMitosisNode = (
...
         continue;
       }
+
       const useKey = key.replace(/^(component\.)?options\./, '');
       if (!useKey.includes('.')) {
         let code = (blockBindings[key] as any).code || blockBindings[key];
-
         const verifyCode = verifyIsValid(code);
         if (!verifyCode.valid) {
+          console.log(`Dropping binding "${key}" due to invalid code: ${code}`);
```
liamdebeasi, line 815: Is there a reason why we can't put this in the `else` block of the existing `verifyCode.valid` check?

liamdebeasi, line 815: I'm confused why this is needed as the `else` case of the `verifyCode.valid` check already has this log.

packages/core/src/parsers/builder/builder.ts
```diff
@@ -779,12 +806,29 @@ export const builderElementToMitosisNode = (
...
+        if (verifyCode.valid) {
+          if (options.escapeInvalidCode) {
+            // When escapeInvalidCode is true, process the code first, then wrap in new Function()
+            let processedCode = processBoundLogic(code);
+            processedCode = processedCode.trim();
+            if (processedCode.startsWith('return ')) {
+              processedCode = processedCode.slice(7).trim();
+            }
+            processedCode = processedCode.replace(/^;+|;+$/g, '');
+            code = `new Function(\`return ${processedCode};\`)()`;
```
liamdebeasi, line 827: can you add a comment explaining why this is needed?

packages/core/src/parsers/builder/builder.ts
```diff
@@ -1324,18 +1368,45 @@ export const builderContentToMitosisComponent = (
...
+      if (verifyCode.valid) {
+        if (options?.escapeInvalidCode) {
+          // When escapeInvalidCode is true, process the code first, then wrap in new Function()
+          let processedCode = processBoundLogic(code);
+          processedCode = processedCode.trim();
+          if (processedCode.startsWith('return ')) {
+            processedCode = processedCode.slice(7).trim();
+          }
+          processedCode = processedCode.replace(/^;+|;+$/g, '');
+          code = `new Function(\`return ${processedCode};\`)()`;
```
liamdebeasi, line 1396: same here, can you add a comment explaining why this is needed?

packages/core/src/parsers/builder/builder.ts
```diff
@@ -779,12 +806,29 @@ export const builderElementToMitosisNode = (
...
         if (!verifyCode.valid) {
+          console.log(`Dropping binding "${key}" due to invalid code: ${code}`);
+        }
+
+        if (verifyCode.valid) {
+          if (options.escapeInvalidCode) {
+            // When escapeInvalidCode is true, process the code first, then wrap in new Function()
+            let processedCode = processBoundLogic(code);
+            processedCode = processedCode.trim();
+            if (processedCode.startsWith('return ')) {
```
liamdebeasi, line 823: why do we remove `return` just to create a string that has it in the `new Function`?

liamdebeasi (commented):
overall looks good! a few small comments. you'll need to get an approval from someone other than me since I'm the PR author

packages/core/src/parsers/builder/builder.ts
midhunadarvin: I think it was added by mistake will remove it.

samijaber (approved):

PR #1743, opened 2025-04-03
Correctly handle Fragments in Builder ("Core:Fragment" component name not "Fragment")

steve8708:


manucorporat (approved):
makes sense to me, but i am not sure

PR #1752, opened 2025-04-14
fix: expressions in action bindings are converted to call expressions

liamdebeasi:
## Description
A [known Mitosis gotcha](https://mitosis.builder.io/docs/gotchas/#functions-cant-be-passed-by-reference-to-jsx-callbacks) is that you cannot pass functions by reference to JSX callbacks (I.e. `onClick={callback}`.

This PR fixes the issue so that these expressions are converted to call expressions.

Note that the snapshots are changing because of the fixed issue. I also removed the gotcha from the documentation.


packages/core/src/parsers/jsx/element-parser.ts
```diff
@@ -294,6 +294,14 @@ export const jsxElementToJson = (
             arguments: args.length ? args : undefined,
             bindingType: 'function',
           });
+        } else if (key.startsWith('on') && types.isExpression(expression)) {
```
kylefowler, line 297: just to make sure im understanding, this change is intended to turn expressions into function executions?

liamdebeasi: That's correct

kylefowler: is there any risk that someone else has a prop that starts with `on` which is in conflict here or are these all mitosis specified events?

kylefowler (approved):

packages/core/src/parsers/jsx/element-parser.ts
liamdebeasi: good thought. added a fix+test for that here: https://github.com/BuilderIO/mitosis/pull/1752/commits/95b260e12f7ff03e211b43449e608616d72f57af

PR #1753, opened 2025-04-16
feat: support serializing data with deeply nested nodes

liamdebeasi:
## Problem

The current way we serialize deeply nested Builder elements can cause invalid JSX to be generated. Consider the following Builder JSON:

```js
{
  data: {
    blocks: [
      {
        '@type': '@builder.io/sdk:Element',
        component: {
          name: 'TestComponent',
          options: {
            items: {
              foo: {
                '@type': '@builder.io/sdk:Element',
                tagName: 'OtherCmp',
                bindings: {
                  prop: 'state.baz',
                },
              },
            },
          },
        },
      },
    ],
  },
}
```

Previously, all options (including Builder element in said options) were stringified as-is. (See **Mitosis JSON Before**). This results in incorrect code being generated because the Builder elements are preserved when they should be JSX elements (See **JSX Results Before**).

## Solution

This PR adds a `blocksSlots` key on MitosisNode which can be used to transform deeply nested objects while preserving the structure of the data it is contained in (See **Mitosis JSON After**).

The JSX generator now writes out `blocksSlots` data as JSX elements while preserving the structure of the data that the elements are in (See **JSX Results After**).

This feature is behind a `enableBlocksSlots` flag as a) this change is destructive for generators that do not support reading `blocksSlots` and b) not every generator will be able to support this kind of serialization. For example, Vue currently cannot do something like this: `<MyCmp :items=[<div />, <br />] />`

### Mitosis JSON
(empty key/value pairs omitted for brevity)
**Before** 

```js
{
  "@type": "@builder.io/mitosis/node",
  "bindings": {
    "items": {
      "bindingType": "expression",
      "code": "{foo:{'@type':'@builder.io/sdk:Element',tagName:'OtherCmp',bindings:{prop:'state.baz'}}}",
      "type": "single",
    },
  },
  "name": "TestComponent"
}
```

**After**

```js
{
  "@type": "@builder.io/mitosis/node",
  "blocksSlots": {
    "items": {
      "foo": {
        "@type": "@builder.io/mitosis/node",
        "bindings": {
          "prop": {
            "bindingType": "expression",
            "code": "state.baz",
            "type": "single",
          },
        },
        "name": "OtherCmp",
      },
    },
  },
  "name": "TestComponent",
}
```

### JSX Results

**Before**
```jsx
import { TestComponent } from \\"@components\\";

  return (
    <TestComponent
      items={{
        foo: {
          \\"@type\\": \\"@builder.io/sdk:Element\\",
          tagName: \\"OtherCmp\\",
          bindings: {
            prop: \\"state.baz\\",
          },
        },
      }}
    />
  );
}
```

**After**

```jsx
import { TestComponent, OtherCmp } from "@components";

  return <TestComponent items={{ foo: <OtherCmp prop={state.baz} /> }} />;
}
```


packages/core/src/types/mitosis-node.ts
```diff
@@ -82,6 +82,8 @@ export type BaseNode = {
    * It is used when a Builder content block has localized values.
    */
   localizedValues?: { [index: string]: BuilderLocalizedValue };
+
+  blocksSlots?: { [key: string]: any }; // TODO fix this
```
liamdebeasi, line 86: Not really sure how best to type this. open to ideas!

liamdebeasi: just tried this and ran into errors. you could have this be multiple levels deep, so the following wouldn't be considered value:

```js
blocksSlots: {
  foo: {
    bar: [...MitosisNodes]
  }
}
```

liamdebeasi: ok I think I fixed the type issue: https://github.com/BuilderIO/mitosis/pull/1753/files#diff-1f814d37f53a063bfb900a99e127835b30dcc8d057c7b010ee6960ef9f21ee89

liamdebeasi:
@samijaber I added the tests for the code we talked about last week in https://github.com/BuilderIO/mitosis/pull/1753/commits/ac3efd47f4d3c3e01069ef19326d33603ab016d3

samijaber (approved):

PR #1757, opened 2025-04-30
Fix single quote issue in swift generation

kylefowler:
## Description
Fixes the single quote -> double quote conversion for swift code generation.


packages/core/src/generators/swift/helpers.ts
```diff
@@ -29,9 +25,51 @@ export const convertConsoleLogToPrint = (code: string): string => {
...
+        if (code[j] === "'" && code[j - 1] !== '\\') {
+          break;
+        }
+        j++;
+      }
+
+      // Add the string content
+      result += code.substring(i + 1, j);
+
+      // Add closing double quote if we found the end
```
manucorporat, line 59: great comments!

manucorporat (approved):

PR #1769, opened 2025-05-29
fix: Builder Text node is preserved when converting to Mitosis JSON

liamdebeasi:
## Problem

Builder Text nodes are compiled away by default when converting to Mitosis JSON (and by extension Mitosis JSX). This means that `<Text text="foo" />` becomes `<div>foo</div>`. When converting Mitosis JSX back to Builder JSON we intentionally do not convert the div back to a Builder Text node because we have no idea if it's supposed to remain as a regular div.

This causes issues because the Builder Text primitive has additional styling built-in. When we convert JSX back to Builder JSON that styling is effectively lost since the Text node is not re-created. This can result in rendering differences due to the loss of the Text node.

## Solution

Converting from Builder JSON to Mitosis JSON no longer compiles away Builder Text nodes by default. As a result, `<Text text="foo" />` remains as-is when converting to Mitosis JSON/JSX. Similarly, the Text primitive is preserved when converting back to Builder JSON.

The `compileAwayBuilderComponents` function has been updated to account for `Text`, so developers who still wish to have this primitive compiled away can use that function to do so.

## Other Information

- The compile away logic for Text that I removed had a lot of Text-specific logic that I don't think we need anymore now that we are preserving Text nodes. I intentionally made the logic in `compileAwayBuilderComponents` simple, but we can add logic there as needed.
- This change revealed some existing bugs in certain generators where the `Text` primitive isn't handled properly. To avoid playing "whack a mole" and making this PR larger than it already is, I left TODOs where necessary.

### DIFF
--- a/packages/core/src/plugins/compile-away-builder-components.ts
+++ b/packages/core/src/plugins/compile-away-builder-components.ts
@@ -571,8 +571,11 @@ export const components: CompileAwayComponentsMap = {
       delete node.bindings.text;
     }
 
-    const hasBindings = Object.keys(node.bindings).length > 0;
-    const hasProperties = Object.keys(node.properties).length > 0;
+    const { _text: propText, $tagName: nodeTagName, ...outerProps } = node.properties;
+    const { _text: bindingText, ...outerBindings } = node.bindings;
+
+    const hasBindings = Object.keys(outerProps).length > 0;
+    const hasProperties = Object.keys(outerBindings).length > 0;
     /**
      * If there are things we need to reflect on the text then we must
      * render a wrapper div so we can put it on that element.
@@ -585,26 +588,18 @@ export const components: CompileAwayComponentsMap = {
        * other bindings need to go on the wrapper div so they get generated.
        */
 
-      const { _text: propText, $tagName: nodeTagName, ...restOfProperties } = node.properties;
-      const { _text: bindingText, ...restOfBindings } = node.bindings;
-
       return createMitosisNode({
         name: 'div',
-        bindings: restOfBindings,
-        properties: restOfProperties,
+        bindings: outerBindings,
+        properties: outerProps,
         children: [
           createMitosisNode({
             ...node,
-            ...(propText && {
-              properties: {
-                _text: propText,
-              },
-            }),
-            ...(bindingText && {
-              bindings: {
-                _text: bindingText,
-              },
-            }),
+            properties: {
+              $tagName: nodeTagName,
+              ...(propText ? { _text: propText } : {}),
+            },
+            bindings: bindingText ? { _text: bindingText } : {},
             name: nodeTagName ?? 'div',
           }),
         ],
@@ -613,7 +608,7 @@ export const components: CompileAwayComponentsMap = {
 
     return createMitosisNode({
       ...node,
-      name: node.properties.$tagName ?? 'div',
+      name: nodeTagName ?? 'div',
     });
   },
 };


packages/cli/src/__tests__/compile/__snapshots__/compile.test.ts.snap
```diff
@@ -405,55 +390,11 @@ function MyComponent(props) {
...
-              <p>
-                Some more great things to elaborate on that wonderful things you
-                have to tell your audience
-              </p>
-            </div>
+            <p>Something Great to Say</p>
+            <p>
+              Some more great things to elaborate on that wonderful things you
+              have to tell your audience
+            </p>
```
liamdebeasi, line 397: This appears to be an existing bug. The CSS is still available on the Text node, but the generator doesn't handle it properly. Interestingly, when I try to reproduce this with emotion styles in `componentToReact` I couldn't repro the issue

packages/core/src/parsers/builder/builder.ts
```diff
@@ -479,166 +479,6 @@ const componentMappers: {
       ),
     });
   },
-  Text: (block, options) => {
```
kylefowler, line 482: will this change impact anyone who has their own Text component when using anything that works with builder json, ie editor ai or normal codegen?

liamdebeasi: yes, this will impact anyone who has a Builder Text component

packages/core/src/__tests__/builder/__snapshots__/builder.test.ts.snap
```diff
@@ -2730,89 +2743,89 @@ export default function MyComponent(props) {
     <div>
       <Fragment>
         <Show when={props.conditionA}>
-          <span>Content0</span>
+          <Text $tagName=\\"span\\" text=\\"Content0\\" />
```
midhunadarvin, line 2746: ❔**question:** 
1. The`span`/`div` tag, for gen2 sdks, would be transformed to the framework specific element right ? So with this change we would need update the corresponding tests / snapshots in the gen 2 sdks repo right ? 
2. Since there are only changes in the snapshots for `builderContentToMitosisComponent` tests and there are no snapshot changes for `componentToBuilder` tests, I'm assuming we won't have any breaking change in Builder JSON with regards to how mitosis is being used within AI services. Just wanted to confirm that.

midhunadarvin (approved):
LGTM. I'll suggest a review from @samijaber / @sidmohanty11 too.

packages/core/src/__tests__/builder/__snapshots__/builder.test.ts.snap
```diff
@@ -3569,32 +3579,35 @@ exports[`Builder > bindings 1`] = `
       "bindings": {
         "css": {
           "bindingType": "expression",
-          "code": "{backgroundColor:'rgba(0, 0, 0, 1)',display:'flex',paddingLeft:'72px',paddingRight:'72px',paddingTop:'25px',paddingBottom:'25px',alignItems:'start',gap:'10px',fontFamily:'Poppins, -apple-system, Roboto, Helvetica, sans-serif',fontSize:'16px',color:'rgba(255, 255, 255, 1)',fontWeight:'700',textTransform:'uppercase',justifyContent:'start'}",
+          "code": "{\\"backgroundColor\\":\\"rgba(0, 0, 0, 1)\\",\\"display\\":\\"flex\\",\\"paddingLeft\\":\\"72px\\",\\"paddingRight\\":\\"72px\\",\\"paddingTop\\":\\"25px\\",\\"paddingBottom\\":\\"25px\\",\\"alignItems\\":\\"start\\",\\"gap\\":\\"10px\\",\\"fontFamily\\":\\"Poppins, -apple-system, Roboto, Helvetica, sans-serif\\",\\"fontSize\\":\\"16px\\",\\"color\\":\\"rgba(255, 255, 255, 1)\\",\\"fontWeight\\":\\"700\\",\\"textTransform\\":\\"uppercase\\",\\"justifyContent\\":\\"start\\"}",
```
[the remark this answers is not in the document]
liamdebeasi: The difference comes down to our use of both `JSON.stringify` and `json5.stringify`. The former adds double quotes, the latter does not.

More specifically, in `main` if we find a node with a single child Text node, we will try to merge code which uses [`json5.stringify` for the CSS](https://github.com/BuilderIO/mitosis/blob/6d343e949ef357e6b03da52f07f1d53619bb40df/packages/core/src/parsers/builder/builder.ts#L993). This code was removed in this PR, so `json5.stringify` is never called and instead the original [`JSON.stringify` formatting is preserved](https://github.com/BuilderIO/mitosis/blob/6d343e949ef357e6b03da52f07f1d53619bb40df/packages/core/src/parsers/builder/builder.ts#L957)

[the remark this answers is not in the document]
liamdebeasi: Also we don't have that code path anymore because by merging the code we compile away the Text node which we are trying to avoid.

PR #1772, opened 2025-05-30
fix: Angular generator handles self closing tags with components properly

liamdebeasi:
## Description

The Angular generator does not correctly check for self closing tags. This can result in invalid JSX. In the provided test case, the Mitosis node is a custom component named `CustomComponent`. However, it should be rendered as an `input` element.

The Angular generator correctly renders the opening tag as `<input>`. However, it does not close it correctly. This is because when it checks to see if the element is a self closing element it looks at `json.name` which is `CustomComponent` in this case.

I fixed the problem by having the generator check `element` which will be the actual rendered tag name.

**Example test** [Link to test run](https://cloud.nx.app/runs/IHOGrzE1oB/task/%40builder.io%2Fmitosis%3Atest)
output:
```diff
FAIL  src/__tests__/angular.tag.test.ts > Angular tag name output > ngmodule
Error: Snapshot `Angular tag name output > ngmodule 1` mismatched

- Expected
+ Received

  "import { NgModule } from \"@angular/core\";
  import { CommonModule } from \"@angular/common\";

  import { Component } from \"@angular/core\";

  @Component({
    selector: \"my-component\",
    template: `
-     <input placeholder=\"placeholder text\" />
-   `,
+
+           <input  placeholder=\"placeholder text\" ></input>
+           `,
    styles: [
      `
        :host {
          display: contents;
        }
      `,
    ],
  })
  export default class MyComponent {}

  @NgModule({
    declarations: [MyComponent],
    imports: [CommonModule, CustomComponentModule],
    exports: [MyComponent],
  })
  export class MyComponentModule {}
  "
```


packages/core/src/generators/angular/signals/blocks.ts
```diff
@@ -335,19 +336,21 @@ const getElementTag = (json: MitosisNode, blockOptions?: AngularBlockOptions) =>
...
+  if (selector) {
+    try {
+      ({ element, classNames, attributes } = parseSelector(`${selector}`));
+    } catch {
+      element = tagName ?? kebabCase(json.name);
+    }
+  }
+  if (!element) {
+    if (isComponent) {
+      element = tagName ?? kebabCase(json.name);
```
liamdebeasi, line 350: The Angular Signals codepath was incorrectly `<CustomComponent ...></CustomComponent>`. The new code here is now the same as how the non-signals Angular generator works.

sidmohanty11 (approved):

PR #1797, opened 2026-05-28
fix[angular]: remove deprecated allowSignalWrites from generated effects

floating-dynamo:
## Description
- Conditionally add the `{allowSignalWrites: true}` from the Angular signals generator only for Angular < 19 cause it is deprecated in 19+.
- This was causing repeated deprecated warnings for projects using Angular 19+ - [Doc](https://angular.love/angular-19-whats-new#:~:text=A%20significant%20change%20is%20the%20elimination%20of%20the%20allowSignalWrites%20flag)

### Before
image

### After
image

image

---

### DIFF
--- a/.changeset/hip-cats-brake.md
+++ b/.changeset/hip-cats-brake.md
@@ -2,4 +2,4 @@
 '@builder.io/mitosis': minor
 ---
 
-Removed the deprecated allowSignalWrites from generated effects for Angular 19+
+Angular signals: gate the deprecated `allowSignalWrites` effect option on the runtime Angular version (`VERSION.major < 19`) instead of dropping it. This silences the deprecation warning on Angular 19+ while keeping backwards compatibility with lower versions.
--- a/.changeset/hip-cats-brake.md
+++ b/.changeset/hip-cats-brake.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/mitosis': minor
+---
+
+Removed the deprecated allowSignalWrites from generated effects for Angular 19+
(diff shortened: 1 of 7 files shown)


sidmohanty11 (changes requested):
we would need to migrate our angular sdk from using v17: https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/angular/package.json#L31 to v19 first as we depend on v17 + mitosis for the built output/** which will remove the allowSignalWrites

sidmohanty11:
in that case we will only support angular 19+ moving forward^^

floating-dynamo:
> we would need to migrate our angular sdk from using v17: [BuilderIO/builder@`main`/packages/sdks/output/angular/package.json#L31](https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/angular/package.json#L31) to v19 first as we depend on v17 + mitosis for the built output/** which will remove the allowSignalWrites

And with that I will have to make a minor upgrade of the gen2 angular sdk right before I merge this PR? @sidmohanty11

sidmohanty11:
yeah!

floating-dynamo:
> yeah!

Can you review this PR - https://github.com/BuilderIO/builder/pull/4633 @sidmohanty11. Let me know if this is what is required.

sidmohanty11 (approved):
lgtm, update the PR description

.changeset/hip-cats-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/mitosis': minor
```
sidmohanty11: oh! this should be patch
```suggestion
'@builder.io/mitosis': patch
```

floating-dynamo:
Also I noticed I still have the e2e package.json changes, I think they are not needed now, I will revert them back.

PR #1750, opened 2025-04-14
fix: issues for angular signals api

nmerget:
## Description

Please provide the following information:

Minor fixes for Angular `api=signals`

Make sure to follow the PR preparation steps in [CONTRIBUTING.md](../CONTRIBUTING.md#preparing-your-pr) before submitting your PR:

- [x] format the codebase: from the root, run `yarn fmt:prettier`.
- [ ] update all snapshots (in core & CLI): from the root, run `yarn test:update`
- [x] add Changeset entry: from the root, run `yarn g:changeset` and follow the CLI instructions. Alternatively, use the Changeset Github Bot to create the file.

PR #1803, opened 2026-06-05
fix(angular): guard onMount execution in SSR with window check

floating-dynamo:
## Description
- We are seeing the angular-17 and angular-19 ssr e2e checks fail for angular sdk side when we ugrade mitosis from 0.11.x to latest 0.13.x of mitosis.
- Angular SDK PR with failing checks - https://github.com/BuilderIO/builder/pull/4640
- The change made in this PR for mitosis is causing the test failures - https://github.com/BuilderIO/mitosis/pull/1750

## Testing
Tested the e2e builds with the mitosis dev release and they are passing now

image

image

### DIFF
--- a/packages/core/src/generators/angular/signals/component.ts
+++ b/packages/core/src/generators/angular/signals/component.ts
@@ -298,9 +298,7 @@ Please add a initial value for every state property even if it's \`undefined\`.`
 
     if (!emptyOnMount) {
       // `onMount` runs only in the browser, after view initialization. Guard it
-      // so it does not execute during SSR. Other lifecycle code (slot rendering
-      // via `_updateView()` and attribute passing) must run on the server too,
-      // and is therefore emitted unguarded.
+      // so it does not execute during SSR.
       addCodeNgAfterViewInit(
         json,
         `if (typeof window !== 'undefined') {
--- a/packages/core/src/generators/angular/signals/component.ts
+++ b/packages/core/src/generators/angular/signals/component.ts
@@ -297,7 +297,16 @@ Please add a initial value for every state property even if it's \`undefined\`.`
     // Hooks
 
     if (!emptyOnMount) {
-      addCodeNgAfterViewInit(json, json.hooks.onMount.map((onMount) => onMount.code).join('\n'));
+      // `onMount` runs only in the browser, after view initialization. Guard it
+      // so it does not execute during SSR. Other lifecycle code (slot rendering
+      // via `_updateView()` and attribute passing) must run on the server too,
+      // and is therefore emitted unguarded.
+      addCodeNgAfterViewInit(
+        json,
+        `if (typeof window !== 'undefined') {
+          ${json.hooks.onMount.map((onMount) => onMount.code).join('\n')}
+        }`,
+      );
     }
 
     // Angular interfaces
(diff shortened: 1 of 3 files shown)


packages/core/src/generators/angular/signals/component.ts
```diff
@@ -436,9 +443,7 @@ Please add a initial value for every state property even if it's \`undefined\`.`
                   .filter(([_, value]) => !isHookEmpty(value))
                   .map(([key, value]) => {
                     return `${key}() {
-                    if (typeof window !== 'undefined') {
                 ${value.code}
-                }
```
sidmohanty11, line 441: i dont fully understand the changes in this pr, could you please explain why this change was needed?

this will remove the window guard from all custom hooks. so if you check in the mitosis plugin cjs file you will see a lot of custom angular hooks added specifically not sure how it would affect it

packages/core/src/generators/angular/signals/component.ts
```diff
@@ -297,7 +297,14 @@ Please add a initial value for every state property even if it's \`undefined\`.`
...
 
     if (!emptyOnMount) {
-      addCodeNgAfterViewInit(json, json.hooks.onMount.map((onMount) => onMount.code).join('\n'));
+      // `onMount` runs only in the browser, after view initialization. Guard it
+      // so it does not execute during SSR.
+      addCodeNgAfterViewInit(
+        json,
+        `if (typeof window !== 'undefined') {
+          ${json.hooks.onMount.map((onMount) => onMount.code).join('\n')}
+        }`,
```
sidmohanty11, line 306: it looks like we are adding window guard but from tests we are removing it? im bit confused

packages/core/src/generators/angular/signals/component.ts
floating-dynamo: Yes, but this also caused the `setAttribute` to be skipped in SSR so there were failing tests:

image

So this was a bigger blanket for the guard, which I have moved only for the onMount case. But I need to understand what you mean by this also - "so if you check in the mitosis plugin cjs file you will see a lot of custom angular hooks added specifically not sure how it would affect it"

sidmohanty11: iiuc, we need the setAttribute to be applied both server side and client side? in that case would that go into ngOnInit without the guard instead of this? can you check where does it live currently in v11 output

floating-dynamo: Sure I will check this once, thanks!

floating-dynamo: I think we cannot move that in ngOnInit cause the ref.current will not be defined during that phase

image

floating-dynamo: And in v11 output also, the setAttributes is in the ngAfterViewInit with no guard for typeof window. It only has window guard for setAttributes running within an effect.

sidmohanty11 (approved):

packages/core/src/generators/angular/signals/component.ts
sidmohanty11: thanks for verifying

floating-dynamo: Thanks for all the suggestions / checks, I do no understand angular much but got to learn a bit about it 🤝

